PR #18, opened 2020-07-29
Feature/scheduler - Optimization of the scheduler system

R0bbyYT:
This pull request fixes the problem that the tasks are not executed by the scheduler at the same time. The tasks which should be executed at the same time always have a delay of 5-250ms in the next executions.

# Why is this a problem?

For example, if you need a countdown for a game mode that counts down by 1 every second, the countdown with the current Minestom system will take 101-105 seconds to reach 0, although the countdown is only 100 seconds. 
With the new system, the countdown takes exactly 100 seconds to reach 0.

# How can I check this?

Example class for the countdown:

```JAVA
import java.util.concurrent.TimeUnit;

public class Countdown {

    private final int countdown;
    private int value;
    private long time;
    private boolean finished;

    public Countdown(int value) {
        this.countdown = value;
        this.value = value;
        this.finished = false;
    }

    public void run() {
        if(this.finished) return;
        if (this.value == this.countdown) {
            time = System.currentTimeMillis();
        }
        if (value == 0) {
            long process = System.currentTimeMillis() - this.time;
            System.out.println("The process took " + process + "ms or " + TimeUnit.MILLISECONDS.toSeconds(process) + "sec");
            this.finished = true;
        }

        this.value -= 1;
    }
}
```

Minestom code example

```JAVA
Countdown countdown = new Countdown(100);
MinecraftServer.getSchedulerManager().addRepeatingTask(new TaskRunnable() {
	@Override
        public void run() {
	    countdown.run();
        }
}, new UpdateOption(1, TimeUnit.SECOND));

```
New System code example
```JAVA
Countdown countdown = new Countdown(100);
MinecraftServer.getSchedulerManager().buildTask(countdown::run).repeat(1, TimeUnit.SECOND).buildTask();
```

Once you have adapted the code to your project you can now run the server and test it.


TheMode:
Merged. The only remaining thing to do would be to update the wiki concerning the scheduler API (which is now outdated): https://github.com/Minestom/Minestom/wiki/Schedulers

PR #27, opened 2020-08-04
Update/Improve Pathfinding

Articdive:


Articdive:
Could be improved by using Tags, will avoid for now.

PR #39, opened 2020-08-18
Rename plugins to Extensions and add basic support.

Articdive:


Eoghanmc22:
Made a test extension seems to work will merge.

PR #66, opened 2020-11-05
Implement Cloneable whereever Applicable

Geolykt:
I didn't change the ItemMeta#copy() function as that is more API-breaking than the rest of the changes. The old Copy methods were deprecated.

My Reasoning behind using Cloneable (and then overriding Object#clone()) is that Cloneable is a relatively widely-accepted API to signify that the Object may be cloned. It is also may weaken some of the ambiguity beween Position#copy(Vector) or Position#copy(Position) and Position#clone() for example.


TheMode:
In fact, it was using #clone() at first but I decided to switch to #copy() instead in order to be able to create our own "Copiable" interface later on.

https://stackoverflow.com/questions/4081858/about-java-cloneable
https://www.artima.com/intv/bloch.html#part13

Geolykt:
Makes sense

PR #72, opened 2020-12-01
Allow EntityCreature in GoalSelector to be null

LeoDog896:
This allows an unsafe operation that allows an EntityCreature to be passed in a constructor as null, or not passed at all, and can be null wherever during the object. If handled improperly, it may lead to unexpected behavior.


LeoDog896:
You can just make a shell creature

PR #81, opened 2020-12-24
Implements player head and firework metas

R0bbyYT:
This pull request adds implementations for the player head and fireworks meta.


TheMode:
I have 2 suggestions:

- Use thread-safe collection when possible (concurrent hashmap, copyonwriteArrayList/Set when there aren't too many writes)
- Remove #getOwningPlayer as players UUID don't necessarily represent their Mojang one and that most of the time the skin used isn't the one of a connected player

R0bbyYT:
All right, I've changed it.

PR #85, opened 2020-12-28
Fix block placement inside entities.

mworzala:
This changes the rounding method of `Instance#getChunkAt` to `Math#floor` as opposed to casting to an int. Casting a float such as -0.5 to an int will result in 0, as opposed to the desired -1. This means that negative x or z chunk borders were off by one.

As a result, entities were being marked as in different chunks than they actually were and were ignored in collision.


TheMode:
Great! Could you just remove the debug print?

PR #88, opened 2020-12-28
Send swing packets to self

DigitalJokerMan:
LivingEntity#swingMainHand and LivingEntity#swingOffHand should show up for the client as well. Executing these functions on a player won't show up for the player themselves.


TheMode:
Did you test it out? IIRC the client itself is not aware of his own entity ID

DigitalJokerMan:
I did test it out, it does swing the arm on my client.

TheMode:
Tested and it does cause an infinite loop in my case because the method is also called when the player signals his arm movement

DigitalJokerMan:
Hmm, I could see there being situations where it is desirable to send the packet to the client and to not send it to the client. Maybe it could be an argument?

TheMode:
I agree that having the option is great, but I don't want to complicate the API too much.
There is probably a better way

Nesaak:
At least in Spigot, LivingEntity#swingOffHand() & LivingEntity#swingMainHand() does send the packet to the client you call it on.

PR #89, opened 2020-12-29
Add BlockIterator and LivingEntity#getTargetBlockPosition

Nesaak:
This was adapted from spigots BlockIterator. The only problem I find here is that since there is no default eye position of entities, it starts tracing from the feet by default. This can be fixed by just setting entities eye position, it may be worth making a note of that somewhere.


TheMode:
Looks like there is an issue with "import javafx.geometry.Pos;", probably a typo since it doesn't seem to be used

Geolykt:
Yep, JavaFX cannot be used in openJDK

PR #90, opened 2020-12-29
Delegate creation of spawn packets to Entity#getSpawnPacket

Nesaak:


TheMode:
Hmm it should be possible to simplify it further by sending the packets directly in Entity#addViewer instead of in the subclasses

Nesaak:
> Hmm it should be possible to simplify it further by sending the packets directly in Entity#addViewer instead of in the subclasses

Yes absolutely, I'm working on that now. Do not merge because I also need to null check, forgot about that part.

TheMode:
Has been implemented in #152

PR #93, opened 2020-12-31
Change ColoredText usages to JsonMessage

mworzala:
ColoredText was rather restrictive, and the client will accept any valid json message in places where some tags (such as `hoverEvent` in an actionbar message) are not supported.

This branch also adds a `title` command to the demo project which will send a title to the sender either as text or any valid json message.


mworzala:
The failure seems due to too many missing javadocs. I didnt particularly change the javadocs (besides a few ColoredChat references), but I can add some of the missing ones if that is actually the issue.

PR #95, opened 2021-01-02
Inventory double click ArrayIndexOutOfBoundsException

mworzala:
When you double click an item in an inventory it would give an array out of bounds exception because the internal slot it was trying to access computed to -9. 

The fix in this PR is to supply an offset of 9 to `playerInventory.getItemStack` in one of the `InventoryClickLoopHandlers`. In my testing it appears to behave correctly, however I am unsure why it was offset previously. I think it is worth someone with more knowledge doing some testing on their own to make sure this doesnt have other side effects.

In other notes, I removed the hotbar loop handler because the inventory one handles those slots, and changed a comment for accuracy.

### DIFF
--- a/src/main/java/net/minestom/server/inventory/Inventory.java
+++ b/src/main/java/net/minestom/server/inventory/Inventory.java
@@ -405,7 +405,7 @@ public class Inventory implements InventoryModifier, InventoryClickHandler, View
         }
 
         if (isInWindow) {
-                setItemStack(slot, clickResult.getClicked());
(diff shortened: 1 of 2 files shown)


TheMode:
Looks great, could you remove the debugging code?

mworzala:
thats done, I even looked for debugging code this time :P apparently not well enough.
I also swapped a 9 out for the player inventory offset.

I'm not super confident on any details of the vanilla behavior, but all of my stacking tests work the same as a vanilla server.

PR #96, opened 2021-01-02
Add Potion Add and Remove events

ThatCreeper:
Adds EntityPotionAddEvent and EntityPotionRemoveEvent


ThatCreeper:
Also fixes an issue with the potion api which would cause the potion to never go away due to time

PR #99, opened 2021-01-09
Add a method to open a book in Player

mworzala:
Currently WrittenBookMeta exists, but with no interaction to open the book ui for the player.

This PR adds a method on the player to open a given `WrittenBookMeta` for the player regardless of whether the book is actually in their hand or not.


mworzala:
I can remove the test command if its better to avoid tiny random commands.

PR #100, opened 2021-01-10
Cleanup and remove most warnings in the project

KrystilizeNevaDies:
Cleanup and remove most warnings in the project

- Added Generic types where sensible
- Fixed some imports
- Also found a bug in EntityDamageEvent where It was not previously Cancellable
- Added give exp to after orb pickup
- Added default serial ID's to serializable elements + a comment on minestom versioning TBD

### DIFF
--- a/src/autogenerated/java/net/minestom/server/instance/block/Block.java
+++ b/src/autogenerated/java/net/minestom/server/instance/block/Block.java
@@ -2495,7 +2495,8 @@ public enum Block {
     @Nullable
     private NamespaceID blockEntity;
 
-    private boolean singleState;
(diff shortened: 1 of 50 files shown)


TheMode:
Most of the changes here are only breaking existing features or remove intent for future support, not to even mention readability

KrystilizeNevaDies:
Okay well does that mean that most of these are just not needed? If so I'll just commit the smaller bugs I found seperately.

TheMode:
Would be better to have them separately yeah

There is no need to treat all the warnings, it is only when it can improve readability. Otherwise we just end up with a bunch of annotations where we don't know what to do with

PR #104, opened 2021-01-15
Microoptimization cleanup

LeoDog896:
Just some small micro optimizations including preferring fastutils, Removing unnecessary checks already dealt with using @NotNull, all since I had extra unplanned free time.

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -476,9 +476,7 @@ public final class MinecraftServer {
         MinecraftServer.chunkViewDistance = chunkViewDistance;
         if (started) {
 
-            final Collection<Player> players = connectionManager.getOnlinePlayers();
-
-            players.forEach(player -> {
+            for (final Player player : connectionManager.getOnlinePlayers()) {
                 final Chunk playerChunk = player.getChunk();
                 if (playerChunk != null) {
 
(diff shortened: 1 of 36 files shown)


TheMode:
I am personally not a fan of returning non-standard types (like in Chunk/DynamicChunk)

LeoDog896:
> I am personally not a fan of returning non-standard types (like in Chunk/DynamicChunk)

Got it, ill change that

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -478,9 +476,7 @@ public static void setChunkViewDistance(int chunkViewDistance) {
         MinecraftServer.chunkViewDistance = chunkViewDistance;
         if (started) {
 
-            final Collection<Player> players = connectionManager.getOnlinePlayers();
-
-            players.forEach(player -> {
+            for (final Player player : connectionManager.getOnlinePlayers()) {
```
Protonull, line 479: Doing this *might* be a bad idea. Take Java's `SynchronizedCollection` for example where `forEach()` is synchronised but the iterator, spliterator, stream, and parallelStream are all explicitly labelled with "Must be manually synched by user!" It might be better to defer to the collection's encapsulation for iteration, particularly when retrieving information outside of an instance's thread. Not sure how much of a problem this is, if any, but just something to keep in mind.

TheMode: You are right but this is not an issue in this case since sending packets is thread-safe, and #refreshVisibleChunks only read. And reading a wrong value there will not mess with anything

PR #107, opened 2021-01-18
Better KeepAlive and Connection handling

LeoDog896:
This uses an alternative conditional function to handle keepalives and uses a stack instead of a queue for connected players.


NickAcPT (approved):

TheMode:
I am reticent about making Player#init public, and I do believe that a queue is better suited for the waiting collection as the order can be important (same in the other places where queues are used)

PR #108, opened 2021-01-19
[Deprecation] Remove type specific methods in Arguments

LeoDog896:
For future knowledge, encourage the use of generics versus type-specific casting. If anything, getObject is still public.


TheMode:
Adding the deprecated annotation would probably be better in the meantime

LeoDog896:
> Adding the deprecated annotation would probably be better in the meantime

Breaking changes have been introduced multiple times before and I don't think it's going to stop with this PR.

PR #109, opened 2021-01-19
Added ExceptionManager for custom exception handling

DerTyan:
First version, open for feedback.

You can set the ExceptionHandler with e.g.
`MinecraftServer.getExceptionManager().setExceptionHandler(Throwable::printStackTrace);`

Throwable::printStackTrace is currently the default

And use it with
`MinecraftServer.getExceptionManager().handleException(exception);`

### DIFF
--- a/src/main/java/net/minestom/server/extensions/ExtensionManager.java
+++ b/src/main/java/net/minestom/server/extensions/ExtensionManager.java
@@ -3,6 +3,7 @@ package net.minestom.server.extensions;
 import com.google.gson.Gson;
 import net.minestom.dependencies.DependencyGetter;
 import net.minestom.dependencies.maven.MavenRepository;
+import net.minestom.server.MinecraftServer;
 import net.minestom.server.extras.selfmodification.MinestomExtensionClassLoader;
 import net.minestom.server.extras.selfmodification.MinestomRootClassLoader;
 import net.minestom.server.ping.ResponseDataConsumer;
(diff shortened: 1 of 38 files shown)


Geolykt:
What are the use-cases for this? It is rare than an exception is printed and suppressing it will not have much of an effect as it would be in the catch block at this point. Additionally in this state it is kind-of hard to know from where the exception came from

DerTyan:
@Geolykt I tried setting up an ExceptionHandler for uncaught exceptions like NullPointer, OutOfBounds, etc.
For example if such an Exception is thrown in the Instance#tick, Minestom will print the stack trace and I have no option to catch that exception myself and send it to Sentry, our error reporting tool.

And the Throwable does contain the StackTrace, so it would know where it came from.

Geolykt:
Fair point.

src/main/java/net/minestom/server/extensions/ExtensionManager.java
```diff
@@ -113,7 +114,7 @@ public void loadExtensions() {
                 attemptSingleLoad(discoveredExtension);
             } catch (Exception e) {
                 discoveredExtension.loadStatus = DiscoveredExtension.LoadStatus.LOAD_FAILED;
-                e.printStackTrace();
+                MinecraftServer.getExceptionManager().handleException(e);
                 LOGGER.error("Failed to load extension {}", discoveredExtension.getName());
                 LOGGER.error("Failed to load extension", e);
```
mworzala, line 119: This will print the stack trace twice (once with logger, once without). Probably the second log message should be gone and the exception handler should be called afterwards.

mworzala (commented):
Im a big fan of this change.

src/main/java/net/minestom/server/extensions/ExtensionManager.java
DerTyan: Done.

mworzala (approved):

LeoDog896:
Love this! :heart:

PR #110, opened 2021-01-20
Added withProperties in BlockAlternative.java

MrBretze:
Example: 

blockAlternative.withProperties("facing");

Return the facing of the block or a empty string


TheMode:
Wouldn't `getProperty` be better?

MrBretze:
I can change this yes, is not a big change 😁😅

PR #112, opened 2021-01-23
Adding hasArgument and hasProperty

MrBretze:
Hello, i simple add to check if Arguments.java has a argument or not


src/main/java/net/minestom/server/instance/block/BlockAlternative.java
```diff
@@ -41,6 +41,10 @@ public String getProperty(String key) {
         return null;
     }
 
+    public boolean hasProperty(String key) {
+        return Arrays.stream(properties).anyMatch(s -> s.equals(key));
```
TheMode, line 45: Should't it be
```java
return getProperty(key) != null;
```
?

Geolykt: either way this implementation has it's issues as it does not match the implementation of getProperty() as well as that using a Stream might be a bit too overkill for that

MrBretze: Yeah, i dont not why i am using a Stream sorry ><

TheMode:
Currently `CommandContext#has`

PR #115, opened 2021-01-25
Add ability to get Block type from given position

iamceph:
A little nice to have shortcut in the Instance & Chunk API :)


KrystilizeNevaDies:
I think it would be a better idea to omit this from Chunk and only have it in Instance, the idea being that chunk deals with the actual short array while instance can have this method as a shorthand, will have to check with other contributors though.

iamceph:
I get what are you saying here but it is basically the same method as getBlockStateId(). I'd say that you are right, I just wanted to add ability to getBlock too, as in BlockModifier you can do setBlock and I ended up in adding it into the Chunk too 😅 I can make the changes if needed.

KrystilizeNevaDies:
Will have to see what TheMode says

KrystilizeNevaDies:
Talked with TheMode and he also thinks that the getblock method should be only in instance. If you'd like to just change that, I'll merge, if not, I'll do it myself.

iamceph:
Fixed :)

PR #122, opened 2021-01-27
Add Resource loading API

### LINKED ISSUE #106: Resource management
It would be great if extensions had access to an established resource management API. What do I mean by that? Take Bukkit/Spigot/Paper for example and how it handles plugin configs, where config files are stored within the plugin jar and can be saved into the plugin's data folder, and when loading the config, the plugin prefers the data file rather than the internal resource. Though if added, it would be great if this were expanded to any arbitrary file rather than just a specific config file.

It would establish a standard within extensions (that can be made specifically thread safe if necessary) meaning reduced duplicate code, as well as explicit support for data files, particularly human readable data files.

So I can just save / load config files and other such things and get on with making my extension.

### LINKED ISSUE #252: Get data folder for extension
Extensions should have a method to retrieve a data folder for the extension to store stuff in. Ideally, this wouldn't be in the `extensions` folder so we can reduce clutter a bit.

mworzala:
This adds a simple api to read files from the extension jar resources, as well as preference over files in the extension directory.

The obvious use case here is config files.

resolves #106 
resolves #252

### DIFF
--- a/src/main/java/net/minestom/server/extensions/Extension.java
+++ b/src/main/java/net/minestom/server/extensions/Extension.java
@@ -80,11 +80,12 @@ public abstract class Extension {
      */
     @Nullable
     public InputStream getResource(@NotNull String fileName) {
-        Path targetFile = getDataDirectory().resolve(fileName);
+        final Path targetFile = getDataDirectory().resolve(fileName);
         try {
             // Copy from jar if the file does not exist in extension directory
-            if (!Files.exists(targetFile))
+            if (!Files.exists(targetFile)) {
                 savePackagedResource(fileName);
+            }
 
             return Files.newInputStream(targetFile);
         } catch (IOException ex) {
@@ -102,7 +103,7 @@ public abstract class Extension {
     @Nullable
     public InputStream getPackagedResource(@NotNull String fileName) {
         try {
-            URL url = getClass().getClassLoader().getResource(fileName);
+            final URL url = getClass().getClassLoader().getResource(fileName);
             if (url == null) {
                 getLogger().debug("Resource not found: {}", fileName);
                 return null;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/extensions/Extension.java
```diff
@@ -55,6 +65,79 @@ protected Logger getLogger() {
...
+     * Gets a resource from the extension directory, or from inside the jar if it does not
+     * exist in the extension directory.
+     * <p>
+     * If it does not exist in the extension directory, it will be copied from inside the jar.
+     *
+     * @param fileName The file to read
+     * @return The file contents, or null if there was an issue reading the file.
+     */
+    @Nullable
+    public InputStream getResource(@NotNull String fileName) {
```
kezz, line 82: I wonder if this should be a path instead - they can be relative to the jar/data directory. I just hare dealing with separators :')

src/main/java/net/minestom/server/extensions/Extension.java
```diff
@@ -55,6 +65,79 @@ protected Logger getLogger() {
...
+    /**
+     * Gets a resource from inside the extension jar.
+     *
+     * @param fileName The file to read
+     * @return The file contents, or null if there was an issue reading the file.
+     */
+    @Nullable
+    public InputStream getPackagedResource(@NotNull String fileName) {
+        try {
+            final URL url = getClass().getClassLoader().getResource(fileName);
```
kezz, line 106: Should this not use the extension class loader? Can be pulled from the origin.

src/main/java/net/minestom/server/extensions/ExtensionManager.java
```diff
@@ -216,6 +216,18 @@ private Extension attemptSingleLoad(@NotNull DiscoveredExtension discoveredExten
             LOGGER.error("Main class '{}' in '{}' has no logger field.", mainClass, extensionName, e);
         }
 
+        // Set dataDirectory
+        try {
+            Field dataDirectoryField = Extension.class.getDeclaredField("dataDirectory");
```
kezz, line 221: This could live in the origin - would enable others to easily get the data folder for an extension and also reduce the amount of reflection we do.

kezz (changes requested):
Looks great to me! Just a few minor comments

mworzala:
Closing due to issue with origin repository. Recreated as #302

PR #124, opened 2021-01-29
WIP: Event enhancements

iamceph:
Modified events:

- PlayerBlockPlaceEvent - add originalBlockID (block before replacement with this event)

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerBlockPlaceEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerBlockPlaceEvent.java
@@ -25,18 +25,21 @@ public class PlayerBlockPlaceEvent extends PlayerEvent implements CancellableEve
     private Data blockData;
     private final BlockPosition blockPosition;
     private final Player.Hand hand;
+    private final short originalBlockStateId;
 
     private boolean consumeBlock;
 
     private boolean cancelled;
 
     public PlayerBlockPlaceEvent(@NotNull Player player, @NotNull Block block,
-                                 @NotNull BlockPosition blockPosition, @NotNull Player.Hand hand) {
+                                 @NotNull BlockPosition blockPosition, @NotNull Player.Hand hand,
+                                 short originalBlockStateId) {
         super(player);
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/event/player/PlayerMoveEvent.java
```diff
@@ -11,15 +11,27 @@
...
         this.newPosition = newPosition;
     }
 
+    /**
+     * Gets current position of the player.
+     *
+     * @return the current position
+     */
+    @NotNull
+    public Position getCurrentPosition() {
```
mworzala, line 31: In my opinion this should be `getOldPosition`. It is correct that the position has yet to change, however it can be confusing if somebody does not pay attention.

src/main/java/net/minestom/server/event/player/PlayerBlockPlaceEvent.java
```diff
@@ -149,6 +152,16 @@ public BlockPosition getBlockPosition() {
         return hand;
     }
 
+    /**
+     * Original block on the {@link BlockPosition} which the player is trying to replace.
+     * Mostly {@link Block#AIR}, but can be different.
+     *
+     * @return an ID of block on the location
+     */
+    public short getOriginalBlockStateId() {
```
mworzala, line 161: I think this should be `getOldBlockStateId`, for the same reason as in `PlayerMoveEvent`

mworzala (changes requested):
For the most part, I think this is good. It does, however, open up a question of whether more events should have this level of detail. 
For example, should `EntityVelocityEvent` have the old velocity for consistency?

src/main/java/net/minestom/server/event/player/PlayerBlockPlaceEvent.java
iamceph: Agree, I will fix that. :)

TheMode:
The new block API will offer change to the place event, `PlayerMoveEvent#getCurrentPosition` might want to be retargeted. Though I am not sure if this is a necessary addition?

PR #123, opened 2021-01-29
Event Handler support for Extensions

ch99q:
We currently have the `addEventCallback()` method implementation in global and instance scope, but we need to add this for Extensions.

A few things are missing in the Extension API and event hooks are one of them, currently, when hooking an event from an extension and reloading the extension, there will not happen any unregister/unhooking of the event. Therefore making an event handler duplication.

### Implementation suggestion
Like we see in the Instance structure, the Instance class has en declared method called `addEventCallback()`, the suggestion is to add the same thing to the Extension.
#### Suggested methods
`EventHandler.addEventCallback(Extension extension, Class<E> eventClass, EventCallback<E> eventCallback)`
`EventHandler.removeEventCallback(Extension extension, Class<E> eventClass, EventCallback<E> eventCallback)`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ gradle-app.setting
 
 # When compiling we get a docs folder
 /docs
+
+# Mac
+.DS_STORE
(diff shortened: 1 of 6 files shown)


KrystilizeNevaDies:
Why did you close this issue? I thought it would be a good idea.

PR #129, opened 2021-02-03
Improve extension system

### LINKED ISSUE #127: Improve the extension system
Currently, the extension has a few flaws:

- [x] Unloading an extension does not automatically remove its callbacks. Event callback ownership could help on this point. 
 Should this include an API change or not? It might be possible to use the extension's classloader to determine if a callback is owned by a given extension when unloading that extension. 
 If we are willing to trade some memory for performance, Minestom could even cache the callbacks when they are added to immediately remove the extensions callbacks on unloading, without iterating through all active callbacks.
- [x] If Mixin are not used by an extension, and the Mixin library does not load, this should not crash the server.
- [x] Extensions should be able to mixin into some Minestom classes before they are loaded (InstanceContainer for instance). This might require to load extensions earlier than the `MinecraftServer.init()` call inside the Bootstrap class
 Because extension loading is currently toggleable, this option may require a command line option or a System property to be added to Minestom to turn extension loading.
- [x] Investigate issues with "new" JRE versions when J9 modules are present in the classpath, as security has been reinforced.

jglrxavpok:
Improvements over the current extension system

resolves #127

### DIFF
--- a/src/main/java/net/minestom/server/Bootstrap.java
+++ b/src/main/java/net/minestom/server/Bootstrap.java
@@ -1,5 +1,6 @@
 package net.minestom.server;
 
+import net.minestom.server.extensions.ExtensionManager;
 import net.minestom.server.extras.selfmodification.MinestomRootClassLoader;
 import net.minestom.server.extras.selfmodification.mixins.MixinCodeModifier;
 import net.minestom.server.extras.selfmodification.mixins.MixinServiceMinestom;
@@ -21,6 +22,8 @@ public final class Bootstrap {
             ClassLoader classLoader = MinestomRootClassLoader.getInstance();
             startMixin(args);
             MinestomRootClassLoader.getInstance().addCodeModifier(new MixinCodeModifier());
+
+            ExtensionManager.loadCodeModifiersEarly();
 
             MixinServiceMinestom.gotoPreinitPhase();
             // ensure extensions are loaded when starting the server
--- a/src/main/java/net/minestom/server/Bootstrap.java
+++ b/src/main/java/net/minestom/server/Bootstrap.java
@@ -7,6 +7,7 @@ import net.minestom.server.extras.selfmodification.mixins.MixinServiceMinestom;
 import org.spongepowered.asm.launch.MixinBootstrap;
 import org.spongepowered.asm.launch.platform.CommandLineOptions;
 import org.spongepowered.asm.mixin.Mixins;
+import org.spongepowered.asm.service.ServiceNotAvailableError;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -21,7 +22,12 @@ public final class Bootstrap {
         try {
             ClassLoader classLoader = MinestomRootClassLoader.getInstance();
             startMixin(args);
-            MinestomRootClassLoader.getInstance().addCodeModifier(new MixinCodeModifier());
+            try {
+                MinestomRootClassLoader.getInstance().addCodeModifier(new MixinCodeModifier());
+            } catch (RuntimeException e) {
+                e.printStackTrace();
+                System.err.println("Failed to add MixinCodeModifier, mixins will not be injected. Check the log entries above to debug.");
+            }
 
             ExtensionManager.loadCodeModifiersEarly();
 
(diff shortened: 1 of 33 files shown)


src/main/java/net/minestom/server/extras/selfmodification/MinestomRootClassLoader.java
```diff
@@ -279,4 +311,23 @@ public void addURL(URL url) {
...
+     * by an extension.
+     *
+     * While not perfect, this should detect any callback created via extension code.
+     * It is possible this current version of the implementation might struggle with callbacks created through external
+     * libraries, but as libraries are loaded separately for each extension, this *should not*(tm) be a problem.
+     *
+     * @param obj the object to get the extension of
+     * @return <code>Optional.empty()</code> if no extension has been found, <code>Optional.of(&lt;name&gt;)</code> with 'name' being the extension name
+     */
+    public static Optional<String> findExtensionObjectOwner(@NotNull Object obj) {
```
mworzala, line 326: This should return a nullable string rather than an optional to fit style.

mworzala:
Is it possible for an extension to tell whether its code modifiers were loaded or not? Missing them may cause the extension to lose its functionality, so there should be a way for them to know.

src/main/java/net/minestom/server/extras/selfmodification/MinestomRootClassLoader.java
jglrxavpok: Fixed in commit ee158c0deafeb6ded77496b32f5fde9f07f2f2ac

jglrxavpok:
> 
> 
> Is it possible for an extension to tell whether its code modifiers were loaded or not? Missing them may cause the extension to lose its functionality, so there should be a way for them to know.

The latest commit adds this feature, exposed via `getDescription()` and a convenience method inside Extension: `areCodeModifiersAllLoadedCorrectly()`

jglrxavpok:
With this last commit, the PR looks ready on my end.

mworzala (approved):
I think the method of determining an extension's owner is somewhat fickle, but I don't think it is a huge problem since the system should not be used in a production environment anyway.

jglrxavpok:
> 
> 
> I think the method of determining an extension's owner is somewhat fickle, but I don't think it is a huge problem since the system should not be used in a production environment anyway.

Are you talking about MinestomRootClassLoader#findExtensionObjectOwner ? If so, do you have ideas to improve upon it, or where it might fail?

src/main/java/net/minestom/server/extensions/Extension.java
```diff
@@ -14,6 +19,14 @@
     @SuppressWarnings("unused")
     private Logger logger;
 
+    /**
+     * Observers that will be notified of events related to this extension.
+     * Kept as WeakReference because entities can be observers, but could become candidate to be garbage-collected while
+     * this extension holds a reference to it. A WeakReference makes sure this extension does not prevent the memory
+     * from being cleaned up.
+     */
+    private Set<WeakReference<IExtensionObserver>> observers = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
TheMode, line 28: You probably want a way to "cleanup" the Set http://www.java2s.com/Code/Java/Collections-Data-Structure/Abagofweaklyreferencedobjects.htm

src/main/java/net/minestom/server/Bootstrap.java
```diff
@@ -20,7 +22,14 @@ public static void bootstrap(String mainClassFullName, String[] args) {
         try {
             ClassLoader classLoader = MinestomRootClassLoader.getInstance();
             startMixin(args);
-            MinestomRootClassLoader.getInstance().addCodeModifier(new MixinCodeModifier());
+            try {
+                MinestomRootClassLoader.getInstance().addCodeModifier(new MixinCodeModifier());
+            } catch (RuntimeException e) {
+                e.printStackTrace();
```
TheMode, line 28: Exceptions should be handled using `ExceptionManager` https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/exception/ExceptionManager.java#L17

jglrxavpok: No ExceptionManager exists at this point, so it can't be handled through an ExceptionManager

src/main/java/net/minestom/server/extensions/Extension.java
jglrxavpok: Fix in e40186a2aad5eca86a8cc9dd408bc1629ff5d267

PR #132, opened 2021-02-05
Use FastMath from Mad's data structures

LeoDog896:
This gives faster collision checking amongst other things -- and should at least _inspire_ us in the future to use fast classes.

### DIFF
--- a/src/main/java/net/minestom/server/chat/ColoredText.java
+++ b/src/main/java/net/minestom/server/chat/ColoredText.java
@@ -21,9 +21,9 @@ public class ColoredText extends JsonMessage {
 
     private static final char SEPARATOR_START = '{';
     private static final char SEPARATOR_END = '}';
-    private static final String COLOR_PREFIX = "#";
-    private static final String TRANSLATABLE_PREFIX = "@";
-    private static final String KEYBIND_PREFIX = "&";
+    private static final char COLOR_PREFIX = '#';
+    private static final char TRANSLATABLE_PREFIX = '@';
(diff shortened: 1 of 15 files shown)


src/main/java/net/minestom/server/chat/ColoredText.java
```diff
@@ -237,7 +235,7 @@ public JsonObject getJsonObject() {
                     continue;
                 }
                 // Translatable component
-                if (formatString.startsWith(TRANSLATABLE_PREFIX)) {
+                if (formatString.charAt(0) == TRANSLATABLE_PREFIX) {
```
KrystilizeNevaDies, line 238: Should be using .equals for string/chat comparisons.

src/main/java/net/minestom/server/command/builder/arguments/relative/ArgumentRelativeBlockPosition.java
```diff
@@ -34,7 +34,7 @@ public RelativeBlockPosition parse(@NotNull String input) throws ArgumentSyntaxE
 
         for (int i = 0; i < split.length; i++) {
             final String element = split[i];
-            if (element.startsWith(RELATIVE_CHAR)) {
+            if (element.charAt(0) == RELATIVE_CHAR) {
```
KrystilizeNevaDies, line 37: Should be using .equals for string/chat comparisons.

KrystilizeNevaDies (commented):
Should be using .equals for string/chat comparisons.

PR #133, opened 2021-02-05
[WIP] Entity Rework

LeoDog896:
Entity UUIDs are only used in specific use scenarios and not storing them is a good way to save a lot of Ram.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -465,6 +465,7 @@ public abstract class Entity implements Viewable, EventHandler, DataContainer, P
 
                 Vector newVelocityOut = new Vector();
 
+                // TODO fix "non-vanilla" like gravity
                 // Gravity force
                 final double gravityY = !hasNoGravity() ? Math.min(
                         gravityDragPerTick + (gravityAcceleration * (double) gravityTickCount),
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -54,7 +54,6 @@ import java.util.function.Consumer;
 public abstract class Entity implements Viewable, EventHandler, DataContainer, PermissionHandler {
 
     private static final Map<Integer, Entity> entityById = new ConcurrentHashMap<>();
-    private static final Map<UUID, Entity> entityByUuid = new ConcurrentHashMap<>();
     private static final AtomicInteger lastEntityId = new AtomicInteger();
 
     // Metadata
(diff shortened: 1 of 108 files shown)


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -195,8 +195,10 @@ public static int getPlayerSynchronizationTickDelay(int viewersCount) {
     // Tick related
     private final PlayerTickEvent playerTickEvent = new PlayerTickEvent(this);
 
+    protected UUID uuid;
+
     public Player(@NotNull UUID uuid, @NotNull String username, @NotNull PlayerConnection playerConnection) {
-        super(EntityType.PLAYER, uuid);
+        super(EntityType.PLAYER);
         this.username = username;
```
jglrxavpok, line 202: The player's UUID is not set

LeoDog896: Good point!

TheMode:
I don't really get the advantage, did you profile the memory usage difference?
For me it does a lot more harm than good

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -531,7 +519,7 @@ public void tick(long time) {
                     this.velocity.setX(velocity.getX() * drag);
                     this.velocity.setZ(velocity.getZ() * drag);
 
-                    if (velocity.equals(new Vector())) {
```
TheMode: This line is required to compare against Vector#getEpsilon

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -181,17 +156,18 @@ public static Entity getEntity(int id) {
...
-     * @param uuid the entity UUID
-     * @return the entity having the specified uuid, null if not found
+     * @param uuid the entity unique universal id
+     * @return the entity having the specified id, null if not found
      */
     @Nullable
-    public static Entity getEntity(@NotNull UUID uuid) {
-        return Entity.entityByUuid.getOrDefault(uuid, null);
+    public static Entity getEntity(UUID uuid) {
+        return Entity.entityById.getOrDefault((int) uuid.getLeastSignificantBits(), null);
```
TheMode, line 168: I am personally still not ok with this, it does add a bit of confusion by giving Entity and Player a different meaning of UUID

TheMode (changes requested):

src/main/java/net/minestom/server/entity/Entity.java
LeoDog896: That really _is_ the point -- An Entity's UUID is **not** the same as a Player's UUID.

TheMode: What is the difference? Without it you cannot save any entity persistently, cannot change UUID after initialization, and `Entity#getEntity(UUID)` will not work for players

LeoDog896:
Any insight into the gravity issue?

PR #139, opened 2021-02-19
Add Position PlaySound methods

LeoDog896:
Convenience methods for specifying a position instead of XYZ coordinates.


TheMode:
No problem for me, except that it would probably be preferable for those to call the already existing `playSound` methods to reduce boilerplate code

LeoDog896:
Yeah, good idea

PR #140, opened 2021-02-20
Adding has(Argument<T> argument) to Arguments.java

MrBretze:


src/main/java/net/minestom/server/command/builder/Arguments.java
```diff
@@ -49,6 +49,10 @@
         });
     }
 
+    public boolean has(@NotNull Argument<T> argument) {
```
TheMode, line 52: You should probably use the wildcard `?` instead (it doesn't compile right now)

MrBretze: Yes, i have failed this ><

PR #141, opened 2021-02-20
Fixed somme read for NBT Tag to ItemStack

MrBretze:


src/main/java/net/minestom/server/utils/NBTUtils.java
```diff
@@ -159,10 +159,11 @@ public static void loadDataIntoItem(@NotNull ItemStack item, @NotNull NBTCompoun
...
                     uuid = Utils.intArrayToUuid(uuidArray);
                 }
-                final double value = attributeNBT.getDouble("Amount");
-                final String slot = attributeNBT.getString("Slot");
+
+                final double value = attributeNBT.getNumber("Amount").doubleValue();
+                final String slot = attributeNBT.containsKey("Slot") ? attributeNBT.getString("Slot") : "MAINHAND";
                 final String attributeName = attributeNBT.getString("AttributeName");
-                final int operation = attributeNBT.getInt("Operation");
+                final int operation = attributeNBT.getInt("Operation").intValue();
```
Geolykt, line 166: Why not let autoboxing do it's thing?

src/main/java/net/minestom/server/utils/NBTUtils.java
```diff
@@ -174,11 +175,9 @@ public static void loadDataIntoItem(@NotNull ItemStack item, @NotNull NBTCompoun
                 if (attributeOperation == null) {
                     break;
                 }
+
                 final AttributeSlot attributeSlot = AttributeSlot.valueOf(slot.toUpperCase());
                 // Wrong attribute slot, stop here
-                if (attributeSlot == null) {
-                    break;
-                }
```
Geolykt, line 181: what's the motivation of removing this case?

Geolykt (commented):
The changes of the PotionMeta and NBTUtils classes do not make much sense to me.
The main contributions looks ok, but we would need to check if that can actually happen.

src/main/java/net/minestom/server/utils/NBTUtils.java
MrBretze: Oups, failed that ><

src/main/java/net/minestom/server/utils/NBTUtils.java
MrBretze: Because atribute slot is never null, i changed that because the client by default dont send the attributeSlot

TheMode: It is here because you cannot trust the client

src/main/java/net/minestom/server/utils/NBTUtils.java
```diff
@@ -163,7 +163,7 @@ public static void loadDataIntoItem(@NotNull ItemStack item, @NotNull NBTCompoun
                 final double value = attributeNBT.getNumber("Amount").doubleValue();
                 final String slot = attributeNBT.containsKey("Slot") ? attributeNBT.getString("Slot") : "MAINHAND";
                 final String attributeName = attributeNBT.getString("AttributeName");
-                final int operation = attributeNBT.getInt("Operation").intValue();
+                final int operation = attributeNBT.getNumber("Operation").intValue();
```
jglrxavpok, line 166: You can just use NBTCompound#getAsInt

See: https://github.com/jglrxavpok/Hephaistos/blob/master/src/main/kotlin/org/jglrxavpok/hephaistos/nbt/NBTCompound.kt#L223

src/main/java/net/minestom/server/item/attribute/AttributeSlot.java
```diff
@@ -6,5 +6,22 @@
...
+            case "LEGS":
+                return LEGS;
+            case "CHEST":
+                return CHEST;
+            case "HEAD":
+                return HEAD;
+            default:
+                return MAINHAND;
+        }
+    }
```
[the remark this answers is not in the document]
MrBretze: Yes the items just not giving, vanilla always set to MAINHAND by default

src/main/java/net/minestom/server/utils/NBTUtils.java
MrBretze: No, its throw a error if Operation is set to byte

jglrxavpok: Are you talking about `getInt`? Because I am talking about `get*As*Int`, which should have the exact same behaviour as your code.

PR #142, opened 2021-02-21
Custom item display of a single ItemStack for different Players

RinesThaix:
Introduces `ItemStack#setupCustomDisplay(Function<Player, ItemDisplay>)`.


TheMode:
I did make a branch some time ago to add this feature https://github.com/Minestom/Minestom/tree/item-display
But decided to do not merge it because of how much "complication" it adds.

There is also an optimization that we allow, which is writing/compress a packet only once for multiple players. I wouldn't mind such a feature but only if it were almost invisible, but I am open to discussion.

PR #146, opened 2021-02-22
Arrows and RangedAttackGoal

RinesThaix:
Arrows can be tested with `/shoot default`: it launches an arrow right from your face.
I also wouldn't recommend testing on custom stone blocks (that are turning into golden ones in demo scene), because once you remove them there's some bug in the Minestom custom blocks system and diamond block does exist despite the fact it's not shown.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -91,7 +91,8 @@ public abstract class Entity implements Viewable, EventHandler, DataContainer, P
     protected Entity vehicle;
 
     // Velocity
-    protected Vector velocity = new Vector(); // Movement in block per second
+    protected Vector  velocity   = new Vector(); // Movement in block per second
+    protected boolean hasPhysics = true;
 
     protected double gravityDragPerTick;
     protected double gravityAcceleration;
@@ -483,7 +484,11 @@ public abstract class Entity implements Viewable, EventHandler, DataContainer, P
                         getVelocity().getZ() / tps
                 );
 
-                this.onGround = CollisionUtils.handlePhysics(this, deltaPos, newPosition, newVelocityOut);
+                if (this.hasPhysics) {
+                    this.onGround = CollisionUtils.handlePhysics(this, deltaPos, newPosition, newVelocityOut);
+                } else {
+                    newVelocityOut = deltaPos;
+                }
 
                 // World border collision
                 final Position finalVelocityPosition = CollisionUtils.applyWorldBorder(instance, position, newPosition);
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -863,6 +863,16 @@ public abstract class Entity implements Viewable, EventHandler, DataContainer, P
     }
 
     /**
+     * Gets the distance squared between two entities.
+     *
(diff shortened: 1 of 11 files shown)


TheMode:
Looks amazing! (even if I do not like the formatting style you use)

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -483,7 +484,11 @@ public void tick(long time) {
                         getVelocity().getZ() / tps
                 );
 
-                this.onGround = CollisionUtils.handlePhysics(this, deltaPos, newPosition, newVelocityOut);
+                if (this.hasPhysics) {
+                    this.onGround = CollisionUtils.handlePhysics(this, deltaPos, newPosition, newVelocityOut);
+                } else {
+                    newVelocityOut = deltaPos;
```
TheMode, line 490: Shouldn't you set it to `getVelocity()` instead?

src/main/java/net/minestom/server/entity/type/projectile/EntityAbstractArrow.java
```diff
@@ -0,0 +1,161 @@
...
+
+            Chunk currentChunk = instance.getChunkAt(pos);
+            if (currentChunk != chunk) {
+                chunk = currentChunk;
+                entities = instance.getChunkEntities(chunk)
+                        .stream()
+                        .filter(entity -> entity instanceof LivingEntity)
+                        .collect(Collectors.toSet());
+            }
+            if (getAliveTicks() < 3) {
```
TheMode, line 93: Could you add a comment as to why it is necessary? (maybe also in other part of the code, as it is pretty hard to get at first glance)

src/main/java/net/minestom/server/entity/Entity.java
RinesThaix: Look at the `hasPhysics` branch: this variable is being passed to `handlePhysics` to be populated there.

src/main/java/net/minestom/server/entity/type/projectile/EntityAbstractArrow.java
RinesThaix: Because arrows spawns in the shooter and without this check it will instantly damage him. And we also can't exclude shooter from the targets list, because you shooter should be able to launch an arrow vertically up and then take damage from it.

src/main/java/net/minestom/server/entity/Entity.java
TheMode: `hasPhysics` branch? Where is it?
I meant that `newVelocityOut` should probably be set to `getVelocity()` because the vector is copied later

RinesThaix: I was talking about if branch 2 lines up :)

TheMode: Oh right. I am saying this because `deltaPos` represents the movement delta and not the velocity out

RinesThaix: Yes, but later we're setting `velocity` to `deltaPos`, multiplied by `tps`, so it seems absolutely correct.

TheMode: My bad, forgot about it

PR #147, opened 2021-02-22
AbstractProjectile and arrows now call EntityAttackEvent, without damaging entities

RinesThaix:


src/main/java/net/minestom/server/entity/type/projectile/AbstractProjectile.java
```diff
@@ -0,0 +1,25 @@
...
+
+import net.minestom.server.entity.Entity;
+import net.minestom.server.entity.EntityType;
+import net.minestom.server.entity.ObjectEntity;
+import net.minestom.server.entity.type.Projectile;
+import net.minestom.server.utils.Position;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public abstract class AbstractProjectile extends ObjectEntity implements Projectile {
```
RinesThaix, line 11: All projectiles seem to be `ObjectEntities`, so I guess it's ok to force them to inherit it.

PR #151, opened 2021-02-22
New Entity AI

RinesThaix:
1. First of all, allows to have multiple `GoalSelector`s running at the same time, because now an entity can have multiple independent `EntityAIGroup`s: each with own goal and target selectors. It could be usable, for example, in case of an entity having goal selectors `MeleeAttackGoal` and `SwimGoal` (that allows entity to not be drowned), which must be handled simultaneously when an entity is in water.
2. Secondly, previous implementation had an issue: whether some `GoalSelector` was selected, no one else could be chosen before the current one ends, even if there are some other goals with higher priority (i.e. were added to the goal selectors list earlier) that should start.


src/main/java/net/minestom/server/entity/ai/EntityAIGroup.java
```diff
@@ -0,0 +1,143 @@
...
+    public void setCurrentGoalSelector(@Nullable GoalSelector goalSelector) {
+        Check.argCondition(
+                goalSelector != null && goalSelector.getAIGroup() != this,
+                "Tried to set goal selector attached to another AI group!"
+        );
+        this.currentGoalSelector = goalSelector;
+    }
+
+    public void tick(long time) {
+        GoalSelector currentGoalSelector = getCurrentGoalSelector();
```
jglrxavpok, line 70: When possible, try to avoid reusing the same name than a field.
It's by no mean a big issue but it did confuse me a bit a few lines down.

jglrxavpok (approved):
Looks good to me

PR #152, opened 2021-02-23
EntityMeta

RinesThaix:
1. Changed `EntityType` generator so that it includes base width and height of an entity.
2. Now there's an `Entity#getEntityMeta()` method to retrieve `EntityMeta`. It allows developers to modify entity meta independently of which `Entity` class implementation was used for an entity instantiation. So, if developer wants to have ai+nav for an armor stand, he can instantiate it as an `EntityCreature`; if developer doesn't want to have features on liveness and ai+nav on a chicken, he can instantiate it as just an `Entity`.
3. Added `ArgumentEnum` parser for commands.

Can be tested with `/horse` command on demo scene.


src/autogenerated/java/net/minestom/server/entity/EntityType.java
```diff
@@ -10,51 +31,51 @@
  */
 @SuppressWarnings({"deprecation"})
 public enum EntityType {
```
TheMode, line 33: Due to the nature of those enums (retrieved from exported data), they are autogenerated from there: https://github.com/Minestom/Minestom/tree/master/src/generators/java/net/minestom/codegen
It seems that you changed the constructor manually.

Additionally, I am not sure about defining default entities class as this could lead to some unexpected behavior when created an entity directly extending EntityCreature (or just changing the entity type after its instantiation)

RinesThaix: I wanted to introduce some sort of `instantiate an Entity from EntityType for me`, but I agree that this way may be not the best one.
Do you think it would be better to completely remove this or just to move to separate file?

TheMode: I do not think that this is a terrible idea but might not work well enough when you need more customization. It could also appeal developers to check for entity type using `instanceof` (which should never be the case).

I did think multiple time about moving all the entities into an external repo, if this happens, instantiating an entity from its `EntityType` would be logical as a shortcut (but also indicate that those are not "official")

Nothing is set in stone yet

PR #154, opened 2021-02-28
Added CanPlaceOn and CanDestroy for ItemStacks

Kebab11noel:
Added CanPlaceOn and CanDestroy properties for ItemStacks, also implemented checks for them in the corresponding listeners

### DIFF
--- a/src/main/java/net/minestom/server/item/ItemStack.java
+++ b/src/main/java/net/minestom/server/item/ItemStack.java
@@ -67,6 +67,9 @@ public class ItemStack implements DataContainer, PublicCloneable<ItemStack> {
     private StackingRule stackingRule;
     private Data data;
 
+    private Set<String> canDestroy;
+    private Set<String> canPlaceOn;
+
     {
         if (defaultStackingRule == null)
             defaultStackingRule = VANILLA_STACKING_RULE;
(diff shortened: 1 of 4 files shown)


TheMode:
Would probably be better to use `Collection#isEmpty` instead of checking the size

Kebab11noel:
Will change it, also I forgot about `net.minestom.server.item.ItemStack#isSimilar`

src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -195,6 +203,76 @@ public boolean equals(Object o) {
...
+     */
+    public void removeCanPlaceOn(String block) {
+        canPlaceOn.remove(block);
+    }
+
+    /**
+     * Gets the blocks that this item can be placed on
+     * @return an unmodifiable {@link Set} of blocks
+     */
+    public Set<String> getCanPlaceOn() {
```
TheMode, line 238: If you want to have a way to add/remove and retrieve the whole `Set`, you could as well have a single method returning a modifiable collection

Kebab11noel: I introduced it because `net.minestom.server.utils.NBTUtils#saveDataIntoNBT` had to get the whole set, but kept the other methods to provide a shorter way of doing additions and removals, but I can change it if you want to.

Issue #156, opened 2021-03-01
Implement the Adventure API


kezz:
Adventure is a server-side user interface library for Minecraft which is fast becoming the standard user interface system for many Minecraft related projects. It is currently implemented in [Velocity](https://velocitypowered.com/), [Sponge 8](https://www.spongepowered.org/) and [Paper](https://papermc.io/).

Moving to use Adventure has the following benefits:
- Brings Minestom into step with other server implementations.
- Allows developers to move projects to Minestom more easily.
- Reduces all user interface maintenance in Minestom.
- Future-proofs Minestom developers against upcoming changes (for example, the upcoming Title changes).
- Allows developers to use the advanced API that Adventure provides.

mworzala:
This has been discussed in Discord many times, and as far as I am aware there has been no change in stance on the subject.

I think it's likely that this is a waste of time at the moment. That being said, [this](https://github.com/mworzala/adventure-platform-minestom) may be a helpful reference.

TheMode:
I personally still do not get the advantage over a wrapper as the one @mworzala did.
But well I get the concern and the fact that it will help to reduce maintenance

kezz:
Closed with #157

PR #157, opened 2021-03-01
Adventure Time!

### LINKED ISSUE #159: MapMeta uses the wrong colour format
The `MapMeta` class incorrectly sends colours using the legacy colour code rather than an RGB int.

### LINKED ISSUE #161: ChatHoverEvent#showEntity and #showItem do not render correct output
### What is the current behavior?
Currently, creating a `ChatHoverEvent` for an entity or an item will produce the following output:
```json
"hoverEvent": {
    "action": "show_entity",
    "value": "{\"id\":\"7e59f8e9-68ed-4fc4-b632-6e3854e32250\",\"type\":\"minecraft:bee\"}"
}
```

The same is also the case for the `show_item` action, which serialises the NBT of the whole item rather than just the tag (if any).

### What is the expected behavior?
The correct formatting for hover events with arbitrary NBT data is as follows:
```json
"hoverEvent": {
    "action": "show_entity",
    "contents": {
        "type": "minecraft:bee",
        "id": "7e59f8e9-68ed-4fc4-b632-6e3854e32250"
    }
}
```

### What steps will reproduce the problem?
Simply create a `ChatHoverEvent` and call `toString` on the event. This can be compared to hover events produced by online tools such as https://www.minecraftjson.com/.

### What operating system and java version is being used?
Ubuntu 20.04 and Java 15 (replicated on Java 11).

### Additional information / Possible thoughts on why this bug is happening
The reliance on `NBTCompound#toSNBT` is incorrect as the compound formatting differs to the string version of NBT data. This should be swapped to constructed JSON elements instead.

kezz:
# Adventure Time!
This PR represents progress towards using [Adventure](https://github.com/KyoriPowered/adventure) as standard in Minestom.

## Why Adventure?
Adventure is a server-side user interface library for Minecraft which is fast becoming the standard user interface system for many Minecraft related projects. It is currently implemented in [Velocity](https://velocitypowered.com/), [Sponge 8](https://www.spongepowered.org/) and [Paper](https://papermc.io/).

Moving to use Adventure has the following benefits:
- Brings Minestom into step with other server implementations.
- Allows developers to move projects to Minestom more easily.
- Reduces all user interface maintenance in Minestom.
- Future-proofs Minestom developers against upcoming changes (e.g. upcoming title changes).
- Allows developers to use the advanced API that Adventure provides.
- Simplifies all user interface processes (see the [Audiences](#audiences) section below).
- Built-in localisation (or other component translation interception functionality) framework.
- Ability to use Adventure's Block, Entity and Storage NBT components, in addition to the score, team and statistic components.
- Ability to use Adventure's replacement (easily perform component replacements on a component) and joining (concatenating two different components) features.
- We get to include Minestom in the platforms section of the Adventure docs, thus providing some great free advertising :wink:.

For discussion on implementing Adventure into Minestom, see #156.

### Audiences
This PR also adds a `asAudience` method to the `Viewable` interface. This allows developers to send messages, show books or send boss bars to every player who can view any object that implements `Viewable`. 

This is a powerful addition that will greatly improve the usefulness of the `Viewable` interface, allowing developers to create their own custom `Audience`s with ease.

## Progress
In order for this PR to become feature complete, the following changes need to be made:
- [x] Deprecate chat package.
- [x] Implement Audience where applicable. 
  - [x] `Player`
  - [x] `ConsoleSender`
  - [x] `Team`
  - [x] `Instance`
  - [x] `MinecraftServer`
  - [x] `ConnectionManager`
  - [x] `Viewable`
- [x] Deprecate Player UI methods.
- [x] Implement Adventure methods.
  - [x] `Player`
    - [x] Boss bars
    - [x] Books
    - [x] Action bar
    - [x] Sounds
    - [x] Messages
  - [x] `ConsoleSender`
    - [x] Messages
  - [x] `Team`
    - [x] Forwarding audiences
  - [x] `Instance`
    - [x] Forwarding audiences
  - [x] `MinecraftServer`
    - [x] Forwarding audiences
  - [x] `ConnectionManager`
    - [x] Forwarding audiences 
- [x] Deprecate boss bar package.
- [x] Deprecate sound package.
- [x] Add converter to/from `Component` into `JsonMessage`.
- [x] Ensure implementation is thread safe.
- [x] Add ability to swap serializer.
- [x] Serialize `JsonMessage`s using Adventure's Gson instances. 
- [x] Use locale-based serialisation methods.
- [x] Make methods returning/accepting `JsonMessage` use `Component`.
- [x] Replace internal deprecated usages with Adventure alternatives.
- [x] Update demo and tests.

# Related Issues
This PR closes #157, fixes #159 and fixes #161.


build.gradle
```diff
@@ -179,6 +179,13 @@ dependencies {
 
 
     generatorsImplementation("com.squareup:javapoet:1.13.0")
+
+    // Adventure, for text messages
+    api platform("net.kyori:adventure-bom:4.5.1")
+    api "net.kyori:adventure-api"
+    implementation "net.kyori:adventure-text-serializer-gson"
+    implementation "net.kyori:adventure-text-serializer-plain"
+    implementation "net.kyori:adventure-text-serializer-legacy"
```
kashike, line 188: It might be worth exposing these - they are exposed as API with Sponge, Velocity, and Paper.

src/main/java/net/minestom/server/chat/JsonMessage.java
```diff
@@ -14,6 +16,7 @@
  *
  * @see <a href="https://wiki.vg/Chat">Chat Format</a>
  */
+@Deprecated
 public abstract class JsonMessage {
```
kashike, line 20: This can implement `ComponentLike` - provides `Component asComponent()`.

build.gradle
```diff
@@ -179,6 +179,13 @@ dependencies {
 
 
     generatorsImplementation("com.squareup:javapoet:1.13.0")
+
+    // Adventure, for text messages
+    api platform("net.kyori:adventure-bom:4.5.1")
```
kashike, line 184:
```suggestion
    api platform("net.kyori:adventure-bom:4.6.0")
```

kashike (changes requested):

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -62,7 +66,7 @@
  * The server needs to be initialized with {@link #init()} and started with {@link #start(String, int)}.
  * You should register all of your dimensions, biomes, commands, events, etc... in-between.
  */
-public final class MinecraftServer {
+public final class MinecraftServer implements ForwardingAudience.Single {
```
kashike, line 69:
```suggestion
public final class MinecraftServer implements ForwardingAudience {
```

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -804,4 +832,8 @@ private static void checkInitStatus(@Nullable Object object) {
                         "if you are developing an extension be sure to retrieve them at least after Extension#preInitialize");*/
     }
 
+    @Override
+    public @NotNull Audience audience() {
+        return getConnectionManager();
+    }
```
kashike, line 838: Should be implementing `Iterable` variant (with above interface change), and return an `Iterable` consisting of all online players and the server console.

kashike (changes requested):

src/main/java/net/minestom/server/adventure/AdventureUtils.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Attempts to get an NMS sound from an Adventure key.
+     *
+     * @param name the key
+     *
+     * @return the sound, if found
+     */
+    public static @Nullable Sound asSound(@NotNull Key name) {
+        return SOUND_MAP.get(name.asString());
+    }
```
jglrxavpok, line 28: Is there a reason not to use Registries#getSound ?

src/main/java/net/minestom/server/chat/ChatClickEvent.java
```diff
@@ -5,6 +5,7 @@
 /**
  * Represents a click event for a specific portion of the message.
  */
+@Deprecated
```
jglrxavpok, line 8: Why deprecate this event?

src/main/java/net/minestom/server/chat/ChatHoverEvent.java
```diff
@@ -1,15 +1,24 @@
...
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jglrxavpok.hephaistos.nbt.NBTCompound;
 
+import java.util.UUID;
+
 /**
  * Represents a hover event for a specific portion of the message.
  */
+@Deprecated
```
jglrxavpok, line 21: Same as ChatClickEvent

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -387,12 +387,12 @@ public boolean damage(@NotNull DamageType type, float value) {
...
-                SoundCategory soundCategory;
+                Source soundCategory;
                 if (this instanceof Player) {
-                    soundCategory = SoundCategory.PLAYERS;
+                    soundCategory = Source.PLAYER;
                 } else {
                     // TODO: separate living entity categories
-                    soundCategory = SoundCategory.HOSTILE;
+                    soundCategory = Source.HOSTILE;
                 }
```
jglrxavpok, line 397: Why are you replacing the Minestom SoundCategory with Source?
Because this is linked to the MC protocol, I believe Minestom should keep its own enum.
Adding a `toAdventureSource()` to the Sound enum could help porting.

src/main/java/net/minestom/server/network/packet/server/play/ChatMessagePacket.java
```diff
@@ -35,9 +81,29 @@ public int getId() {
...
+
+        /**
+         * Gets this position as an Adventure message type. Note this will return
+         * {@code null} for {@link #GAME_INFO} as it is preferred to use
+         * {@link TitlePacket} due to <a href="https://bugs.mojang.com/browse/MC-119145">MC-119145</a>.
+         * @return the message type
+         */
+        public @Nullable MessageType asMessageType() {
+            return this.messageType;
+        }
```
jglrxavpok, line 107: The MC bug is marked as resolved, is the null return still necessary?

src/main/java/net/minestom/server/network/packet/server/play/DisconnectPacket.java
```diff
@@ -7,16 +7,32 @@
 import org.jetbrains.annotations.NotNull;
 
 public class DisconnectPacket implements ServerPacket {
+    public String message;
 
-    public JsonMessage message; // Only text
+    /**
+     * @deprecated Use {@link #message}
+     */
+    @Deprecated public JsonMessage messageJson;
```
jglrxavpok, line 15: If you are replacing a JsonMessage by a String, do delete the JsonMessage field, there is no point in keeping it.
You are already breaking the API by changing the type, go all the way!

Same goes for other JsonMessages inside other packet classes.

src/main/java/net/minestom/server/network/packet/server/play/EntitySoundEffectPacket.java
```diff
@@ -9,15 +10,20 @@
...
-    public SoundCategory soundCategory;
+    public Sound.Source soundSource;
     public int entityId;
     public float volume;
     public float pitch;
 
+    /**
+     * @deprecated Use {@link #soundSource}
+     */
+    @Deprecated public SoundCategory soundCategory;
```
jglrxavpok, line 21: Same as JsonMessages in earlier comment: either keep Minestom type or delete it completly. 
But in this case, I prefer keeping SoundCategory, but I am open to debate.

src/main/java/net/minestom/server/network/packet/server/play/NamedSoundEffectPacket.java
```diff
@@ -9,15 +9,20 @@
 public class NamedSoundEffectPacket implements ServerPacket {
 
     public String soundName;
-    public SoundCategory soundCategory;
+    public int soundCategory;
```
jglrxavpok, line 12: Either use Source or SoundCategory, but don't replace an enum by an int inside packets!
We can use Java's compiler to ensure the given category is valid thanks to types.

src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
```diff
@@ -11,15 +12,24 @@
 public class SoundEffectPacket implements ServerPacket {
 
     public int soundId;
-    public SoundCategory soundCategory;
+    public int soundCategory;
```
jglrxavpok, line 15: Same as earlier comment: don't replace an enum type to its integral representation.

src/main/java/net/minestom/server/network/packet/server/play/TeamsPacket.java
```diff
@@ -39,20 +40,38 @@
...
+    /**
+      @deprecated Use {@link #teamPrefix}
+     */
+    @Deprecated public JsonMessage teamPrefixJson;
+    /**
+      @deprecated Use {@link #teamSuffix}
+     */
+    @Deprecated public JsonMessage teamSuffixJson;
+
     /**
```
jglrxavpok, line 75: You can delete all these.

src/main/java/net/minestom/server/sound/SoundCategory.java
```diff
@@ -1,5 +1,9 @@
 package net.minestom.server.sound;
 
+/**
+ * @deprecated Use {@link net.kyori.adventure.sound.Sound.Source}
+ */
+@Deprecated
 public enum SoundCategory {
```
jglrxavpok, line 7: Is there a reason why?
Also, does Adventure enum respects the order the MC protocol expects the values?

jglrxavpok (commented):
I am trusting you on the Adventure implementation front.

I have a few issues with the PR in its actual state, mostly that you deprecate parts of Minestom instead of simply replacing them with something new. Minestom is not stable yet, so you can afford to do it!
For my other remarks, check the comments I left during the review.

src/main/java/net/minestom/server/chat/ChatClickEvent.java
kezz: Both this event and `ChatHoverEvent` are replaced with Adventure's `HoverEvent` and `ClickEvent`, these events are deprecated following the deprecation of `JsonMessage` and the current `chat` package.

src/main/java/net/minestom/server/entity/LivingEntity.java
kezz: Yes this is a fair point! I suppose the reason why I swapped these is because Adventure's `Source` is an exact replica of Minestom's `SoundCategory` enum. I didn't think there was much point having them both, especially considering that `Source` is used to play sounds, etc. Happy to swap back to `SoundCategory` if you feel it's best.

src/main/java/net/minestom/server/network/packet/server/play/DisconnectPacket.java
kezz: I was keeping the old fields in order to maintain compatibility - in the writing methods it will write the `messageJson` if it exists. I'm happy to remove all these if needs be but I think it might be better to remove them alongside the removal of other deprecated methods at a later point in order to give people a bit of time to change over.

src/main/java/net/minestom/server/network/packet/server/play/NamedSoundEffectPacket.java
kezz: So the main reason that I removed a lot of the enums from these packets was to make the packets more raw, thus allowing for further customisation and control by developers. For example, modded sound categories would require changing the enum but with this method they could be done on the fly if needs be. I suppose the enum could always be edited, but there are many places where raw values are taken in the packet.

src/main/java/net/minestom/server/network/packet/server/play/ChatMessagePacket.java
kezz: Well Adventure doesn't support `sendMessage` with a `MessageType` that indicates the message should be sent in the game info position. Following the reasons mentioned below, this `MessageType` enum should actually be switched to an int describing the message type. I'll wait for your response about the switch from enum to int.

src/main/java/net/minestom/server/sound/SoundCategory.java
kezz: I'll leave discussion of the deprecation of SoundCategory to the other thread, but this does raise the point that Adventure doesn't *guarantee* the ordering - so I'll some methods to the utils class.

src/main/java/net/minestom/server/adventure/AdventureUtils.java
kezz: Not really - in fact, this can probably be moved into the sound enum directly.

src/main/java/net/minestom/server/network/packet/server/play/NamedSoundEffectPacket.java
kezz: As per discussion in Discord, this has been reverted to use enum.

src/main/java/net/minestom/server/network/packet/server/play/ChatMessagePacket.java
kezz: Resolved as per other discussions.

src/autogenerated/java/net/minestom/server/entity/EntityType.java
```diff
@@ -352,6 +354,8 @@
     @NotNull
     private final EntitySpawnType spawnType;
 
+    private Key key;
```
kashike, line 357:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/fluids/Fluid.java
```diff
@@ -20,11 +22,14 @@
 
     LAVA("minecraft:lava");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 27:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/instance/block/Block.java
```diff
@@ -2480,25 +2482,27 @@
...
     @Nullable
-    private NamespaceID blockEntity;
+    private final NamespaceID blockEntity;
 
-    private boolean singleState;
+    private final boolean singleState;
 
     private List<BlockAlternative> alternatives = new java.util.ArrayList<>();
 
+    private Key key;
```
kashike, line 2504:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/item/Enchantment.java
```diff
@@ -86,11 +88,14 @@
 
     VANISHING_CURSE("minecraft:vanishing_curse");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 93:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/item/Material.java
```diff
@@ -1966,19 +1968,22 @@
...
+    private final String namespaceID;
 
-    private int maxDefaultStackSize;
+    private final int maxDefaultStackSize;
 
     @Nullable
-    private Block correspondingBlock;
+    private final Block correspondingBlock;
+
+    private Key key;
```
kashike, line 1978:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/particle/Particle.java
```diff
@@ -154,11 +156,14 @@
 
     WHITE_ASH("minecraft:white_ash");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 161:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/potion/PotionEffect.java
```diff
@@ -74,11 +76,14 @@
 
     HERO_OF_THE_VILLAGE("minecraft:hero_of_the_village");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 81:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/potion/PotionType.java
```diff
@@ -96,11 +98,14 @@
 
     LONG_SLOW_FALLING("minecraft:long_slow_falling");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 103:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/sound/Sound.java
```diff
@@ -1994,11 +1996,14 @@
 
     ENTITY_ZOMBIE_VILLAGER_STEP("minecraft:entity.zombie_villager.step");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 2001:
```suggestion
    private final Key key;
```

src/autogenerated/java/net/minestom/server/stat/StatisticType.java
```diff
@@ -158,11 +160,14 @@
 
     INTERACT_WITH_SMITHING_TABLE("minecraft:interact_with_smithing_table");
 
-    private String namespaceID;
+    private final String namespaceID;
+
+    private Key key;
```
kashike, line 165:
```suggestion
    private final Key key;
```

kashike (changes requested):

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2675,6 +2677,11 @@ public void setLocale(@Nullable Locale locale) {
         settings.locale = locale == null ? null : locale.toLanguageTag();
     }
 
+    @Override
+    public @NonNull Identity identity() {
+        return Identity.identity(this.uuid);
```
kashike, line 2680: Should cache the `Identity` in a field, ideally, as this can end up being called often and is wasteful to create a new one each time.

kashike (changes requested):

build.gradle
```diff
@@ -184,11 +184,10 @@ dependencies {
     generatorsImplementation("com.squareup:javapoet:1.13.0")
 
     // Adventure, for text messages
-    api platform("net.kyori:adventure-bom:4.7.0")
```
kashike: `api platform("net.kyori:adventure-bom:$adventureVersion")` and then keep the rest the same?

kezz: Yeah I had this with the BOM originally but it was causing problems developing projects that depend on Minestom. I think there's something else wrong with this build file but I couldn't quite work it out and this seemed to fix it, although it's not the best solution that's for sure.

src/main/java/net/minestom/server/network/packet/server/play/NamedSoundEffectPacket.java
```diff
@@ -9,15 +12,20 @@
...
-    public SoundCategory soundCategory;
+    public Source soundSource;
     public int x, y, z;
     public float volume;
     public float pitch;
 
+    /**
+     * @deprecated Use {@link #soundSource}
+     */
+    @Deprecated public SoundCategory soundCategory;
```
jglrxavpok, line 23: No longer used, remove

src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
```diff
@@ -11,15 +13,29 @@
...
+    public Source soundSource;
     public int x, y, z;
     public float volume;
     public float pitch;
 
-    public static SoundEffectPacket create(SoundCategory category, Sound sound, Position position, float volume, float pitch) {
+    /**
+     * @deprecated Use {@link #soundSource}
+     */
+    @Deprecated public SoundCategory soundCategory;
```
jglrxavpok, line 24: No longer used, remove

src/main/java/net/minestom/server/utils/NamespaceID.java
```diff
@@ -62,6 +63,10 @@ public static NamespaceID from(String id) {
         return from(getDomain(id), getPath(id));
     }
 
+    public static NamespaceID from(Key key) {
```
jglrxavpok, line 66: NamespaceID could use a `asKey()` function.

kezz: Great idea - they can implement `Keyed` actually which would allow them to be used directly throughout the API.

mworzala (approved):
Pretty minor, I am not super concerned if they are done or not.

PR #162, opened 2021-03-06
Fixing projectiles gravity

RinesThaix:


src/main/java/net/minestom/server/entity/type/projectile/EntityProjectile.java
```diff
@@ -31,19 +31,22 @@
...
+        setup();
     }
 
     private void setup() {
         super.hasPhysics = false;
         if (getEntityMeta() instanceof ProjectileMeta) {
             ((ProjectileMeta) getEntityMeta()).setShooter(this.shooter);
         }
+        setGravity(0.02f, 0.04f, 1.96f);
         setBoundingBox(.5F, .5F, .5F);
```
TheMode, line 50: I didn't check this previously, but any reason why the bounding box is hardcoded instead of using `EntityType`?

PR #174, opened 2021-03-17
Gravity fix

Kebab11noel:
This PR makes gravity calculation vanilla like.


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -133,6 +132,84 @@ public Entity(@NotNull EntityType entityType, @NotNull UUID uuid) {
 
         setAutoViewable(true);
 
+        // Set default gravity values
+        // according to https://minecraft.gamepedia.com/Entity#Motion_of_entities
+        double gravityAcceleration;
```
TheMode, line 137: If it depends on the entity type, it might be better to put it in the enum to be easily accessible

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -532,9 +609,17 @@ public void tick(long time) {
                 Vector newVelocityOut = new Vector();
 
                 // Gravity force
-                final double gravityY = !hasNoGravity() ? Math.min(
-                        gravityDragPerTick + (gravityAcceleration * (double) gravityTickCount),
-                        gravityTerminalVelocity) : 0;
+                double gravityY;
```
TheMode, line 612: I'd suggest creating a 'protected getGravityValue()' with this code, so it is properly separated from the rest and can be overridden to allow custom gravity algorithms

Kebab11noel: That's a good idea, moved it to a method (and left only the has gravity check there).

TheMode:
Looks great for me!

src/main/java/net/minestom/server/entity/Entity.java
Kebab11noel: Okay, moved it in commit 0d2dc80

TheMode (approved):

PR #177, opened 2021-03-22
Chunk Light API

Zanctarian:
I'm currently working on the Chunk Light API / Engine implementation. The API is implemented for the most part, though could always use feedback in terms of optimizations. 

A potential error is certain chunks not rendering the lights at all, but more experimentation will have to be done to confirm. 

An example use for the API as it is now is (naive example):
`
instanceContainer#setChunkLightEngine(ChunkLightEngine.class);
`

And for a sample with the engine, refer to [here](https://github.com/Zanctarian/Minestom/blob/chunk-light-api/src/main/java/net/minestom/server/instance/lighting/DefaultChunkLightEngine.java)
Apologies in advance for formatting issues, not sure why it's spaced so crudely.


src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
```diff
@@ -0,0 +1,155 @@
...
+        int coordinateIndex = getIndex(x, y, z);
+        int positionIndex = coordinateIndex >> 1;
+
+        if(isFirst(coordinateIndex))
+            currentLightArray[positionIndex] = (byte)(currentLightArray[positionIndex] & 0xF0 | lightLevel & 0xF);
+        else
+            currentLightArray[positionIndex] = (byte)(currentLightArray[positionIndex] & 0xF | (lightLevel & 0xF) << 4);
+
+        // Ensure the array is in the list
+        currentLightArrayList.set(y, currentLightArray);
```
Zanctarian, line 86: Just realized I don't need this anymore. Before using Arrays.asList, I was doing a more dynamic approach.. but forgot to remove this as well.

src/main/java/net/minestom/server/instance/Chunk.java
```diff
@@ -81,11 +84,15 @@
     // Data
     protected Data data;
 
+    // Generators
+    protected ChunkLightEngine chunkLightEngine;
```
mworzala, line 88: I don't think that we should allow different light engines per chunk. That seems like it would create very odd and undesirable behavior.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
```diff
@@ -0,0 +1,155 @@
...
+import net.minestom.server.MinecraftServer;
+import net.minestom.server.instance.Chunk;
+import net.minestom.server.network.packet.server.play.UpdateLightPacket;
+
+/**
+ * Configure chunk light data.
+ * <p>
+ * Notably used in {@link ChunkLightEngine#lightChunk(ChunkLight)} when implementing {@link ChunkLightEngine}
+ */
+public class ChunkLight {
```
mworzala, line 16: This class shouldn't exist in my opinion. It should just be a property of every Chunk and should only live there. 

In a more ideal world, this would be per chunk section, but there is some consideration of how this will be handled in current and future versions (1.17)

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
```diff
@@ -0,0 +1,155 @@
...
+
+    // Data
+    final Chunk chunk;
+
+    // List of sky's light levels
+    List<byte[]> skyLight;
```
mworzala, line 22: Efficiency wise, this is incorrect.

The data should be stored as nibbles, or 4 bits each. We only need 4 bits to represent the 16 light values, and should only use that. This means that block and skylight can each be represented by an array of 2048 bytes for each chunk section.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public int getChunkX() {
+        return chunk.getChunkX();
+    }
+
+    public int getChunkZ() {
+        return chunk.getChunkZ();
+    }
+
+    public UpdateLightPacket convertDataIntoPacket() {
```
mworzala, line 97: This should be cached as long as there aren't changes to section light.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
```diff
@@ -0,0 +1,155 @@
...
+     * @param y Chunk block Y
+     * @param z Chunk block Z
+     * @param skyLightLevel Light emission level from 0 to 15
+     */
+    public void setSkyLight(int x, int y, int z, int skyLightLevel) {
+        setLight(x, y, z, skyLightLevel, false);
+    }
+
+    // Sky and block uses same exact parser
+    private void setLight(int x, int y, int z, int lightLevel, boolean isBlock) {
```
mworzala, line 64: I am unsure whats happening here. `y` is being used for an index in the chunk section as well as the chunk section itself. This seems incorrect, unless I am missing something.

src/main/java/net/minestom/server/instance/Chunk.java
```diff
@@ -399,28 +406,19 @@ public ChunkDataPacket getFreshPartialDataPacket() {
...
-        for (int i = 0; i < 14; ++i) {
-            temp.add(bytes);
-        }
-        for (int i = 0; i < 6; ++i) {
-            temp2.add(bytes);
-        }
-        updateLightPacket.skyLight = temp;
-        updateLightPacket.blockLight = temp2;
+        ChunkLight lightData = new ChunkLight(this);
+        chunkLightEngine.lightChunk(lightData);
```
mworzala, line 410: Light should not be calculated when the packet is needed, it should be completely generated when the chunk is generated, and then only updated when lights are added or removed.

mworzala:
[This](https://www.seedofandromeda.com/blogs/29-fast-flood-fill-lighting-in-a-blocky-voxel-game-pt-1) is an extremely helpful article on how to do light propagation.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
Zanctarian: So far, in 1.17, the only major change according to wiki.vg is just the addition of array length-- but I don't know much more than that. The idea of making the class specific to sections isn't a bad idea either, I'll make the adjustment.. however, I do believe we should keep it as a wrapper for readability sake and for ease of use.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
Zanctarian: This class was originally meant to be per-chunk instead of per-chunk-section. The List is meant for the data in EACH chunk section, so each byte array is 2048 bytes in length, with each index representing a chunk section in a chunk.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
Zanctarian: That's my bad on my part. I'll be sure to make the adjustment to cache the packet.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
Zanctarian: This may of been a blunder on my part.

The class was originally meant to handle per-chunk instead of per-chunk-section. With that stated, my intention was to read a raw coordinate (-1,256) and get the chunk section and chunk coordinate Y based on the info. So 33 would be index 2 at a Y value of 1.

src/main/java/net/minestom/server/instance/Chunk.java
Zanctarian: As stated in the cache comment, I'll be sure to make the adjustment.

src/main/java/net/minestom/server/instance/Chunk.java
Zanctarian: > [This](https://www.seedofandromeda.com/blogs/29-fast-flood-fill-lighting-in-a-blocky-voxel-game-pt-1) is an extremely helpful article on how to do light propagation.

Thank you for the resource! I'll make sure to look into it once the API gets sorted out.

Zanctarian:
> [This](https://www.seedofandromeda.com/blogs/29-fast-flood-fill-lighting-in-a-blocky-voxel-game-pt-1) is an extremely helpful article on how to do light propagation.

Thank you for the resource! I'll be sure to check it out!

src/main/java/net/minestom/server/instance/Chunk.java
Zanctarian: Ah, I didn't notice changing the light engine per chunk. I'll make adjustments so it's just per-instance.

src/main/java/net/minestom/server/instance/lighting/ChunkLight.java
mworzala: There is some question of how much Minestom will change to use sections over chunks, and this will depend somewhat on that.

In terms of keeping it as a separate class, I am curious of other's opinions, but I feel it fits with the rest of the chunk style choices for it to be a property of Chunk. 

cc @TheMode

src/main/java/net/minestom/server/instance/Chunk.java
```diff
@@ -400,25 +399,18 @@ public ChunkDataPacket getFreshPartialDataPacket() {
...
-     * @param chunkLightEngine The ChunkLightEngine implementation
+     * @param chunk the current chunk
+     * @return the chunk light data for a given chunk
      */
     @NotNull
-    public void setChunkLightEngine(@Nullable ChunkLightEngine chunkLightEngine) {
-        this.chunkLightEngine = chunkLightEngine;
+    public UpdateLightPacket getChunkLightPacket()
+    {
+        ChunkLightEngine engine = chunkLightEngine;
```
Zanctarian, line 410: forgot to remove this, as it's very unneeded. Was doing something different at the time, but forgot about this.

KrystilizeNevaDies:
Closing as this is very outdated.

PR #180, opened 2021-03-23
Extension exposing and cleanup

LeoDog896:
In order for other plugins to manage extensions better, they need to know more about them, and this exposes the internal DiscoveredExtension object instead of the old ExtensionDescription.

### DIFF
--- a/src/main/java/net/minestom/server/extensions/DiscoveredExtension.java
+++ b/src/main/java/net/minestom/server/extensions/DiscoveredExtension.java
@@ -11,24 +11,59 @@ import java.net.URL;
 import java.util.LinkedList;
 import java.util.List;
 
+/**
+ * Represents an extension from an `extension.json` that is capable of powering an Extension object.
+ *
+ * This has no constructor as its properties are set via GSON.
+ */
 public final class DiscoveredExtension {
 
-    public final static Logger LOGGER = LoggerFactory.getLogger(DiscoveredExtension.class);
+    /** Static logger for this class. */
+    public static final Logger LOGGER = LoggerFactory.getLogger(DiscoveredExtension.class);
 
+    /** The regex that this name must pass. If it doesn't, it will not be accepted. */
     public static final String NAME_REGEX = "[A-Za-z][_A-Za-z0-9]+";
+
+    /** Name of the DiscoveredExtension. Unique for all extensions. */
     private String name;
+
+    /** Main class of this DiscoveredExtension, must extend Extension. */
     private String entrypoint;
+
+    /** Version of this extension, highly reccomended to set it. */
     private String version;
+
+    /** Points to sponge mixin config in resources folder. */
     private String mixinConfig;
+
+    /** People who have made this extension. */
     private String[] authors;
+
+    /** All code modifiers (the classes they point to) */
     private String[] codeModifiers;
+
+    /** List of extension names that this depends on. */
     private String[] dependencies;
+
+    /** List of Repositories and URLs that this depends on. */
     private ExternalDependencies externalDependencies;
+
+    /** A list of any missing code modifiers to be used for logging. */
     private final List<String> missingCodeModifiers = new LinkedList<>();
+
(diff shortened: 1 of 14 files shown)


jglrxavpok (approved):
Looks good to me

src/main/java/net/minestom/server/extensions/DiscoveredExtension.java
```diff
@@ -81,7 +120,27 @@ public File getOriginalJar() {
...
-    static void verifyIntegrity(@NotNull DiscoveredExtension extension) {
+    public MinestomExtensionClassLoader removeMinestomExtensionClassLoader() {
+        MinestomExtensionClassLoader oldClassLoader = getMinestomExtensionClassLoader();
+        setMinestomExtensionClassLoader(null);
+        return oldClassLoader;
+    }
+
+    public void setMinestomExtensionClassLoader(@Nullable MinestomExtensionClassLoader minestomExtensionClassLoader) {
+        this.minestomExtensionClassLoader = minestomExtensionClassLoader;
+    }
```
jglrxavpok, line 131: Why make these functions public? That's opening a way for users to break how extension classloaders work, and someone will complain some day.

src/main/java/net/minestom/server/extras/selfmodification/MinestomExtensionClassLoader.java
```diff
@@ -105,8 +105,14 @@ protected void finalize() throws Throwable {
...
+
+        if (mainClassName.equals(name))
             return true;
-        return children.stream().anyMatch(c -> c.isMainExtensionClass(name));
+
+        for (MinestomExtensionClassLoader child : children) {
+            if (child.isMainExtensionClass(name)) return true;
+        }
+
+        return false;
```
jglrxavpok, line 116: While I don't mind the change, I'm curious as to why you changed this to something that has the same semantic?

src/main/java/net/minestom/server/extras/selfmodification/MinestomRootClassLoader.java
```diff
@@ -298,11 +296,7 @@ public boolean loadModifier(URL[] urls, String codeModifierClass) {
             }
             return true;
         } catch (ClassNotFoundException | InvocationTargetException | InstantiationException | IllegalAccessException | NoSuchMethodException e) {
-            if(MinecraftServer.getExceptionManager() != null) {
-                MinecraftServer.getExceptionManager().handleException(e);
-            } else {
-                e.printStackTrace();
-            }
+            MinecraftServer.getExceptionManager().handleException(e);
```
jglrxavpok, line 299: I might have missed the change previously, but the exception manager is not guaranteed to be present at this stage. Especially if you are loading a class before `MinecraftServer#init` is called.

jglrxavpok (changes requested):

src/main/java/net/minestom/server/extras/selfmodification/MinestomExtensionClassLoader.java
LeoDog896: I'm personally not a big fan of streams, though I'm not sure where streams stand in Minestom -- like we mainly use for loops but theres streams (mainly for testing / debugging) so I wanted to follow that with this.

src/main/java/net/minestom/server/extras/selfmodification/MinestomRootClassLoader.java
LeoDog896: _Did not notice that_ Ill patch that really quickly then.

jglrxavpok (approved):

src/main/java/net/minestom/server/extras/selfmodification/MinestomExtensionClassLoader.java
TheMode: I believe that the stream in this case is cleaner

PR #182, opened 2021-03-25
Use CompletableFutures

TheMode:
Replace most of the `Runnable` and `ChunkCallback` to `CompletableFuture`


TheMode:
Merged in #303

PR #200, opened 2021-03-30
Per-chunk packet buffers & proxy scaling

TheMode:
The goal of this PR is to change how packets are serialized and sent to each individual channel.

Each chunk gets a list of buffers, each associated with a packet type (eg; one for movement, another for teleport, particle, etc...), those buffers are sent to the chunk's viewers at the end of the tick asynchronously. Bringing performance benefit by changing exponential operations to a linear one and exploiting the fact that most packets do not need to be sent in a precise order.

TODO:
- Handle chunk packets asynchronously
- Improve `Player#refreshVisibleEntities(Chunk)` performance
- Abstract it further to forward operations to proxies


TheMode:
Closed as #431 has been merged

PR #205, opened 2021-03-30
Add Tick as a TemporalUnit

kezz:
This PR adds a `TemporalUnit` implementation called `Tick`, which represents one server tick.

This means that the server tick can be used when calculating durations. For example, the following code will produce the number of milliseconds in 10 ticks, regardless of the length of each server tick.
```java
Tick.ticks(10).toMillis()
```


jglrxavpok (approved):

TheMode:
Implemented by #346

PR #206, opened 2021-03-31
New Item API

TheMode:
Make ItemStack immutable, giving us multiple advantages:
- Thread safety
- No more unwanted side effect
- Ability to cache item-related packets
- Reuse `ItemStack` objects to save on memory


src/main/java/net/minestom/server/item/ItemMeta.java
```diff
@@ -0,0 +1,132 @@
...
+import org.jglrxavpok.hephaistos.nbt.NBTCompound;
+
+import java.lang.ref.SoftReference;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class ItemMeta implements Cloneable {
```
mworzala: Does not need to be cloneable as far as I can tell.

src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -1,912 +1,177 @@
...
+        int result = stackingRule.hashCode();
+        result = 31 * result + material.hashCode();
+        result = 31 * result + amount;
+        result = 31 * result + meta.hashCode();
+        return result;
+    }
+
+    @Contract(value = "-> new", pure = true)
+    protected @NotNull ItemBuilder builder() {
+        return new ItemBuilder(material, meta.builder())
```
mworzala, line 174: Is there a reason we can't clone and store the builder used to create this item initially rather than create a new one? Similar to what is done for meta.

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+        try {
+            var builder = (ItemMetaBuilder) super.clone();
+            builder.displayName = displayName;
+            builder.lore = new ArrayList<>(lore);
+            builder.enchantmentMap = new HashMap<>(enchantmentMap);
+            deepClone(builder);
+            return builder;
+        } catch (CloneNotSupportedException e) {
+            e.printStackTrace();
+            throw new UnsupportedOperationException("Weird thing happened");
```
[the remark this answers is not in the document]
TheMode: This exception should normally never happen, it is mostly there to make the compiler happy

src/main/java/net/minestom/server/item/ItemStack.java
TheMode: I did it this way on meta because there can be an infinite number of fields but in this case only a material & amount so I didn't consider it as necessary, might be wrong

mworzala: Fair enough, I think stylistically it makes sense to do the same for each.

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Contract("_ -> this")
+    public @NotNull ItemMetaBuilder unbreakable(boolean unbreakable) {
+        this.unbreakable = unbreakable;
+        return this;
+    }
+
+    @Contract("_ -> this")
+    public @NotNull ItemMetaBuilder hideFlag(int hideFlag) {
```
mworzala, line 40: Would be nice to just pass a varargs ItemHideFlag here and have it generate the mask

TheMode: Done

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,366 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.UnaryOperator;
+
+/**
+ * Represents an inventory where items can be modified/retrieved.
+ */
+public abstract class AbstractInventory implements InventoryClickHandler, DataContainer {
```
kezz, line 24: Could this implement `Viewable`?

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,366 @@
...
+
+    /**
+     * Adds an {@link ItemStack} to the inventory and sends relevant update to the viewer(s).
+     * <p>
+     * Even the item cannot be fully added, the amount of {@code itemStack} will be updated.
+     *
+     * @param itemStack the item to add
+     * @return true if the item has been successfully fully added, false otherwise
+     */
+    public synchronized boolean addItemStack(@NotNull ItemStack itemStack) {
```
kezz, line 61: If `ItemStack`s are immutable, shouldn't this return an `ItemStack`, perhaps being nullable to indicate it fit completely?

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,366 @@
...
+
+    /**
+     * Takes an {@link ItemStack} from the inventory and sends relevant update to the viewer(s).
+     * <p>
+     * Even the item cannot be fully taken, the amount of {@code itemStack} will be updated.
+     *
+     * @param itemStack the item to take
+     * @return true if the item has been successfully fully taken, false otherwise
+     */
+    public boolean takeItemStack(@NotNull ItemStack itemStack) {
```
kezz, line 173: Same with this, it should return an `ItemStack` containing what couldn't be taken.

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    /**
+     * Checks whether {@link ItemStack}s can be fully taken from the inventory.
+     *
+     * @param itemStacks items to be checked
+     * @return true if all the items can be fully taken from the inventory, false otherwise
+     */
+    public boolean canTakeItemStacks(@NotNull List<ItemStack> itemStacks) {
+        return itemStacks.stream().allMatch(this::canTakeItemStack);
```
kezz, line 257: Same here - doesn't need to be a stream.

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,366 @@
...
+     * Gets all the {@link ItemStack} in the inventory.
+     * <p>
+     * Be aware that the returned array does not need to be the original one,
+     * meaning that modifying it directly may not work.
+     *
+     * @return an array containing all the inventory's items
+     */
+    @NotNull
+    public ItemStack[] getItemStacks() {
+        return itemStacks.clone();
```
kezz, line 298: If we store the `ItemStack`s as a list, we could just return an unmodifiable view of the stack. Saves a clone.

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,366 @@
...
+        for (int i = 0; i < itemStacks.length; i++) {
+            final ItemStack itemStack = itemStacks[i];
+            Check.notNull(itemStack, "The item array cannot contain any null element!");
+            setItemStack(i, itemStack);
+        }
+    }
+
+    @Nullable
+    @Override
+    public Data getData() {
```
kezz, line 358: Move the `@Nullable` into the method.
```java
public @Nullable Data getData() {
```

src/main/java/net/minestom/server/inventory/EquipmentHandler.java
```diff
@@ -146,8 +139,7 @@ default void setItemInHand(@NotNull Player.Hand hand, @NotNull ItemStack stack)
      * @param slot the equipment to get the item from
      * @return the equipment {@link ItemStack}
      */
-    @NotNull
-    default ItemStack getEquipment(@NotNull EntityEquipmentPacket.Slot slot) {
+    default @NotNull ItemStack getEquipment(@NotNull EntityEquipmentPacket.Slot slot) {
```
kezz, line 142: This should be moved into another enum. There's also a similar enum in `Player` as well that could use a new enum for equipment slots.

src/main/java/net/minestom/server/inventory/Inventory.java
```diff
@@ -46,41 +38,32 @@
...
+        this.offset = getSize();
+    }
 
-        ArrayUtils.fill(itemStacks, ItemStack::getAirItem);
+    /**
+     * @deprecated use {@link Inventory#Inventory(InventoryType, Component)}
+     */
+    @Deprecated
+    public Inventory(@NotNull InventoryType inventoryType, @NotNull String title) {
+        this(inventoryType, Component.text(title));
```
kezz, line 66: If this string was written directly, could it contain JSON? If so, this should be `GsonComponentSerializer.serializer().deserialize(title)`.

src/main/java/net/minestom/server/inventory/Inventory.java
```diff
@@ -115,7 +98,7 @@ public String getTitle() {
      *
      * @param title the new inventory title
      */
-    public void setTitle(@NotNull String title) {
```
kezz, line 118: If you're removing these methods you might as well remove the deprecated constructor.

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,155 @@
...
```
kezz, line 40: I think you could remove this - the enum is fine.

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Contract("_ -> this")
+    public @NotNull ItemMetaBuilder lore(Component... lore) {
+        lore(Arrays.asList(lore));
+        return this;
+    }
+
+    @Contract("_ -> this")
+    public @NotNull ItemMetaBuilder enchantments(@NotNull Map<Enchantment, Short> enchantments) {
```
kezz, line 73: `@NotNull` on `Enchantment` and the `Short`. Could we use a list of a new class, `EnchantmentLevel` that holds a primitive short?

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+    @Contract("-> new")
+    public abstract @NotNull ItemMeta build();
+
+    public abstract void read(@NotNull NBTCompound nbtCompound);
+
+    public abstract void write(@NotNull NBTCompound nbtCompound);
+
+    protected abstract void deepClone(@NotNull ItemMetaBuilder metaBuilder);
+
+    protected abstract @NotNull Supplier<@NotNull ItemMetaBuilder> getSupplier();
```
kezz, line 127: What does this method do that `clone` doesn't?

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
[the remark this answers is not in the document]
kezz: A comment for this would be nice - just to let contributors know why it isn't using the exception handler.

src/main/java/net/minestom/server/item/ItemStackBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Contract(value = "_ -> this")
+    public @NotNull ItemStackBuilder displayName(@Nullable Component displayName) {
+        this.metaBuilder.displayName(displayName);
+        return this;
+    }
+
+    @Contract(value = "_ -> this")
+    public @NotNull ItemStackBuilder lore(List<@NotNull Component> lore) {
```
kezz, line 63: Is this list `@Nullable`?

src/main/java/net/minestom/server/item/ItemTag.java
```diff
@@ -0,0 +1,97 @@
...
+                (nbtCompound, value) -> nbtCompound.setByteArray(key, value));
+    }
+
+    public static @NotNull ItemTag<String> String(@NotNull String key) {
+        return new ItemTag<>(key,
+                nbtCompound -> nbtCompound.getString(key),
+                (nbtCompound, value) -> nbtCompound.setString(key, value));
+    }
+
+    // TODO List/Compound
```
kezz, line 83: Perhaps one for `Component`?

src/main/java/net/minestom/server/item/meta/MapMeta.java
```diff
@@ -0,0 +1,278 @@
...
+    public int getMapScaleDirection() {
+        return mapScaleDirection;
+    }
+
+    /**
+     * Gets the map decorations.
+     *
+     * @return a modifiable list containing all the map decorations
+     */
+    public List<MapDecoration> getDecorations() {
```
kezz, line 60: `@Nullable`?

src/main/java/net/minestom/server/item/meta/MapMeta.java
```diff
@@ -0,0 +1,278 @@
...
+            this.mapScaleDirection = value;
+            return this;
+        }
+
+        public Builder decorations(List<MapDecoration> value) {
+            this.decorations = value;
+            return this;
+        }
+
+        public Builder mapColor(Color value) {
```
kezz, line 106: Should make this `RgbLike` instead of `Color`.

kezz (commented):
Looks great so far - just a few minor comments!

src/main/java/net/minestom/server/item/ItemStackBuilder.java
TheMode: I have been thinking about it and I believe that the list should not be null but empty instead. In order to reduce boilerplate

src/main/java/net/minestom/server/item/ItemTag.java
TheMode: There should only be methods for official NBT types https://wiki.vg/NBT#Specification
`Component` can be serialized to string as a workaround

src/main/java/net/minestom/server/item/meta/MapMeta.java
TheMode: Same reasoning as for the lore, should we perhaps consider all lists non-null but possibly empty?

src/main/java/net/minestom/server/inventory/AbstractInventory.java
TheMode: It does currently use an array as the size is final, I do not think that cloning is a huge concern

src/main/java/net/minestom/server/inventory/AbstractInventory.java
TheMode: What should the `ItemStack` represent? for example the item might merge with multiple items inside the inventory. We could replace it with a list of affected item, but I am unsure about its usefulness.

The method has been modified to "fail" when the item cannot completely fit, as it is error prone

src/main/java/net/minestom/server/item/meta/MapMeta.java
kezz: Empty lists is definitely better than nullable lists imo

src/main/java/net/minestom/server/inventory/AbstractInventory.java
kezz: The Javadoc 1:states 
> Even the item cannot be fully added, the amount of {@code itemStack} will be updated.

Is this wrong then?

src/main/java/net/minestom/server/inventory/AbstractInventory.java
kezz: Well the size could be set on a list, but this isn't a massive issue ig

src/main/java/net/minestom/server/inventory/AbstractInventory.java
TheMode: Oops, fixed

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
TheMode: Gues I would have to make the other hideFlag method Nullable, though this method is convenient when reading from vanilla nbt

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,362 @@
...
+     *
+     * @param slot the slot to check
+     * @return the item in the slot {@code slot}
+     */
+    public @NotNull ItemStack getItemStack(int slot) {
+        return itemStacks[slot];
+    }
+
+    /**
+     * Gets all the {@link ItemStack} in the inventory.
```
kezz, line 288: Might be worth noting that this specifically returns a clone of the items in the inventory. Perhaps with an `@see` or a note mentioning that getting each individual item stack would be considerably faster.

We could even have a `.forEachItem(ObjIntConsumer<ItemStack>)` method that allows people to iterate through an inventory without having to clone the array? In this case, the int would be the "slot" number.

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,362 @@
...
+    public int getSize() {
+        return size;
+    }
+
+    /**
+     * Gets the size of the "inner inventory" (which includes only "usable" slots).
+     *
+     * @return inner inventory's size
+     */
+    public int getInnerSize() {
```
kezz, line 313: Could this be renamed to `getUsableSlots` or something that represents better what this method actually returns?

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    /**
+     * Places all the items of {@code itemStacks} into the internal array.
+     *
+     * @param itemStacks the array to copy the content from
+     * @throws IllegalArgumentException if the size of the array is not equal to {@link #getSize()}
+     * @throws NullPointerException     if {@code itemStacks} contains one null element or more
+     */
+    public void copyContents(@NotNull ItemStack[] itemStacks) {
```
kezz, line 342: Is there a use case for this? I feel like `getContents` is probably good enough unless there's a specific use case for this method I'm missing.

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+            this.nbt.removeTag("AttributeModifiers");
+        }
+
+        return this;
+    }
+
+    @Contract("_ -> this")
+    public @NotNull ItemMetaBuilder customModelData(int customModelData) {
+        this.customModelData = customModelData;
+        this.nbt.setInt("CustomModelData", customModelData);
```
kezz, line 151: Is modifying the NBT a relatively costly operation? Might be better to just create the NBT once the builder is built if so.

src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -1,903 +1,207 @@
...
+    public boolean isSimilar(@NotNull ItemStack itemStack) {
+        return material.equals(itemStack.material) &&
+                meta.equals(itemStack.meta);
     }
 
     @Override
-    public @NotNull HoverEvent<ShowItem> asHoverEvent(@NotNull UnaryOperator<ShowItem> op) {
-        return HoverEvent.showItem(op.apply(ShowItem.of(this.material, this.amount, NBTUtils.asBinaryTagHolder(this.toNBT().getCompound("tag")))));
-    }
+    public boolean equals(Object o) {
```
kezz, line 173: Is this method missing a closing bracket or is git being weird??

src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -1,903 +1,207 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
 
-    // Callback events
+        ItemStack itemStack = (ItemStack) o;
+        if (uuid.equals(itemStack.uuid)) return true;
```
kezz, line 178: I think we should do more than compare the UUID. Would two items created that are the same not have different UUIDs?

src/main/java/net/minestom/server/item/meta/CrossbowMeta.java
```diff
@@ -0,0 +1,190 @@
...
+        private boolean triple;
+        private ItemStack projectile1, projectile2, projectile3;
+        private boolean charged;
+
+        /**
+         * Sets the projectile of this crossbow.
+         *
+         * @param projectile the projectile of the crossbow
+         */
+        public Builder projectile(@NotNull ItemStack projectile) {
```
kezz, line 94: I wonder if this (and the other method) should be nullable as there's no way to "remove" the projectile at the moment.

src/main/java/net/minestom/server/item/meta/EnchantedBookMeta.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @return an unmodifiable map containing the item stored enchantments
+     */
+    public @NotNull Map<Enchantment, Short> getStoredEnchantmentMap() {
+        return Collections.unmodifiableMap(storedEnchantmentMap);
+    }
+
+    public static class Builder extends ItemMetaBuilder {
+
+        private Map<Enchantment, Short> enchantments = new HashMap<>();
```
kezz, line 36: Should we have a getter for this map to retrieve currently stored enchantments?

src/main/java/net/minestom/server/item/meta/FireworkMeta.java
```diff
@@ -0,0 +1,92 @@
...
+                NBTList<NBTCompound> explosions = new NBTList<>(NBTTypes.TAG_Compound);
+                for (FireworkEffect effect : this.effects) {
+                    explosions.add(effect.asCompound());
+                }
+                nbtCompound.set("Explosions", explosions);
+            });
+            return this;
+        }
+
+        public Builder flightDuration(byte flightDuration) {
```
kezz, line 54: Should we be holding this as a `Duration` and converting it to a byte when it's sent?

TheMode: We could probably yes

src/main/java/net/minestom/server/item/meta/CrossbowMeta.java
TheMode: Right, but I find it a bit weird for it to be the only place where an item can be null, what about using AIR instead?

src/main/java/net/minestom/server/item/ItemStack.java
TheMode: Honestly we could just remove the identifier, as we do not plan to introduce server-side data it is currently unused

src/main/java/net/minestom/server/inventory/AbstractInventory.java
TheMode: In my case, retrieving all player items from a database

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -0,0 +1,362 @@
...
+
+    /**
+     * Adds an {@link ItemStack} to the inventory and sends relevant update to the viewer(s).
+     * <p>
+     * Does nothing if the item cannot be fully added.
+     *
+     * @param itemStack the item to add
+     * @return true if the item has been successfully added, false otherwise
+     */
+    public synchronized boolean addItemStack(@NotNull ItemStack itemStack) {
```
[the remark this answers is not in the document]
TheMode: Yep no complain, we will have to think about the naming carefully

src/main/java/net/minestom/server/inventory/AbstractInventory.java
TheMode: I feel that this should be replaced by a kind of Inventory Section API

src/main/java/net/minestom/server/item/meta/EnchantedBookMeta.java
TheMode: You mean in the builder? I dont think that we should have any getter there

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
TheMode: What is the goal? I believe that `Map#of` should be enough

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -146,6 +149,30 @@
         return this;
     }
 
+    @Contract("_ -> this")
+    public @NotNull ItemMetaBuilder canPlaceOn(@NotNull Set<@NotNull Block> blocks) {
```
kezz, line 153: I think it might be nice to have varargs methods for these setters that just overload the set methods.

TheMode: Done :)

src/main/java/net/minestom/server/item/ItemStoreBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ItemStoreBuilder {
```
mworzala, line 11: same comment

src/main/java/net/minestom/server/item/meta/CompassMeta.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        @Override
+        public @NotNull CompassMeta build() {
+            return new CompassMeta(this, lodestoneTracked, lodestoneDimension, lodestonePosition);
+        }
+
+        @Override
+        public void read(@NotNull NBTCompound nbtCompound) {
+            if (nbtCompound.containsKey("LodestoneTracked")) {
```
mworzala, line 87: Only sort of relevant to this PR, would be nice to have some `getOrDefault` type methods in the nbt lib

src/main/java/net/minestom/server/item/ItemMeta.java
```diff
@@ -0,0 +1,164 @@
...
+    public @NotNull Map<Enchantment, Short> getEnchantmentMap() {
+        return Collections.unmodifiableMap(enchantmentMap);
+    }
+
+    @Contract(pure = true)
+    public @NotNull List<@NotNull ItemAttribute> getAttributes() {
+        return Collections.unmodifiableList(attributes);
+    }
+
+    @Contract(pure = true)
```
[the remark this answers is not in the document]
TheMode: I was thinking of cleaning up the API progressively, perhaps not necessary

src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    @Contract(value = "_, _ -> new", pure = true)
+    public static @NotNull ItemMetaBuilder fromNBT(@NotNull ItemMetaBuilder src, @NotNull NBTCompound nbtCompound) {
+        ItemMetaBuilder dest = src.getSupplier().get();
+        NBTUtils.loadDataIntoMeta(dest, nbtCompound);
+        return dest;
+    }
+
+    public interface Provider<T> {
```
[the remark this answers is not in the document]
TheMode: Currently used inside every meta to specify the builder https://github.com/Minestom/Minestom/blob/item-api/src/main/java/net/minestom/server/item/meta/CompassMeta.java#L12

PR #208, opened 2021-03-31
Testing framework

### LINKED ISSUE #128: Testing framework
Minestom lacks tests to ensure its feature work properly.

Having a testing framework for Minestom could allow to avoid regression bugs, make sure features are implemented correctly, and finally, can help Minestom developers (extensions, libraries or server implementations) to test their own code easily.

Testing features inside Minestom can quickly become overwhelming: setting up the server, registering instances, setting up the test and finally writing the test code can take a long time to get right, and needs to be repeated for each test.

### Additional Information
While it is possible to greatly reduce the length of code required via JUnit's `@Before` annotation, using a Minestom-specific framework could go a long way to improve the user experience.

This is heavily inspired by Mojang's own testing framework: https://www.youtube.com/watch?v=vXaWOJTCYNg

jglrxavpok:
Will fix #128


KrystilizeNevaDies:
https://github.com/Minestom/Minestom/issues/128#issuecomment-1032219399

PR #225, opened 2021-04-05
Tablist API

Joeyoey:
This will allow (others) to make and change a player's TabList that they are viewing, this also allows (others) to make and manage multiple separate TabLists and enable and disable them for individual players.

All can be accessed through the TabListManager, where you can get a player's active TabList , set TabLists, set spoofed gamemodes, set spoofed ping, and even allows the addition of Fake players to the TabList.


src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,304 @@
...
+        return new PlayerInfoPacket.AddPlayer.Property("textures", textures, signature);
+    }
+
+    /**
+     * Whether the TabList is configured to automatically send player latency
+     * updates to all its viewers
+     *
+     * @return boolean whether player latency will be auto-updated on TabLists
+     */
+    public boolean isLatencyUpdates() {
```
kezz, line 272: I wonder if this could be named better - maybe `hasLatencyUpdates()`?

src/main/java/net/minestom/server/tab/TabListManager.java
```diff
@@ -0,0 +1,96 @@
...
+
+public class TabListManager {
+
+    private final Set<TabList> tabLists = new CopyOnWriteArraySet<>();
+    private TabListPopulator tabListPopulator; // needs to be init to default at minimum
+
+    public TabListManager() {
+        this.setTabListPopulator(new DefaultTabPopulator(this));
+        MinecraftServer.getGlobalEventHandler().addEventCallback(PlayerSpawnEvent.class, event -> this.tabListPopulator.onJoin(event.getPlayer()));
+        MinecraftServer.getGlobalEventHandler().addEventCallback(PlayerDisconnectEvent.class, event -> this.tabListPopulator.onLeave(event.getPlayer()));
```
kezz, line 23: I think this might be better to do somewhere else - perhaps in player init? @TheMode

kezz (changes requested):
Looks good so far! Just requested a few changes but I love the idea of a tab list API. 

I wonder if it could be possible to add fake players to the tab list without the overhead making a `FakePlayer`? Would be interested to hear your/others thoughts on it but perhaps if we make a `TabListMember` interface that provides the methods needed to populate the packet. Then this could be implemented by `Player` but we could also provide a `TabListMember.of(String username, etc)` method so that people can make fake tab list members without having to make a `FakePlayer`?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -625,11 +630,18 @@ public boolean removeViewer0(@NotNull Player player) {
...
+            if (tabList.getDisplayedPlayers().contains(this)) {
+                tabList.removeDisplayedPlayer(this);
+            }
+        }
+
 
         // Team
         if (this.getTeam() != null && this.getTeam().getMembers().size() == 1) {// If team only contains "this" player
-            viewerConnection.sendPacket(this.getTeam().createTeamDestructionPacket());
+         //   viewerConnection.sendPacket(this.getTeam().createTeamDestructionPacket()); // todo does this need to be changed before merge? What's up here
```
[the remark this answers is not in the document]
ZakShearman: @Joeyoey I added this todo as I don't know what this was added here for. When you've got time if you can take a look.

src/main/java/net/minestom/server/tab/TabListManager.java
ZakShearman: I think this would be a good thing to be done inside play init.

Would this therefore be done similarly to setting a player's spawning instance where if it's not set it'll throw an exception?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2501,29 +2521,43 @@ public long getLastKeepAlive() {
...
 
-        playerInfoPacket.playerInfos.add(addPlayer);
+    /**
+     * Gets the packet to remove all the players from tablist.
+     *
+     * @return a {@link PlayerInfoPacket} to remove the players
+     */
+    protected @NotNull PlayerInfoPacket getRemoveAllPlayerToList() {
+        PlayerInfoPacket playerInfoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.REMOVE_PLAYER);
+        for (Player player : MinecraftServer.getConnectionManager().getOnlinePlayers()) {
```
TheMode, line 2557: Not a fan of all player loop

src/main/java/net/minestom/server/tab/TabListManager.java
TheMode: I also believe that the main repo should not rely on events, those are for the library users.

There is a default `TabList`, so you could apply it on login

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1407,6 +1412,26 @@ private void sendChangeGameStatePacket(@NotNull ChangeGameStatePacket.Reason rea
         playerConnection.sendPacket(changeGameStatePacket);
     }
 
+    /**
+     * Gets the TabList the player is viewing
+     * Will be null if the {@link net.minestom.server.tab.TabListPopulator} has not assigned a TabList
+     *
+     * @return the viewed {@link TabList}
+     */
+    public @Nullable TabList getTabList() {
```
TheMode, line 1421: Should be `NotNull`, as the default `TabList` should be used (showing all online players)

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2501,29 +2521,43 @@ public long getLastKeepAlive() {
...
+     * Gets the packet to render all the players that are online.
      *
-     * @return a {@link PlayerInfoPacket} to add the player
+     * @return a {@link PlayerInfoPacket} containing all online players
      */
-    @NotNull
-    protected PlayerInfoPacket getAddPlayerToList() {
+    protected @NotNull PlayerInfoPacket getAddAllPlayerToList() {
         PlayerInfoPacket playerInfoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.ADD_PLAYER);
+        for (Player player : MinecraftServer.getConnectionManager().getOnlinePlayers()) {
```
TheMode, line 2530: Same as before, all players loop are terrible

TheMode (changes requested):

src/main/java/net/minestom/server/entity/Player.java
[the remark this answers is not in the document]
kezz: If you're removing a viewer, you're often removing players from the server or moving them into a different instance, etc. Both of these remove players from teams. So this bit of code is for checking if a team is now empty as a result of a viewer being removed.

At least, I think this is the case. Someone else might have better insight.

src/main/java/net/minestom/server/entity/Player.java
ZakShearman: In order for player skins to display, we currently add all players to a user's tab when they join then remove them if necessary.

I don't know of a better way to do this unless there's some way to tell when a user comes into another user's render distance server side.

src/main/java/net/minestom/server/tab/TabList.java
Joeyoey: It isn't has since it's not a queue/updating system it's a toggle as to whether the tablist will automatically update latency, or if latency is spoofed.

src/main/java/net/minestom/server/tab/TabListManager.java
Joeyoey: Moved to player init and player#remove

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1302,9 +1314,8 @@ public void setDisplayName(@Nullable JsonMessage displayName) {
     public void setDisplayName(@Nullable Component displayName) {
         this.displayName = displayName;
 
-        PlayerInfoPacket infoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.UPDATE_DISPLAY_NAME);
-        infoPacket.playerInfos.add(new PlayerInfoPacket.UpdateDisplayName(getUuid(), displayName));
-        sendPacketToViewersAndSelf(infoPacket);
+        assert this.getTabList() != null;
```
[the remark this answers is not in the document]
Joeyoey: removed.

src/main/java/net/minestom/server/entity/Player.java
Joeyoey: It's the only way for player skins to be cached.

src/main/java/net/minestom/server/entity/Player.java
TheMode: `Player#addViewer`?

ZakShearman: So in the addViewer method we then do the add/remove if they're not on the tablist? Then additionally this means we'll end up doing this multiple times unless we keep a cache of every player skin each player has received.

src/main/java/net/minestom/server/tab/TabList.java
kezz: I still think `isLatencyUpdates` just doesn't make any sense. Perhaps `doesLatencyUpdates` or `handlesLatencyUpdates`?

src/main/java/net/minestom/server/entity/Player.java
TheMode: Why cannot you trust the tablist populator? Feels like it should manage which players you need to send instead of the whole server.
I am probably missing something

ZakShearman: Yh. It's designed so you don't have to add everyone that a player can see to the tablist. E.g you could only display your faction members in the tablist.

We still have to send all player skins as otherwise, say they go to spawn and see other users, they would all have alex/steve skins.

TheMode: I don't think that sending a new PlayerInfo packet in `Player#addViewer` is an issue, caching should be unnecessary.

So as you said, just sending the players you need to see in the tablist on login, and then ensure that player can be seen in the addViewer method

ZakShearman: Sounds good 👍🏻. @joeyoey if you've got time for this otherwise I'll do it in a few days.

src/main/java/net/minestom/server/tab/TabList.java
Joeyoey: I agree will change to doesLatencyUpdates

src/main/java/net/minestom/server/entity/Player.java
Joeyoey: Will do, basically there can be multiple tablists and each tab list can have as many or as few players as are chosen. However, typically without zak and I’s api a player that is not on the tab list can not be seen as the client sees the other player as invisible or Steve/Alex since they’re not on their tab list, so to avoid this we HAVE to cache to keep the skins and other players visible. However the populator is over complicated for what it is, so I’ll move the add/remove of players from tablists to the addviewer, removeviewer, player#init, and player#disconnect/player#remove. Not sure about the last one as in preliminary tests the removal of the player on disconnect breaks some things. I’ll debug and figure it out in the next few days.

src/main/java/net/minestom/server/entity/Player.java
Joeyoey: This only gets called in the init, to add and then remove all players to tablist, I'll consolidate them into one for loop.

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2501,32 +2530,40 @@ public long getLastKeepAlive() {
...
 
     /**
-     * Gets the packet to add the player from the tab-list.
+     * Sends the packet to render all the players that are online.
      *
-     * @return a {@link PlayerInfoPacket} to add the player
      */
-    @NotNull
-    protected PlayerInfoPacket getAddPlayerToList() {
+    protected void getAllPlayerToList() {
```
TheMode, line 2536: `get` feels a bit misleading as it doesn't return anything

TheMode (changes requested):

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -562,6 +571,15 @@ public boolean isOnGround() {
     public void remove() {
         callEvent(PlayerDisconnectEvent.class, new PlayerDisconnectEvent(this));
 
+        // todo issue not caused by any of this
+        // tab list
```
jglrxavpok, line 575: Could you clarify what is meant by this comment?

src/main/java/net/minestom/server/entity/Player.java
jglrxavpok: Rename to something like `sendAllPlayerInList()`

src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,304 @@
...
+import net.minestom.server.utils.PacketUtils;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+public class TabList implements Viewable {
```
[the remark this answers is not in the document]
jglrxavpok: I second this.

src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * Whether the TabList is configured to automatically send player gamemode
+     * updates to all its viewers
+     *
+     * @return boolean whether player gamemode will be auto-updated on TabLists
+     */
+    public boolean isGamemodeUpdates() {
+        return this.gamemodeUpdates;
+    }
```
jglrxavpok, line 286: Use the same wording as `doesLatencyUpdates()`

src/main/java/net/minestom/server/tab/TabListManager.java
```diff
@@ -0,0 +1,117 @@
...
+import net.minestom.server.adventure.audience.PacketGroupingAudience;
+import net.minestom.server.entity.Player;
+import net.minestom.server.network.packet.server.play.PlayerInfoPacket;
+import net.minestom.server.utils.PacketUtils;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
+public class TabListManager {
```
jglrxavpok, line 12: You should probably add a javadoc comment to describe the manager a bit.

jglrxavpok (changes requested):
I can see a few changes to make, but other than that, it looks good to me.

I'll add a comment made by @kezz on Discord:
> I am iffy on how the tab lists are being sent - I think it might be worth asking for some code being added to the demo section to demonstrate how it works

src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,296 @@
...
+
+            // Skin support
+            if (displayedPlayer.getSkin() != null) {
+                addPlayer.properties.add(this.createPropertyPacket(player));
+            }
+
+            playerInfoPacket.playerInfos.add(addPlayer);
+        }
+        player.getPlayerConnection().sendPacket(playerInfoPacket);
+        PacketUtils.sendGroupedPacket(this.viewers, this.generateHeaderAndFooterPacket());
```
TheMode, line 189: See `Viewable#sendPacketToViewers`

src/main/java/net/minestom/server/entity/Player.java
TheMode: So, is the all-player loop still necessary? I expect it to be a huge bottleneck with high amount of players

src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,296 @@
...
+    public void setDisplayedPing(@NotNull Player player, int latency) {
+        if (this.latencyUpdates)
+            throw new IllegalStateException("Cannot set fake latency unless latencyUpdates is set to false");
+        if (!this.displayedPlayers.contains(player))
+            throw new IllegalStateException("Cannot set fake latency for a player not displayed on the TabList");
+
+        PlayerInfoPacket playerInfoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.UPDATE_LATENCY);
+        playerInfoPacket.playerInfos.add(new PlayerInfoPacket.UpdateLatency(player.getUuid(), latency));
+
+        PacketUtils.sendGroupedPacket(this.viewers, playerInfoPacket);
```
TheMode, line 248: Again, `Viewable#sendPacketToViewers`. there are likely other places in this code where it should be used

TheMode (changes requested):

src/main/java/net/minestom/server/entity/Player.java
Joeyoey: Does Player#addViewer get called when the player logs in? If so I can remove it entirely.

src/main/java/net/minestom/server/entity/Player.java
Joeyoey: We had an issue due to players logging out and having all the players removed from the TabList, it has since been fixed. Comment will be removed :D

src/main/java/net/minestom/server/tab/TabList.java
Joeyoey: Added this, didn't realize that was a thing.

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -250,7 +252,13 @@ public void UNSAFE_init(@NotNull Instance spawnInstance) {
...
         this.skin = skinInitEvent.getSkin();
         // FIXME: when using Geyser, this line remove the skin of the client
-        playerConnection.sendPacket(getAddPlayerToList());
+
+        // tab list
+        this.tabList = MinecraftServer.getTabListManager().getDefaultTabList();
+        this.tabList.addViewer(this);
+        this.tabList.addDisplayedPlayer(this);
+
+
```
kezz, line 261: Unnecessary whitespace

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -562,6 +570,13 @@ public boolean isOnGround() {
     public void remove() {
         callEvent(PlayerDisconnectEvent.class, new PlayerDisconnectEvent(this));
 
+        this.tabList.removeViewer(this);
+        for (TabList tabList : MinecraftServer.getTabListManager().getTabLists()) {
```
kezz, line 574: There's got to be a better way to do this... Could we keep a map of Player -> TabList in the manager?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1302,9 +1315,7 @@ public void setDisplayName(@Nullable JsonMessage displayName) {
     public void setDisplayName(@Nullable Component displayName) {
         this.displayName = displayName;
 
-        PlayerInfoPacket infoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.UPDATE_DISPLAY_NAME);
-        infoPacket.playerInfos.add(new PlayerInfoPacket.UpdateDisplayName(getUuid(), displayName));
-        sendPacketToViewersAndSelf(infoPacket);
+        this.getTabList().updateDisplayName(this);
```
kezz, line 1318: Do these methods (display name, latency, gamemode, etc) not need to update the display name for any tab list the player is in?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1335,27 +1345,22 @@ public synchronized void setSkin(@Nullable PlayerSkin skin) {
...
-        playerConnection.sendPacket(removePlayerPacket);
+        for (TabList tabList : MinecraftServer.getTabListManager().getTabLists()) {
+            if (tabList.getDisplayedPlayers().contains(this)) {
+                tabList.removeDisplayedPlayer(this);
+                tabList.addDisplayedPlayer(this);
+            } else {
+                tabList.addDisplayedPlayer(this);
+                tabList.removeDisplayedPlayer(this);
+            }
+        }
```
kezz, line 1361: Is it possible that with how the networking side of things is set up that these packets could be sent out of order? @TheMode 

Also, if the player isn't in a given tab list, why do you bother adding and removing them? Shouldn't we still be sending the packet to all viewers here in case they aren't in a tab list together but can still view a given player?

src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,297 @@
...
+        return new PlayerInfoPacket.AddPlayer.Property("textures", textures, signature);
+    }
+
+    /**
+     * Whether the TabList is configured to automatically send player latency
+     * updates to all its viewers
+     *
+     * @return boolean whether player latency will be auto-updated on TabLists
+     */
+    public boolean doesLatencyUpdate() {
```
kezz, line 265: `doesLatencyUpdates`

src/main/java/net/minestom/server/tab/TabListManager.java
```diff
@@ -0,0 +1,116 @@
...
+        return tablist;
+    }
+
+    /**
+     * Gets the registered TabLists
+     *
+     * @return the registered {@link TabList}s
+     */
+    public @NotNull Set<TabList> getTabLists() {
+        return this.tabLists;
```
kezz, line 32: Should this return an unmodifiable view of the tab list?

kezz (changes requested):

src/main/java/net/minestom/server/entity/Player.java
mworzala: +1

src/main/java/net/minestom/server/tab/TabList.java
```diff
@@ -0,0 +1,297 @@
...
+     * This will not update the tab list immediately so you must call {TabList#update}
+     *
+     * @param footer the new footer content
+     */
+    public void setFooter(@Nullable Component footer) {
+        this.footer = footer;
+        this.sendPacketToViewers(this.generateHeaderAndFooterPacket());
+    }
+
+    private PlayerListHeaderAndFooterPacket generateHeaderAndFooterPacket() {
```
mworzala, line 93: Stylistically I would prefer the private methods to be kept at the bottom

src/main/java/net/minestom/server/tab/TabListManager.java
mworzala: +1

src/main/java/net/minestom/server/tab/TabListManager.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * This TabList is assigned to a player when they log in.
+     *
+     * @return the default TabList
+     */
+    public @NotNull TabList getDefaultTabList() {
```
mworzala, line 40: As far as I can tell there isn't a way to provide a default tab list per player. May be worth either exposing it in `PlayerSpawnEvent` (or an earlier one), or changing this to be a `TabListProvider` or something where the default ignores the player.

mworzala (changes requested):

KrystilizeNevaDies:
Closing as this is very much outdated.

PR #228, opened 2021-04-07
Add ArgumentMap

LeoDog896:
This adds ArgumentMap, along with the functional interface ArgumentMapper, to simplify processing of complex objects in commands.


src/main/java/net/minestom/server/command/builder/ArgumentMapper.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Represents a lambda that can turn an input into an output
+ * that also allows the throwing of ArgumentSyntaxException
+ *
+ * @param <I> The input expected from the Argument
+ * @param <O> The desired output type from this lambda.
+ */
+@FunctionalInterface
+public interface ArgumentMapper<I, O> {
```
TheMode, line 13: Why not just use a `Function`?

LeoDog896: Because it can throw an error -- originally I did use a function but exceptions cant be thrown in it without being caught inside the lambda.

src/main/java/net/minestom/server/command/builder/arguments/Argument.java
```diff
@@ -220,6 +217,18 @@ public boolean hasSuggestion() {
...
 
+    /**
+     * Maps this argument's output to another result.
+     *
+     * @param mapper The mapper to use (this argument's input --> desired output)
+     * @param <O> The type of output expected.
+     *
+     * @return A new ArgumentMap that can get this complex object type.
+     */
+    public <O> ArgumentMap<T, O> map(ArgumentMapper<T, O> mapper) {
```
TheMode, line 228: Could you annotate it with `@Beta`?

I am ready to merge it, but we might find an alternate path in the future

PR #229, opened 2021-04-07
Better catching for events/commands

LeoDog896:
This adds try/catch statements to commands/events for Minestom.

### DIFF
--- a/src/main/java/net/minestom/server/event/handler/EventHandler.java
+++ b/src/main/java/net/minestom/server/event/handler/EventHandler.java
@@ -109,22 +109,27 @@ public interface EventHandler extends IExtensionObserver {
      */
     default <E extends Event> void callEvent(@NotNull Class<E> eventClass, @NotNull E event) {
 
-        // Global listeners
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/event/handler/EventHandler.java
```diff
@@ -109,22 +109,27 @@
...
-                runEvent(instance.getEventCallbacks(eventClass), event);
+            // Call the same event for the current entity instance
+            if (this instanceof Entity) {
+                final Instance instance = ((Entity) this).getInstance();
+                if (instance != null) {
+                    runEvent(instance.getEventCallbacks(eventClass), event);
+                }
             }
+        } catch (Exception exception) {
+            MinecraftServer.getExceptionManager().handleException(exception);
```
jglrxavpok, line 132: Chould this be done per-event callback instead? If one callback throws, the remaining callbacks won't be called right?

LeoDog896: Yeah, most likely

Issue #153, opened 2021-02-25
Game data


TheMode:
The goal is to remove our dependency on prismarine minecraft-data and Mojang data generators.

It might also allow us to become registry-based instead of generating enums. Allowing users to create their own blocks, entities, etc... with a modded client.

Here is our wish list of fields to retrieve from the vanilla client.

## Blocks
- id
- display name
- hardness
- bounding box information (even for complex shapes such as stairs)
- id for every state
- Block Entity namespace id
- luminance (emitted light)
- opacity (consumed light)

## Entity types
- id
- display name
- bounding box

## Items
- id
- Max default stack size
- Corresponding block

## Map Colors
- name
- id
- colorValue

jglrxavpok:
# Additional info
## Blocks
* BlockEntity namespace id

## Items
* id
* Max default stack size
* Corresponding block

Articdive:
Now closed by #309

Issue #136, opened 2021-02-07
Possibly simplify initializing block alternatives, which could help with Block States API later on?


MichaelPriebe:
Here's an example of code that can generate all the magic ids of block states by only providing the default: https://gist.github.com/MichaelPriebe/c4c9c24e37ef25073f3dbf43946a5cd3

Its output matches all the traditionally hardcoded states: https://github.com/Minestom/Minestom/blob/master/src/autogenerated/java/net/minestom/server/instance/block/states/OakStairs.java

I am not sure how it would be implemented though, but it could lead to a more enjoyable usage of block states, instead of funky "key=value" lists.

TheMode:
One issue with this approach could be memory consumption. It does create a bunch of unnecessary objects

PR #236, opened 2021-04-09
Use new data generator

Articdive:
# Checklist
- [x] Blocks
- [x] Entities
- [x] Villager types & professions
- [x] BlockEntities
- [x] Particles
- [x] Sounds
- [x] Items
- [x] Enchantments
- [x] Map Colors
- [x] Attributes
- [x] Fluids
- [x] Change ResourceGatherer to get the new data. (Included in the JAR)
- [x] Remove PrismarineJS, Burger, other tools.
- [x] Implement VanillaData / RawData query system. NOTE: Implemented without community approval.
- [x] Potion Effects
- [x] Potion Types
- [x] Statistics
- [ ] Give the new code javadocs.
- [x] Start work on the data repository and use gradle to include that in resources instead of statically including it in the main repo.
# Postponed due to extremely controversial and difficult to implement ideas and changes:
- Decide how to handle custom blocks.
--> Current behaviour will be kept for custom blocks.
- Run a code linter on the generated code.
--> There are genuinely no good code linters supporting IntelliJ configurations or Java 11. Let alone both. Decided to improve the code style a little bit by fixing the indentation and some other stylistic stuff.
# Optional:
- [x] Use subprojects for the generators instead of a source set.
- [x] Improve properties for BlockState (maybe interesting for: #136).
- [x] Stop using Mojang data extraction (directly)


Articdive:
This would close #153.

src/main/java/net/minestom/server/entity/pathfinding/PFBlockDescription.java
```diff
@@ -76,74 +83,42 @@ public boolean isFullyBounded() {
...
         // Beds
-        if (block.name().toUpperCase().endsWith("BED")) {
+        if (block.getName().toUpperCase().endsWith("BED")) {
             return false;
         }
         // Glass Panes
-        if (block.name().toUpperCase().endsWith("PANE")) {
+        if (block.getName().toUpperCase().endsWith("PANE")) {
             return false;
         }
```
kezz, line 111: Not really related to this PR but this should absolutely be replaced with something that doesn't rely on names at some point.

src/main/java/net/minestom/server/entity/pathfinding/PFBlockDescription.java
```diff
@@ -76,74 +83,42 @@ public boolean isFullyBounded() {
...
-            }
-            default: {
-                return true;
-            }
+        if (List.of(
+                CHORUS_FLOWER, CHORUS_PLANT, BAMBOO, BAMBOO_SAPLING, SEA_PICKLE, TURTLE_EGG, SNOW, FLOWER_POT,
+                LILY_PAD, ANVIL, CHIPPED_ANVIL, DAMAGED_ANVIL, CAKE, CACTUS, BREWING_STAND, LECTERN, DAYLIGHT_DETECTOR,
+                CAMPFIRE, SOUL_CAMPFIRE, ENCHANTING_TABLE, CHEST, ENDER_CHEST, GRINDSTONE, TRAPPED_CHEST, SOUL_SAND,
+                SOUL_SOIL, LANTERN, COCOA, CONDUIT, GRASS_PATH, FARMLAND, END_ROD, STONECUTTER, BELL
+        ).contains(block)) {
```
kezz, line 118: This list should be stored as a constant somewhere.

src/main/java/net/minestom/server/entity/pathfinding/PFBlockObject.java
```diff
@@ -85,79 +92,46 @@ public boolean isFullyBounded() {
...
-            case BELL: {
-                return false;
-            }
-            default: {
-                return true;
-            }
+        if (List.of(CHORUS_FLOWER, CHORUS_PLANT, BAMBOO, BAMBOO_SAPLING, SEA_PICKLE, TURTLE_EGG, SNOW, FLOWER_POT,
+                LILY_PAD, ANVIL, CHIPPED_ANVIL, DAMAGED_ANVIL, CAKE, CACTUS, BREWING_STAND, LECTERN, DAYLIGHT_DETECTOR,
+                CAMPFIRE, SOUL_CAMPFIRE, ENCHANTING_TABLE, CHEST, ENDER_CHEST, GRINDSTONE, TRAPPED_CHEST, SOUL_SAND,
+                SOUL_SOIL, LANTERN, COCOA, CONDUIT, GRASS_PATH, FARMLAND, END_ROD, STONECUTTER, BELL
```
kezz, line 125: Same here, should be cached.

src/autogenerated/java/net/minestom/server/map/MapColors.java
```diff
@@ -0,0 +1,201 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * AUTOGENERATED by MapColorsGenerator
+ */
+public final class MapColors {
```
kezz, line 11: Not sure I like the name clash with this and `MapColor`. Can't quite think of a better option though..

src/main/java/net/minestom/server/registry/Registry.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minestom.server.potion.PotionEffect;
+import net.minestom.server.potion.PotionType;
+import net.minestom.server.sound.SoundEvent;
+import net.minestom.server.statistic.StatisticType;
+
+/**
+ * AUTOGENERATED
+ */
+public final class Registry {
+    public static final IdCrossMapRegistry.Defaulted<Block> BLOCK_REGISTRY = new IdCrossMapRegistry.Defaulted<>(() -> Block.AIR);
```
kezz, line 23: These don't need to end with registry. I think `Registry.BLOCKS` would be enough for example.

src/main/java/net/minestom/server/registry/IRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @NotNull
+    static <T extends Keyed> T get(@NotNull IRegistry<T> registry, @NotNull NamespaceID id) {
+        return registry.get(id);
+    }
+
+    @NotNull
+    static <T extends Keyed> T get(@NotNull IRegistry<T> registry, @NotNull Key key) {
+        return registry.get(key);
+    }
```
kezz, line 28: Would be nice for this to have a `Keyed` overload.

src/main/java/net/minestom/server/raw_data/RawMaterialData.java
```diff
@@ -0,0 +1,24 @@
...
+    public SoundEvent drinkingSound;
+    @Nullable
+    public RawArmorData armorData;
+
+    public static class RawArmorData {
+        public int defense;
+        public double toughness;
+        public EntityEquipmentPacket.Slot slot; // TODO: Maybe better class
+    }
+}
```
kezz, line 24: More line end issues here.

src/main/java/net/minestom/server/raw_data/RawMaterialData.java
```diff
@@ -0,0 +1,24 @@
+package net.minestom.server.raw_data;
```
kezz, line 1: Also not a fan of the underscore in this package name either.

src/main/java/net/minestom/server/raw_data/RawMaterialData.java
```diff
@@ -0,0 +1,24 @@
...
+
+import net.minestom.server.instance.block.Block;
+import net.minestom.server.network.packet.server.play.EntityEquipmentPacket;
+import net.minestom.server.sound.SoundEvent;
+import org.jetbrains.annotations.Nullable;
+
+public class RawMaterialData {
+    public boolean damageable;
```
kezz, line 9: As these are non-final, would it allow someone to edit these values and would that cause any issues if they did?

src/main/java/net/minestom/server/raw_data/DataInitializer.java
```diff
@@ -0,0 +1,319 @@
...
+                blockStateData.lightEmission = blockStateJson.get("lightEmission").getAsInt();
+                blockStateData.occluding = blockStateJson.get("doesOcclude").getAsBoolean();
+                blockStateData.pushReaction = blockStateJson.get("pushReaction").getAsString();
+                blockStateData.blocksMotion = blockStateJson.get("blocksMotion").getAsBoolean();
+                blockStateData.flammable = blockStateJson.get("isFlammable").getAsBoolean();
+                blockStateData.liquid = blockStateJson.get("isLiquid").getAsBoolean();
+                blockStateData.replaceable = blockStateJson.get("isReplaceable").getAsBoolean();
+                blockStateData.solid = blockStateJson.get("isSolid").getAsBoolean();
+                blockStateData.solidBlocking = blockStateJson.get("isSolidBlocking").getAsBoolean();
+                blockStateData.mapColor = MapColors.values()[blockStateJson.get("mapColorId").getAsInt()];
```
kezz, line 101: I personally think we should deprecate the `.values()` method and mark it as for removal. This is creating and allocating a whole new array each time which is way too much work.

src/main/java/net/minestom/server/network/packet/server/play/TagsPacket.java
```diff
@@ -40,18 +40,18 @@ public TagsPacket() {}
...
     @Override
     public void read(@NotNull BinaryReader reader) {
-        readTags(reader, blockTags, id -> NamespaceID.from("minecraft", Block.values()[id].getName()));
-        readTags(reader, itemTags, id -> NamespaceID.from("minecraft", Material.values()[id].getName()));
-        readTags(reader, fluidTags, id -> NamespaceID.from(Fluid.values()[id].getNamespaceID()));
-        readTags(reader, entityTags, id -> NamespaceID.from(EntityType.values()[id].getNamespaceID()));
+        readTags(reader, blockTags, id -> Block.values().get(id).getId());
+        readTags(reader, itemTags, id -> Material.values().get(id).getId());
+        readTags(reader, fluidTags, id -> Fluid.values().get(id).getId());
+        readTags(reader, entityTags, id -> EntityType.values().get(id).getId());
```
kezz, line 54: As before, we should find a better solution than creating a new list each time this function is called.

kezz (changes requested):
Looks absolutely amazing so far! Left a whole bunch of comments, mostly minor nitpicks.

Some general thoughts:

- I'm not a fan of capitalised `Generators` module, it should be all lowercase.
- I think we should probably put the auto-generated data in its own module too, it makes a lot more sense than multiple source sets and will allow for a cleaner build system.

I think that's probably it? Although I definitely think this needs a look over from everyone else it seems great to me.

Generators/src/main/java/net/minestom/code_generation/AllGenerators.java
```diff
@@ -0,0 +1,112 @@
+package net.minestom.code_generation;
```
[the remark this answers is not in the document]
Articdive: The main reasoning to give it code_generation is that using "generated" or "generation" could be misinterpreted with terrain generation for example. I can change it but since this package is in it's own subproject I don't think this is a necessary change  but it's a simple refractor in IntelliJ if you insist we can change this :).

Generators/src/main/java/net/minestom/code_generation/MinestomCodeGenerator.java
```diff
@@ -0,0 +1,26 @@
...
+    public static final File DEFAULT_OUTPUT_FOLDER = new File("src/autogenerated/java");
+
+    public abstract void generate();
+
+    protected void writeFiles(@NotNull List<JavaFile> fileList, File outputFolder) {
+        for (JavaFile javaFile : fileList) {
+            try {
+                javaFile.writeTo(outputFolder);
+            } catch (IOException e) {
+                // TODO: Better error handling here.
```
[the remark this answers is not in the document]
Articdive: A simple message output should do here, we handle many exceptions similarly and if something breaks here the source code won't even compile.

Generators/src/main/java/net/minestom/code_generation/AllGenerators.java
```diff
@@ -0,0 +1,112 @@
...
+import net.minestom.code_generation.sound.SoundGenerator;
+import net.minestom.code_generation.statistics.StatisticGenerator;
+
+import java.io.File;
+
+public class AllGenerators {
+    public static void main(String[] args) {
+        String targetVersion;
+        if (args.length < 1) {
+            System.err.println("Usage: <MC version> [target folder]");
```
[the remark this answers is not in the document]
Articdive: Fixed
Commit called 'Use Logger instead of System.out in code generation.'

src/main/java/net/minestom/server/entity/pathfinding/PFBlockDescription.java
Articdive: This can be done using the Hitbox actually which would be better for the behaviour. But all the hitbox code is down to @TheMode. I don't even know if I am extracting the correct hitbox data.

Generators/src/main/java/net/minestom/code_generation/blocks/BlockEntityGenerator.java
```diff
@@ -0,0 +1,207 @@
...
+                        block.get("id").getAsString()
+                );
+            }
+            // Add to static init.
+            staticBlock2.addStatement("$T.BLOCK_ENTITY_REGISTRY.register($N)", registryClassName, blockEntityName);
+        }
+
+        blockEntityClass.addStaticBlock(staticBlock.build());
+        blockEntityClass.addStaticBlock(staticBlock2.build());
+        blockEntityClass.addStaticBlock(staticBlock3.build());
```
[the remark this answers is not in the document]
Articdive: In-theory they could all be in the same static block but it does look more organized and easier to follow. In Bytecode it's all the same anyway. Let me know if you really think it's necessary.

Generators/src/main/java/net/minestom/code_generation/blocks/BlockEntityGenerator.java
```diff
@@ -0,0 +1,207 @@
...
+
+        // Write files to outputFolder
+        writeFiles(
+                Collections.singletonList(
+                        JavaFile.builder("net.minestom.server.instance.block", blockEntityClass.build())
+                                .indent("    ")
```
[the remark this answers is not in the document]
Articdive: I did spend a while looking for something and tried to configure spotless [here](https://github.com/diffplug/spotless) but it didn't end up working too well.

Generators/src/main/java/net/minestom/code_generation/item/MaterialGenerator.java
```diff
@@ -0,0 +1,316 @@
...
+
+import javax.lang.model.element.Modifier;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public final class MaterialGenerator extends MinestomCodeGenerator {
```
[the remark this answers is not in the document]
Articdive: Well in code we store Items using their Materials in reality this is generating the Material class so no?

src/main/java/net/minestom/server/raw_data/RawMaterialData.java
Articdive: The idea is that this data is loaded from the json file which is non final. So plugins can change the hardness of a block for example by entering this class. Now the class itself should probably be final (not extendable) but the fields are supposted to be accessible. Maybe it is better to use getters and setters. 

Would it cause problems:
- Really just depends on what your changing and how you are changing it.

Generators/src/main/java/net/minestom/code_generation/sound/SoundGenerator.java
```diff
@@ -0,0 +1,181 @@
...
+
+import javax.lang.model.element.Modifier;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public final class SoundGenerator extends MinestomCodeGenerator {
```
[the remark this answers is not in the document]
Articdive: Done in 'Rename SoundGenerator to SoundEventGenerator'

Articdive:
For the CI to work we will need to add the creation of "eula.txt" to the GitHub action. This shouldn't be a problem.

Generators/src/main/java/net/minestom/code_generation/blocks/BlockEntityGenerator.java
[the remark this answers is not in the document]
kezz: Nah I think it's fine if it all ends up the same bytecode.

TheMode:
The new data generator is currently used in the block branch.

PR #238, opened 2021-04-12
Server list ping overhaul (dynamic responses, multiple versions, open to LAN, UT3 query)

### LINKED ISSUE #256: Support 'Ping via LAN'
As an optional feature, perhaps in the `extras` package, the server could send the Ping via LAN packet to allow people in the local network to join without the need for entering the IP manually, etc. This could also be useful for testing purposes.

kezz:
This PR adds a new class `NamedAndIdentified` to indicate an object provides a `Component` name and a `UUID` identity. It also provides a simple implementation class alongside utility methods to construct an instance of `NamedAndIdentified` from static methods.

This is also implemented in the `ResponseData` class and other methods. These now use an interface to generate the response data for the client. This means that all ping versions can be responded to with one event.

Finally, this PR also adds the `OpenToLAN` extra class for creating a repeating ping that will display the server in the LAN section of the server list for servers on the local network, taken from #257, and will close #256.


thiccaxe:
[ScoreboardLine](https://github.com/Minestom/Minestom/blob/cdaef2b4f0c05bd9ebaed4a73e056697df56f50a/src/main/java/net/minestom/server/scoreboard/Sidebar.java#L272) (A possible use case) currently uses a String ID - Possibly changing that to a UUID or maybe even parameterizing the Identified interface - like the Named.

src/main/java/net/minestom/server/network/packet/client/status/StatusRequestPacket.java
```diff
@@ -27,15 +30,22 @@ public void process(@NotNull PlayerConnection connection) {
...
         // Call event
         ServerListPingEvent statusRequestEvent = new ServerListPingEvent(responseData, connection);
         MinecraftServer.getGlobalEventHandler().callCancellableEvent(ServerListPingEvent.class, statusRequestEvent,
                 () -> {
-                    ResponsePacket responsePacket = new ResponsePacket();
-                    responsePacket.jsonResponse = responseData.build().toString();
+                    final ResponsePacket responsePacket = new ResponsePacket();
+
+                    // check if we need to use a legacy response
+                    if (connection.getProtocolVersion() >= 713) {
```
BomBardyGamer: Would probably be a bit clearer if this was a constant, e.g. `LEGACY_PROTOCOL_VERSION`. Also, due to there being two sets of versioning (versioning was reset for the Netty rewrite), so this could be confused with the new versions, couldn't it?

BomBardyGamer: Nevermind me, looks like this has been solved

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1354,6 +1374,7 @@ public String getUsername() {
      */
     public void setUsernameField(@NotNull String username) {
         this.username = username;
+        this.usernameComponent = null;
```
TheMode, line 1377: Should probably create the component here, I do not think that it is worth the lazy evaluation

src/main/java/net/minestom/server/extras/lan/OpenToLANConfig.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Sets the delay between outgoing pings.
+     *
+     * @param delay the delay
+     * @return {@code this}, for chaining
+     */
+    @Contract("_ -> this")
+    public @NotNull OpenToLANConfig setDelayBetweenPings(@NotNull UpdateOption delay) {
```
TheMode, line 48: `#setDelay` would probably be enough

kezz: Well we need to make it clear the difference between the delay between pings and the delay between event calls.

BomBardyGamer: Perhaps `#setPingDelay` would be better then?

BomBardyGamer: Also, if this is a sort of builder, then I think it would be a bit nicer if you omitted the `set` prefixes from these, as we aren't really changing state in the same way we would with a regular object, since we're using this object to build a new object.

src/main/java/net/minestom/server/extras/query/response/QueryResponse.java
```diff
@@ -0,0 +1,17 @@
+package net.minestom.server.extras.query.response;
+
+import net.minestom.server.utils.binary.BinaryWriter;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A query response.
+ */
+public interface QueryResponse {
```
TheMode, line 9: Should extends `Writable`

TheMode (changes requested):

PR #242, opened 2021-04-14
Acquirable API

TheMode:
Draft of the API described here: https://wiki.minestom.com/release-candidates/acquirable-api

The goal is to have a single and clear API for accessing entities safely, no matter how threads are used (per instance, chunk, chunk group, etc...).

This PR only introduces the API, a latter one will come to change all the current signatures (e.g. `Entity` getters will be replaced to `Acquirable<Entity>`)


src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -613,6 +614,14 @@ public static void setGroupedPacket(boolean groupedPacket) {
         MinecraftServer.groupedPacket = groupedPacket;
     }
 
+    public static boolean hasWaitMonitoring() {
```
kezz, line 617: Docs for this?

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1559,6 +1567,11 @@ private UpdateOption getSynchronizationCooldown() {
         return Objects.requireNonNullElse(this.customSynchronizationCooldown, SYNCHRONIZATION_COOLDOWN);
     }
 
+    @Override
+    public @NotNull <T> Acquirable<T> getAcquiredElement() {
```
kezz, line 1571: If we're overriding this anyway, couldn't `LockedElement` be generic so it returns `Aquireable<Entity>`?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -325,6 +326,27 @@ public void update(long time) {
             packet.process(this);
         }
 
+        //System.out.println(getAcquiredElement().getHandler().getBatchThread());
```
kezz, line 329: Should this block be removed?

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param consumer   the consumer called for each of the collection element
+     * @param <E>        the object type
+     */
+    public static <E> void acquireForEach(@NotNull Collection<Acquirable<E>> collection,
+                                          @NotNull Consumer<? super E> consumer) {
+        final Thread currentThread = Thread.currentThread();
+        Map<BatchThread, List<E>> threadCacheMap = retrieveThreadMap(collection, currentThread, consumer);
+
+        // Acquire all the threads one by one
+        {
```
kezz, line 39: Why is this scoped?

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+ * Represents an element which can be acquired.
+ * Used for synchronization purpose.
+ *
+ * @param <T> the acquirable object type
+ */
+public final class Acquirable<T> {
+
+    public static final ThreadLocal<Collection<Entity>> CURRENT_ENTITIES = ThreadLocal.withInitial(Collections::emptyList);
+
+    public static @NotNull Collection<@NotNull Entity> currentEntities() {
```
kezz, line 24: Should be documented.

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+        return CURRENT_ENTITIES.get();
+    }
+
+    @ApiStatus.Internal
+    public static void refreshEntities(@NotNull Collection<@NotNull Entity> entities) {
```
kezz, line 29: Same here.

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+        CURRENT_ENTITIES.set(entities);
+    }
+
+    private final T value;
+    private final Handler handler;
+
+    public Acquirable(@NotNull T value) {
```
kezz, line 36: And here!

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Executes {@code consumer} only if this element can be safely
+     * acquired without any synchronization.
+     *
+     * @param consumer the acquisition consumer
+     * @return true if the acquisition happened without synchronization,
+     * false otherwise
+     */
+    public boolean tryAcquire(@NotNull Consumer<@NotNull T> consumer) {
```
kezz, line 61: This could have a method that doesn't require a consumer that returns the acquired element or null.

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+     * Gets the {@link Handler} of this acquirable element,
+     * containing the currently linked thread.
+     *
+     * @return this element handler
+     */
+    public @NotNull Handler getHandler() {
+        return handler;
+    }
+
+    public static class Handler {
```
kezz, line 103: Is there any reason why this couldn't be inside Acquireable? Should be documented anyway.

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+        public void refreshChunkEntry(@NotNull ThreadProvider.ChunkEntry chunkEntry) {
+            this.chunkEntry = chunkEntry;
+        }
+
+        public BatchThread getBatchThread() {
+            return chunkEntry != null ? chunkEntry.getThread() : null;
+        }
+    }
+
+}
```
kezz, line 121: Why no line ending 🤨

src/main/java/net/minestom/server/lock/Acquirable.java
```diff
@@ -0,0 +1,121 @@
...
+    @ApiStatus.Internal
+    public static void refreshEntities(@NotNull Collection<@NotNull Entity> entities) {
+        CURRENT_ENTITIES.set(entities);
+    }
+
+    private final T value;
+    private final Handler handler;
+
+    public Acquirable(@NotNull T value) {
+        this.value = value;
```
kezz, line 37: Could be worth wrapping this in a non null check.

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public static long getCurrentWaitMonitoring() {
+        return WAIT_COUNTER_NANO.get();
+    }
+
+    public static void resetWaitMonitoring() {
+        WAIT_COUNTER_NANO.set(0);
+    }
+
+    private static class ScheduledAcquisition {
```
kezz, line 170: Can't this generified?

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+                // The element is manager in a different thread, cache it
+                List<E> threadCacheList = threadCacheMap.computeIfAbsent(elementThread, batchThread -> new ArrayList<>());
+                threadCacheList.add(value);
+            }
+        }
+
+        return threadCacheMap;
+    }
+
+    public static long getCurrentWaitMonitoring() {
```
kezz, line 162: Should be documented, same below.

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+            T element = (T) obj;
+            final E value = element.unwrap();
+
+            final BatchThread elementThread = element.getHandler().getBatchThread();
+            if (currentThread == elementThread) {
+                // The element is managed in the current thread, consumer can be immediately called
+                consumer.accept(value);
+            } else {
+                // The element is manager in a different thread, cache it
+                List<E> threadCacheList = threadCacheMap.computeIfAbsent(elementThread, batchThread -> new ArrayList<>());
```
kezz, line 154: You don't need to assign this, just call the add directly afterwards.

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    private static <E, T extends Acquirable<E>> Map<BatchThread, List<E>> retrieveThreadMap(@NotNull Collection<? super T> collection,
+                                                                                            @NotNull Thread currentThread,
+                                                                                            @NotNull Consumer<? super E> consumer) {
+        // Separate a collection of acquirable elements into a map of thread->elements
+        // Useful to reduce the number of acquisition
+
+        Map<BatchThread, List<E>> threadCacheMap = new HashMap<>();
+        for (Object obj : collection) {
```
kezz, line 144: Why do you down cast and immediately upcast again afterwards?

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Acquires a {@link Collection}.
+     * <p>
+     * Order is not guaranteed.
+     *
+     * @param collection the collection to acquire
+     * @param consumer   the consumer called for each of the collection element
+     * @param <E>        the object type
+     */
+    public static <E> void acquireForEach(@NotNull Collection<Acquirable<E>> collection,
```
kezz, line 33: I wonder if this could take a consumer on a collection of E so that it could be grouped together. For example, the retrieve thread map method could collect the ones in the current thread and run them all at once. This method can still exist and overload the other one if needs be.

src/main/java/net/minestom/server/lock/Acquisition.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public static void resetWaitMonitoring() {
+        WAIT_COUNTER_NANO.set(0);
+    }
+
+    private static class ScheduledAcquisition {
+        private final List<Acquirable<Object>> acquirableElements = new ArrayList<>();
+        private final Map<Object, List<Consumer<Object>>> callbacks = new HashMap<>();
+    }
+}
```
kezz, line 174: Do you have the wrong line endings on?

src/main/java/net/minestom/server/lock/LockedElement.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents an element that have a {@link Acquirable} linked to it.
+ * <p>
+ * Useful if you want to provide an access point to an object without risking to compromise
+ * the thread-safety of your code.
+ */
+public interface LockedElement {
```
kezz, line 11: This should definitely be generic, feels so wrong having it not be..

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    public boolean contains(Object o) {
+        try {
+            return collectionView.contains(toViewFunction.apply((E) o));
+        } catch (ClassCastException e) {
+            return false;
+        }
+    }
+
+    @NotNull
```
kezz, line 52: Should inline the annotations.

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    public boolean isEmpty() {
+        return collectionView.isEmpty();
+    }
+
+    @Override
+    public boolean contains(Object o) {
+        try {
+            return collectionView.contains(toViewFunction.apply((E) o));
+        } catch (ClassCastException e) {
```
kezz, line 47: Wouldn't this be better with an instance of check? This might throw an NPE otherwise.

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public boolean add(E e) {
+        return collectionView.add(toViewFunction.apply(e));
+    }
+
+    @Override
+    public boolean remove(Object o) {
+        try {
```
kezz, line 89: Same here.

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Override
+    public boolean removeAll(@NotNull Collection<?> c) {
+        boolean changed = false;
+        try {
+            for (Object e : c) {
+                if (remove(e))
+                    changed = true;
+            }
+        } catch (ClassCastException ignored) {
```
kezz, line 128: This will never be thrown as it's caught in the remove method

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+
+        return true;
+    }
+
+    @Override
+    public boolean addAll(@NotNull Collection<? extends E> c) {
+        boolean changed = false;
+        try {
+            for (Object e : c) {
+                if (add((E) e))
```
kezz, line 111: Why do you downcast and upcast here?

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+        boolean changed = false;
+        try {
+            for (Object e : c) {
+                if (!contains(e)) {
+                    remove(e);
+                    changed = true;
+                }
+            }
+        } catch (ClassCastException ignored) {
+        }
```
kezz, line 145: Same here, this will never be thrown.

src/main/java/net/minestom/server/utils/collection/CollectionView.java
```diff
@@ -0,0 +1,178 @@
...
+
+        return changed;
+    }
+
+    @Override
+    public void clear() {
+        this.collectionView.clear();
+    }
+
+    public static class IteratorView<T, V> implements Iterator<T> {
```
kezz, line 155: This should be in its own class with docs too.

src/main/java/net/minestom/server/lock/AcquirableCollection.java
```diff
@@ -0,0 +1,24 @@
+package net.minestom.server.lock;
+
+import net.minestom.server.utils.collection.CollectionView;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+import java.util.function.Consumer;
+
+public class AcquirableCollection<E extends LockedElement> extends CollectionView<E, Acquirable<E>> {
```
kezz, line 9: Part of me thinks we should just be using Iterable here. We don't need the benefits afforded by the collection interface and it's much simpler.

src/main/java/net/minestom/server/thread/BatchThread.java
```diff
@@ -0,0 +1,94 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class BatchThread extends Thread {
```
kezz, line 14: This all needs documenting.

src/main/java/net/minestom/server/thread/PerChunkThreadProvider.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public class PerChunkThreadProvider extends ThreadProvider {
+
+    public PerChunkThreadProvider(int threadCount) {
+        super(threadCount);
+    }
+
+    @Override
+    public long findThread(@NotNull Chunk chunk) {
+        return chunk.hashCode();
```
kezz, line 17: This isn't a random thread right?

src/main/java/net/minestom/server/thread/PerInstanceThreadProvider.java
```diff
@@ -4,61 +4,17 @@
...
-            }
-        })));
-        return futures;
-    }
-
-    private Set<Chunk> getChunks(Instance instance) {
-        return instanceChunkMap.computeIfAbsent(instance, inst -> ConcurrentHashMap.newKeySet());
+    public long findThread(@NotNull Chunk chunk) {
+        return chunk.getInstance().hashCode();
     }
```
kezz, line 19: Do these classes even implement hash code?

src/main/java/net/minestom/server/thread/ThreadProvider.java
```diff
@@ -1,252 +1,301 @@
...
-    /**
-     * Executes an entity tick for object entities in an instance's chunk.
-     *
-     * @param instance the chunk's instance
-     * @param chunk    the chunk
-     * @param time     the current time in ms
-     */
-    protected void updateObjectEntities(@NotNull Instance instance, @NotNull Chunk chunk, long time) {
-        conditionalEntityUpdate(instance, chunk, time, entity -> entity instanceof ObjectEntity);
+    public synchronized void refreshThreads(long tickTime) {
```
kezz, line 172: These and the following methods need documentation.

src/main/java/net/minestom/server/thread/ThreadProvider.java
```diff
@@ -1,252 +1,301 @@
...
+        int counter = 0;
+        while (true) {
+            Chunk chunk = chunks.pollFirst();
+            if (!ChunkUtils.isLoaded(chunk)) {
+                removeChunk(chunk);
+                return;
+            }
+
+            // Update chunk threads
+            {
```
kezz, line 199: Why is this scoped?

kezz (changes requested):
Looks brilliant to me! Just a bunch of relatively minor changes but everything looks great structurally :)

src/main/java/net/minestom/server/entity/Entity.java
TheMode: But then how could you acquire a `Player`?
I dislike the current safe as it is pretty unsafe, but I didnt find any alternative

src/main/java/net/minestom/server/entity/Player.java
TheMode: Will be once the PR is ready, for now it is pretty useful to test performance

src/main/java/net/minestom/server/thread/PerChunkThreadProvider.java
TheMode: Indeed, otherwise the chunk thread would keep changing over time which is not ideal. I guess that I could replace it to the identity hashcode directly

src/main/java/net/minestom/server/thread/PerInstanceThreadProvider.java
TheMode: Nope, but I assumed that people might want to override it

src/main/java/net/minestom/server/thread/ThreadProvider.java
TheMode: Just an easy way for me to differentiate the different parts of a thread refresh, and to say that this method is independent of the rest

src/main/java/net/minestom/server/lock/LockedElement.java
TheMode: I agree, this was the case at first but inheritance is an isssue

src/main/java/net/minestom/server/thread/PerChunkThreadProvider.java
kezz: That might be better. I think not relying on the identity hash code is always good if we can.

src/main/java/net/minestom/server/UpdateManager.java
```diff
@@ -39,8 +39,8 @@
 
     {
         // DEFAULT THREAD PROVIDER
-        //threadProvider = new PerGroupChunkProvider();
         threadProvider = new PerInstanceThreadProvider();
+        //threadProvider = new PerChunkThreadProvider();
```
kezz, line 43: This is a breaking change. I rely on the comment being above not below - this will break my workflow please revert.

src/main/java/net/minestom/server/entity/acquirable/AcquirableEntity.java
```diff
@@ -0,0 +1,150 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+/**
+ * Represents an element which can be acquired.
```
kezz, line 16: Should read "Represents an **entity** which can be acquired

src/main/java/net/minestom/server/UpdateManager.java
TheMode: Sorry, this must be done for the future of the project

src/main/java/net/minestom/server/entity/acquirable/AcquirableEntity.java
```diff
@@ -143,7 +143,7 @@ public void refreshChunkEntry(@NotNull ThreadProvider.ChunkEntry chunkEntry) {
             this.chunkEntry = chunkEntry;
         }
 
-        public BatchThread getBatchThread() {
+        public TickThread getBatchThread() {
```
kezz, line 146: Should this be renamed to `getTickThread`?

src/main/java/net/minestom/server/utils/consumer/EntityConsumer.java
```diff
@@ -1,8 +1,9 @@
 package net.minestom.server.utils.consumer;
 
 import net.minestom.server.entity.Entity;
+import org.jetbrains.annotations.NotNull;
 
 @FunctionalInterface
 public interface EntityConsumer {
```
kezz, line 7: Why does this need to exist? Can't we just use `Consumer<Entity>`?

TheMode: Because it would be impossible to override the method signature to `Consumer<Player>`

src/main/java/net/minestom/server/entity/acquirable/AcquirableEntity.java
TheMode: Effectively

PR #244, opened 2021-04-14
Compound and list ItemTags

LeoDog896:
This adds compound and list item tags.


LeoDog896:
Added the deep cloning, should I open an issue on the nbt lib or here for immutable nbt lists/compounds?

kezz (approved):
Looks good to me - the internals can always be swapped out later.

TheMode:
Couldn't we instead have an `ItemTag#Nbt` doing the cloning, and later on adding the immutable versions of compound & list?

PR #257, opened 2021-04-22
Add OpenToLAN extra class

### LINKED ISSUE #256: Support 'Ping via LAN'

kezz:
This PR closes #256, adding the `OpenToLAN` extra class. For example, running `OpenToLan#open()` will make the server show up in the LAN section of the server, as seen in the screenshot below. This also supports a component description, that will be downscaled to legacy section sign formatting when being sent.

Screenshot from 2021-04-22 17-17-56

### DIFF
--- a/src/main/java/net/minestom/server/extras/lan/OpenToLAN.java
+++ b/src/main/java/net/minestom/server/extras/lan/OpenToLAN.java
@@ -1,0 +1,189 @@
+package net.minestom.server.extras.lan;
+
+import java.io.IOException;
+import java.net.*;
+
+import java.nio.charset.StandardCharsets;
+
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.format.TextColor;
+import net.kyori.adventure.text.serializer.gson.GsonComponentSerializer;
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+import net.minestom.server.MinecraftServer;
+import net.minestom.server.timer.Task;
+import net.minestom.server.utils.time.TimeUnit;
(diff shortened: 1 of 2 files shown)


TheMode:
Shouldn't it use #238 for the description?
Having a new setter feels a bit unnecessary

src/main/java/net/minestom/server/extras/lan/OpenToLAN.java
```diff
@@ -0,0 +1,192 @@
...
+     * @throws SocketException if an error occurred
+     * @see OpenToLAN
+     */
+    public static boolean openWithException(int port) throws SocketException {
+        if (socket != null) {
+            return false;
+        } else {
+            socket = new DatagramSocket(port);
+            task = MinecraftServer.getSchedulerManager().buildTask(OpenToLAN::ping)
+                    .repeat(15, TimeUnit.SECOND)
```
TheMode, line 112: The wikivg page says 1.5, not 15

kezz: jesus... maybe i should make this configurable lmao

kezz:
> Shouldn't it use #238 for the description?
> Having a new setter feels a bit unnecessary

Potentially, but this would either involve calling the `ServerListPingEvent` every 1.5 seconds or reverting back to the old static consumer method to obtain the description.

If #238 is merged, I could add this as an entry in `ServerListPingVersion`?

kezz:
Merged into #238

PR #259, opened 2021-04-24
Item Events

LeoDog896:
This allows handling events on a namespace level, allowing events on items to work.


src/main/java/net/minestom/server/item/ItemMetaBuilder.java
```diff
@@ -192,6 +194,21 @@
         return this;
     }
 
+    @Beta
+    @Contract(value = "_ -> this")
+    public @NotNull ItemMetaBuilder addItemEvents(@NotNull String... ids) {
```
kezz, line 199: Do you think it could be worth using namespace ids here?

src/main/java/net/minestom/server/item/event/GlobalItemEventManager.java
```diff
@@ -0,0 +1,23 @@
...
+public class GlobalItemEventManager {
+
+    private static final ConcurrentHashMap<String, ItemEventHandler> events = new ConcurrentHashMap<>();
+
+    public GlobalItemEventManager() {
+
+    }
+
+    public ItemEventHandler get(String id) {
+        if (events.containsKey(id)) return events.get(id);
```
kezz, line 14: This could use one of those compute methods instead of this

src/main/java/net/minestom/server/item/event/GlobalItemEventManager.java
```diff
@@ -0,0 +1,23 @@
+package net.minestom.server.item.event;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+public class GlobalItemEventManager {
+
+    private static final ConcurrentHashMap<String, ItemEventHandler> events = new ConcurrentHashMap<>();
```
kezz, line 7: Should this not be static?

src/main/java/net/minestom/server/item/event/ItemEventHandler.java
```diff
@@ -0,0 +1,33 @@
...
+public class ItemEventHandler implements EventHandler {
+
+    private final Map<Class<? extends Event>, Collection<EventCallback>> eventCallbacks = new ConcurrentHashMap<>();
+    private final Map<String, Collection<EventCallback<?>>> extensionCallbacks = new ConcurrentHashMap<>();
+
+    ItemEventHandler(String namespace) {
+
+    }
+
+    @NotNull
```
kezz, line 22: Nullness annotations should be inlined

src/main/java/net/minestom/server/item/event/ItemEventHandler.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public Map<Class<? extends Event>, Collection<EventCallback>> getEventCallbacksMap() {
+        return eventCallbacks;
+    }
+
+    @NotNull
+    @Override
+    public Collection<EventCallback<?>> getExtensionCallbacks(String extension) {
```
kezz, line 30: This could have an overloaded method to take an extension or a discovered extension

kezz (changes requested):
Looking good so far!

src/main/java/net/minestom/server/item/event/ItemEventHandler.java
LeoDog896: Probably should be part of EventHandler instead.

src/main/java/net/minestom/server/item/event/GlobalItemEventManager.java
LeoDog896: Outdated :sweat_smile:

src/main/java/net/minestom/server/item/event/GlobalItemEventManager.java
LeoDog896: Outdated :sweat_smile:

src/main/java/net/minestom/server/item/event/ItemEventRegistry.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Finds an ItemEventIdentifierHandler by the identifier value, and creates a new instance if none were found.
+     *
+     * @param identifier The identifier used to find the ItemEventIdentifierHandler
+     *
+     * @return The found ItemEventIdentifierHandler with a new instance if none were found.
+     */
+    public @NotNull ItemEventIdentifierHandler<T> identifierOrNew(T identifier) {
+        if (identifierHandlerHashMap.containsKey(identifier)) return identifierHandlerHashMap.get(identifier);
```
TheMode, line 26: Couldn't it be replaced by `Map#putIfAbsent`?

src/main/java/net/minestom/server/item/event/ItemEvents.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Gets an existing ItemEventRegistry, or creates one if none was found.
+     *
+     * @param tag The tag used to find the ItemEventRegistry. Used for general identification.
+     * @param <T> The type of Identifier required for the ItemTag
+     *
+     * @return An ItemEventRegistry instance. Will create a new one if none were found.
+     */
+    public static <T> @NotNull ItemEventRegistry<T> getRegistryOrNew(ItemTag<T> tag) {
+        if (events.containsKey(tag)) return events.get(tag);
```
TheMode, line 40: Same as above

src/main/java/net/minestom/server/item/event/ItemEventRegistry.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Finds an ItemEventIdentifierHandler by the identifier value, and creates a new instance if none were found.
+     *
+     * @param identifier The identifier used to find the ItemEventIdentifierHandler
+     *
+     * @return The found ItemEventIdentifierHandler with a new instance if none were found.
+     */
+    public @NotNull ItemEventIdentifierHandler<T> identifierOrNew(T identifier) {
```
TheMode, line 25: Is there really a need for the `orNew` part? same for getRegistryOrNew

TheMode (commented):
I am still a bit unsure about how we should handle events based on where they can be listened to, though it still feels a step in the right direction

src/main/java/net/minestom/server/item/event/ItemEventRegistry.java
LeoDog896: It is mainly meant to be explicit naming what it does, though identifier could work.

src/main/java/net/minestom/server/item/event/ItemEventRegistry.java
LeoDog896: Map#putIfAbsent's return value is null, meaning the check still has to occur resulting in essentially the same code.

LeoDog896: Moved to Map#computeIfAbsent :+1:

PR #261, opened 2021-04-25
Add UniquelyIdentified interface

kezz:
This PR adds a simple interface to get a UUID from a given object which is uniquely identified by the said UUID.


TheMode:
Why not `#getUuid`? Feels a bit like method duplication

mworzala:
I would also say `getUuid` or `getUniqueId` would be better.

PR #262, opened 2021-04-26
Particle API

Bloepiloepi:
This pull request adds a particle api.


Bloepiloepi:
Shapes are now a separate library, data will go into the new-data branch

PR #263, opened 2021-04-26
Entity safety API

TheMode:
The goal is to change current methods signature to use the Acquirable API instead of raw entity objects

### DIFF
--- a/src/main/java/net/minestom/server/Viewable.java
+++ b/src/main/java/net/minestom/server/Viewable.java
@@ -1,13 +1,14 @@
 package net.minestom.server;
 
 import net.kyori.adventure.audience.Audience;
+import net.minestom.server.acquirable.AcquirableCollection;
 import net.minestom.server.adventure.audience.PacketGroupingAudience;
 import net.minestom.server.entity.Player;
 import net.minestom.server.network.packet.server.ServerPacket;
 import net.minestom.server.utils.PacketUtils;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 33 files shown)


src/main/java/net/minestom/server/Viewable.java
```diff
@@ -101,6 +101,6 @@ default void sendPacketToViewersAndSelf(@NotNull ServerPacket packet) {
      * @return the audiences
      */
     default @NotNull Iterable<? extends Audience> getViewersAsAudiences() {
-        return this.getViewers();
+        return this.getViewers().unwrap().collect(Collectors.toList());
```
kezz, line 104: I'm not sure about this, the only thing audience will ever do is directly send packets - is there a way we could access this directly without having to unwrap?

kezz (commented):
I'll give it a proper look over some point this week, but it feels like there's a lot of unwrapping happening for simple packet sending, which should be thread safe right?

src/main/java/net/minestom/server/Viewable.java
TheMode: `unwrap` means that there is no acquisition happening

kezz: Yeah but there must be a better way than going from a collection to a stream and back to a list every time the use just wants to access thread safe methods like this. Could we use the collector view class for this?

TheMode: We could also use `#iterator` https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/BaseStream.html#iterator()
it does however add an ugly warning. And I do not expect this to have a huge performance impact, could be wrong

PR #265, opened 2021-04-28
Particle API

Bloepiloepi:
This pull request adds the particle api.

Particle has been renamed to ParticleType, particle data has been added, and some methods to send particles have been added to Player.


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -947,6 +949,92 @@ public void stopSound() {
...
+     * @param y            y position of the particle
+     * @param z            z position of the particle
+     * @param offsetX      random x offset of the particle
+     * @param offsetY      random y offset of the particle
+     * @param offsetZ      random z offset of the particle
+     * @param speed        speed of the particle
+     * @param count        amount of times this particle is shown
+     */
+    public void sendParticle(@NotNull Particle particle, boolean longDistance, double x, double y, double z,
+                             float offsetX, float offsetY, float offsetZ, float speed, int count) {
```
kezz, line 967: All this might be nicer to have in a builder. This could be done with providing a vector or x/y/z for the location as well.

kezz (approved):
LGTM! Just a minor comment which might be worth considering.

src/main/java/net/minestom/server/entity/Player.java
Bloepiloepi: We could have this in a builder, but it may be better to just have some of those values in the Particle class? Like Particle.setOffset(float, float, float) and the same for count and longDistance.

kezz: Yeah that's probably a better idea - I'd definitely prefer them being in the class than in a super long method.

TheMode: If the plan is to have a builder, I would prefer `Particle` to be immutable

Bloepiloepi: Yeah it isn't right now, but I can change that

Bloepiloepi: The problem is that there are several Particle classes (for different kinds of data) so how would the builder construct the right one? There is Particle.block(BlockState) for example, which could return BlockParticleBuilder, but it is not really clean to make a separate builder for each type

src/autogenerated/java/net/minestom/server/particle/ParticleType.java
```diff
@@ -0,0 +1,292 @@
...
+        Registry.PARTICLE_REGISTRY.register(WHITE_ASH);
+    }
+
+    @NotNull
+    private final NamespaceID id;
+
+    @NotNull
+    private final BiFunction<ParticleType<T>, @Nullable String, T> dataReader;
+
+    protected ParticleType(@NotNull NamespaceID id,
```
TheMode, line 246: Since `ParticleType` contains both the type and the data, wouldnt it make more sense to keep the `Particle` name?

src/autogenerated/java/net/minestom/server/particle/ParticleType.java
```diff
@@ -0,0 +1,292 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.function.BiFunction;
+
+/**
+ * AUTOGENERATED by ParticleTypeGenerator
+ */
+public class ParticleType<T extends Particle> implements Keyed {
```
TheMode, line 20: I am unsure about the decision to use a generic here, in which situation would it help?

src/main/java/net/minestom/server/particle/data/BlockParticle.java
```diff
@@ -0,0 +1,48 @@
...
+import net.minestom.server.particle.ParticleType;
+import net.minestom.server.registry.Registry;
+import net.minestom.server.utils.binary.BinaryWriter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.function.BiFunction;
+
+public class BlockParticle extends Particle {
+    public static final BiFunction<ParticleType<BlockParticle>, @Nullable String, BlockParticle> READER = (particle, data) -> {
```
TheMode, line 13: All particle reader seems to be fairly weak, and will likely become an issue for packet reading as they currently rely on string

src/main/java/net/minestom/server/particle/data/Particle.java
```diff
@@ -0,0 +1,143 @@
...
+        return new Particle(particleType);
+    }
+
+    /**
+     * Creates a {@link BlockParticle} with type block.
+     *
+     * @param state the blockstate for the particle
+     * @return the block particle with its data
+     */
+    public static @NotNull BlockParticle block(BlockState state) {
```
TheMode, line 99: Is it possible to obtain a builder for those special particle effects?

TheMode (changes requested):
I am currently uncertain about the ease of use and flexibility of the API

src/autogenerated/java/net/minestom/server/particle/ParticleType.java
Bloepiloepi: The generic is there to make sure you can't create, for example, a block particle effect with an item particle. You can only create particles without data if the type is without data, and you can only create data if the particle type needs that data.

src/main/java/net/minestom/server/particle/data/Particle.java
Bloepiloepi: Yes, you can create a normal builder and use buildBlock(). This can change, but it was the best solution I could find while not having 4 types of builder.
I decided to leave Particle.block() still there, just as a shortcut if you don't need to modify any other data.

src/autogenerated/java/net/minestom/server/particle/ParticleType.java
Bloepiloepi: The ParticleType doesn't contain the data but it does contain the reader to get the data. I decided to do that because you can't get the generic class (like particleType.getDataType().getReader()). The reader is essentially just a "pointer" to the data class.

src/main/java/net/minestom/server/particle/data/BlockParticle.java
Bloepiloepi: This is true, I could not find a way to have an argument change it's type dependent on the particle type. ArgumentParticle has to be divided into 2 arguments, and the second one should change it's type based on the input of the first one. The second one also shouldn't be sent to the client, because minecraft:particle has this feature already. If this is possible, it would be a better idea.

src/autogenerated/java/net/minestom/server/particle/ParticleType.java
TheMode: Perhaps that it would make more sense to store the particle data inside the type, so the particle effect contains the particle effect and additional information such as position/count. Would also allow better parsing for particle data and the ability to expand it with mods.

Bloepiloepi: When you have the particle data inside the type, how would you change the data? There can be different Particles, with different data but the same type, so how would that work?

Bloepiloepi: The Particle class is what really represents a particle, ParticleType is just the type

TheMode: Could add some methods to create a new `Particle` object with the required data

Bloepiloepi:
@TheMode Can you check the latest commit? I want to make sure this is actually what you want before finishing it.

src/autogenerated/java/net/minestom/server/particle/Particle.java
```diff
@@ -0,0 +1,396 @@
...
+
+import java.util.List;
+import java.util.Scanner;
+
+/**
+ * AUTOGENERATED by ParticleTypeGenerator
+ */
+public class Particle implements Keyed, PublicCloneable<Particle> {
+
+    public static final Particle AMBIENT_ENTITY_EFFECT = new Particle(NamespaceID.from("minecraft:ambient_entity_effect"));
```
kezz, line 27: It feels *really* odd having these in the same place as the `Particle` class. Are we sure that this is what we want? There should absolutely be a separation between the types of the particle and the particle data for an individual "instance" of a particle. Not just for separation of concerns, but also because we don't have much control over the auto-generated code, meaning that changes will be really hard to make.

kezz (changes requested):

src/autogenerated/java/net/minestom/server/particle/Particle.java
Bloepiloepi: I agree, but @TheMode wanted to not have the particle type and its data separated. So I am not really sure how I could make it better without recreating my first approach.

PR #271, opened 2021-05-03
Update DataImpl#clone to actually clone the data

KrystilizeNevaDies:


src/main/java/net/minestom/server/data/DataImpl.java
```diff
@@ -61,7 +61,13 @@ public boolean isEmpty() {
     @NotNull
     @Override
     public DataImpl clone() {
-        return (DataImpl) super.clone();
+        DataImpl newData = new DataImpl();
```
TheMode, line 64: Should still use `super.clone()` for polymorphism

KrystilizeNevaDies: Well that removes the actual cloning as the super implementation just returns the same value.

Asthereon: The `super.clone()` just returns a reference to the same object, so modifying the returned DataImpl causes updates to the object that was cloned which defeats the purpose of a clone method.

Asthereon: After doing a bit more digging, I think this is due to the default `clone()` for HashMaps being a shallow copy, causing the DataImpl to get identical references rather than a deep clone of the ConcurrentHashMaps.

TheMode:
Close in favor of the new tag API.

PR #272, opened 2021-05-03
Revamped event listening API

TheMode:
Code highly inspired from [lucko's helper library](https://github.com/lucko/helper/wiki/helper:-Events).

The goal would be to enforce listeners stored as constants instead as lambda to facilitate their removal or just general management. And perhaps a way to merge multiple listeners in a `Group` to sort them by feature.


src/main/java/net/minestom/server/event/Event.java
```diff
@@ -1,11 +1,94 @@
...
+            private final Class<T> eventType;
+
+            private List<Function<T, Boolean>> filters = new ArrayList<>();
+            private Consumer<T> handler;
+
+            private Builder(Class<T> eventType) {
+                this.eventType = eventType;
+            }
+
+            public Builder<T> filter(Function<T, Boolean> filter) {
```
kezz, line 65: This can be a `Predicate`.

src/main/java/net/minestom/server/event/Event.java
```diff
@@ -1,11 +1,94 @@
...
+                    if (!filters.isEmpty()) {
+                        if (filters.stream().anyMatch(tBooleanFunction -> !tBooleanFunction.apply(event))) {
+                            // Cancelled
+                            return;
+                        }
+                    }
+
+                    // Handler
+                    if (handler != null) {
+                        handler.accept(event);
```
kezz, line 87: Do we want to allow people to create listeners with no actual handler?

src/main/java/net/minestom/server/event/Event.java
```diff
@@ -1,11 +1,94 @@
...
+    public static <T extends EntityEvent> Listener.Builder<T> entity(Class<T> eventType) {
+        return new Listener.Builder<>(eventType);
+    }
+
+    public static <T extends PlayerEvent> Listener.Builder<T> player(Class<T> eventType) {
+        return new Listener.Builder<>(eventType);
+    }
+
+
+    public static class Listener<T extends Event> {
```
kezz, line 29: This should probably end up in its own class at some point, especially if we end up adding an expiry/cancellation system.

src/main/java/net/minestom/server/event/Event.java
```diff
@@ -1,11 +1,94 @@
...
+        private final Class<T> eventType;
+        private final Set<EventHandler> attach = new CopyOnWriteArraySet<>();
+        private final Consumer<T> combined;
+
+        private Listener(Class<T> eventType, @NotNull Consumer<T> combined) {
+            this.eventType = eventType;
+            this.combined = combined;
+        }
+
+        public void attachTo(@NotNull EventHandler handler) {
```
kezz, line 40: I'm not so sure about this. Why do we need to keep a set of the handlers?

src/main/java/net/minestom/server/event/Event.java
TheMode: I assume that we could throw an exception, but just doing nothing doesn't seem like a terrible solution either. Perhaps a warning

src/main/java/net/minestom/server/event/Event.java
TheMode: So we can have a method to detach it from all the handlers. Would be really powerful once listeners group are implemented

src/main/java/net/minestom/server/event/Event.java
TheMode: Perhaps, a bit afraid that it will end up confusing with it and `EventCallback`

src/main/java/net/minestom/server/event/Event.java
```diff
@@ -1,11 +1,94 @@
...
+import java.util.function.Predicate;
 
 /**
  * Object which can be listened to by an {@link EventHandler}.
  * <p>
  * Called using {@link EventHandler#callEvent(Class, Event)}.
  */
 public class Event {
+
+    public static <T extends EntityEvent> Listener.Builder<T> entity(Class<T> eventType) {
```
mworzala, line 20: Why only entity and player event? What about something like server list ping?

mworzala (commented):
I'm a fan of the system, will be happy to see grouping and cancellation

src/main/java/net/minestom/server/event/Event.java
kezz: I think throwing an exception might be the best idea - I can think of zero reason to create a listener that doesn't do anything so it's most likely a developer mistake.

kezz:
Some ideas for meta:
- Cancelling an event after a number of calls calls
- Cancelling an event after a certain duration
- Making an event only fire a certain amount of times for each player/entity/etc

PR #273, opened 2021-05-04
Avoid NPE when getting NBT in ItemTag

LeoDog896:
This stores an internal reference stopping an NPE from being thrown when getting NBT.


LeoDog896:
Old:

```
try {
 return item.get(ItemTag.NBT...
} catch (e: Exception) {
 return null;
}
```

New:
```
return item.get(ItemTag.NBT...
```

No external references changed either :+1:

PR #275, opened 2021-05-05
Respect client chat settings

### LINKED ISSUE #245: Sent messages do not respect the client's chat settings
The current chat system does not respect the client chat settings as sent in the client settings packet. This should include both incoming and outgoing messages. Further information can be found [here](https://wiki.vg/Chat#Processing_chat).

kezz:
This PR moves all incoming and outgoing chat messages to use a new `Messenger` class that respects the client's chat settings because for some unknown reason Mojang didn't do this client side. 

Closes #245.


src/main/java/net/minestom/server/utils/Action.java
```diff
@@ -0,0 +1,17 @@
+package net.minestom.server.utils;
+
+/**
+ * A functional interface to perform an action.
+ */
+@FunctionalInterface
+public interface Action {
```
TheMode, line 7: Why not `Runnable`?

TheMode (changes requested):

PR #278, opened 2021-05-07
Add weather API

kezz:
This PR adds a weather system. The weather system has two main components, the `Weather` object and the `WeatherManager`.

To demo the system, check out the new `/weather` command.

### Weather
Weather consists of a series of components:

- The type of the weather, either rain, thunder or clear.
- The instant in which the weather will end (if any)
- The rain strength (for rain and thunder)
- The thunder strength (for thunder).

This can be created using static methods like `Weather.clear()` and `Weather.rain(Duration.of(10, ChronoUnit.MINUTE))` or using the builder which allows for finer control over the output.

### Managers
A weather manager is a simple class to keep track of the weather for a given object. These objects are either global, instances or players. This means that a player can have weather that is unique to other players which will fall back to the instance's weather when the weather runs out. Setting the weather in an instance will propagate the changes to all players that don't have an overriding weather of their own.

### Containers
Instances and players are provided with a `WeatherContainer` interface that allow you to query or set the weather directly on the object using methods like `player.setWeather(Weather.clear())` or `instance.resetWeather()`.

Additionally, the global weather manager, obtained using `MinecraftServer.getGlobalWeatherManager()` can also be used to set the weather globally. The default weather can also be provided, which will be set to the user when they join.


Articdive:
Replaced with #576.

PR #280, opened 2021-05-10
Don't default to fixed yaw

Kebab11noel:
Some application might require the knowledge of the absolute player rotation, this pr removes the default fix and instead provides the #getFixedYaw() method


TheMode:
Might want to be retargeted at the new coordinate API.

PR #282, opened 2021-05-11
Equipment improvements

Bloepiloepi:
- Added EntityEquipEvent to replace ArmorEquipEvent (it now works with mainhand and offhand as well)
- Moved inner class EntityEquipmentPacket.Slot to outer class EquipmentSlot
- Added `EquipmentSlot.fromAttributeSlot()` to simplify applying item attributes
- Added `EquipmentSlot.isHand()` and `EquipmentSlot.isArmor()`


Bloepiloepi:
> which might be a bit difficult as it cannot be solely be handled inside `#safeItemInsert`

I tested it and this is true, the click in the players inventory needs to be cancelled as well, and also items are duplicated when you pick them up.

TheMode:
Should probably revert the last commit

PR #283, opened 2021-05-12
Make lastPosition useful

Kebab11noel:
Introduced a filed containing the nano time of last position change and #lastPosition is now updated after every tick


Kebab11noel:
`lastPosition` field will be removed when #303 gets merged making this PR irrelevant.

PR #286, opened 2021-05-14
Fixes a crash when the server has only one available processor.

R0bbyYT:
The pull request fixes a crash when the server has only one available processor.


TheMode:
Oops, I did use `Math#min` instead of `max`.
Should be fixed on master

R0bbyYT:
Has been fixed on the master.

PR #293, opened 2021-05-17
Tag API

TheMode:
Basically `ItemTag` but applicable to more classes, with the additions of 3 interfaces

### DIFF
--- a/src/main/java/net/minestom/server/item/ItemTag.java
+++ b/src/main/java/net/minestom/server/item/ItemTag.java
@@ -1,112 +1,65 @@
 package net.minestom.server.item;
 
+import net.minestom.server.tag.Tag;
 import org.jetbrains.annotations.NotNull;
 import org.jglrxavpok.hephaistos.nbt.NBT;
 import org.jglrxavpok.hephaistos.nbt.NBTCompound;
-import org.jglrxavpok.hephaistos.nbt.NBTList;
 
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
-public class ItemTag<T> {
+/**
+ * @deprecated use {@link Tag}.
+ */
+@Deprecated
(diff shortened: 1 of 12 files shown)


src/main/java/net/minestom/server/item/ItemTag.java
```diff
@@ -1,112 +1,65 @@
...
-        this.writeConsumer.accept(nbtCompound, value);
+    public static @NotNull Tag<Byte> Byte(@NotNull String key) {
+        return Tag.Byte(key);
     }
 
-    public static @NotNull ItemTag<Byte> Byte(@NotNull String key) {
-        return new ItemTag<>(key,
-                nbtCompound -> nbtCompound.getByte(key),
-                (nbtCompound, value) -> nbtCompound.setByte(key, value));
+    public static @NotNull Tag<Short> Short(@NotNull String key) {
```
kezz, line 25: This will be a breaking change - is there any chance we could have `ItemTag` be some sort of fake wrapper around `Tag`? Or get this to return `ItemTag`?

src/main/java/net/minestom/server/tag/TagReader.java
```diff
@@ -0,0 +1,48 @@
+package net.minestom.server.tag;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jglrxavpok.hephaistos.nbt.NBTCompound;
+
+/**
+ * Represents an element which can read {@link Tag tags}.
+ */
+public interface TagReader {
```
kezz, line 10: I wonder if this should be named `TagReadable`. `TagReader` to me implied something which reads tags, not something from which tags can be read.

kezz (approved):
Looks great to me! Just a minor comment about the class name. Not sure if we want to care too much about this being a breaking change, but it's worth considering.

PR #295, opened 2021-05-18
Comamnd permission

c-128:
Added command permission, its used like this:
Example Command:
```java
public class Command extends Command {
    public Command() {
        super("command");

        setPermission("commands.examplecommand");
    }
}
```

Set Permission Error:
```java
MinecraftServer.getCommandManager().setNotEnoughPermissionCallback((sender, command) -> {
    // Handle your not enough permission thing
});
```


c-128:
I think I messed it up, I need to go onto another branch

PR #296, opened 2021-05-18
Added Command Permission.

### LINKED ISSUE #253: Permissions for commands
Commands should have permissions that can be assigned to them that will prevent command senders from accessing the permission unless they have the required permission.

c-128:
Added command permission, its used like this:
Example Command:
```java
public class Command extends Command {
    public Command() {
        super("command");

        setPermission("commands.examplecommand");
    }
}
```

Set Permission Error:
```java
MinecraftServer.getCommandManager().setNotEnoughPermissionCallback((sender, command) -> {
    // Handle your not enough permission thing
});
```

Closes #253.

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandManager.java
+++ b/src/main/java/net/minestom/server/command/CommandManager.java
@@ -37,6 +37,7 @@ public final class CommandManager {
     private final CommandDispatcher dispatcher = new CommandDispatcher();
 
     private CommandCallback unknownCommandCallback;
+    private CommandCallback notEnoughPermissionCallback;
 
     public CommandManager() {
     }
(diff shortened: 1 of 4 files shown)


kezz (approved):
LGTM! I wonder if we should allow individual commands to specify their own no permission callbacks in addition to the global default.

c-128:
Thats a good idea!
I will work on a callback in the Command class, and if no callback is spexified, it will run the default one

kezz (approved):
Looks great - thanks for the PR! Happy for this to be merged unless anyone else has any changes.

TheMode:
Why couldn't it be part of `#setCondition`? It seems to have the same purpose

c-128:
Wouldn't that cause that you need to write the not enouph permission message for every command?

kezz:
> Why couldn't it be part of `#setCondition`? It seems to have the same purpose

I think having it separate from the condition system makes a lot of sense. Commands having permissions is something that 99.9% of command authors will use and it being distinct means that the methods used to set the permission are clear and easy to find, without having to either provide it manually using a condition or delegating to some default no-permission callback stored somewhere in the code.

TheMode:
> > Why couldn't it be part of `#setCondition`? It seems to have the same purpose
> 
> I think having it separate from the condition system makes a lot of sense. Commands having permissions is something that 99.9% of command authors will use and it being distinct means that the methods used to set the permission are clear and easy to find, without having to either provide it manually using a condition or delegating to some default no-permission callback stored somewhere in the code.

I am not really against having a simple way to specify permission, but mostly about adding a whole new category of condition. It could be possible to specify a permission inside ‘#setCondition’ instead

KrystilizeNevaDies:
Closed due to inactivity. Feel free to open again.

PR #298, opened 2021-05-18
Send correct number of blocks inside a section,

LeoDog896:
Before, this value used to be 200, and now it dynamically counts based on when blocks are set.


Uraxys:
It shouldn't count air, cave air and void air, https://wiki.vg/Chunk_Format#Data_structure.
I guess Block#isAir() could be used to check if it's air.

PR #299, opened 2021-05-18
Send the real amount of blocks in a chunk section.

Uraxys:
Tested it locally by logging the amount of blocks in a section and it updated correctly when placing / removing blocks.

### DIFF
--- a/src/main/java/net/minestom/server/instance/palette/Section.java
+++ b/src/main/java/net/minestom/server/instance/palette/Section.java
@@ -53,6 +53,8 @@ public class Section implements PublicCloneable<Section> {
     private int valuesPerLong;
     private boolean hasPalette;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/instance/palette/Section.java
```diff
@@ -91,6 +96,13 @@ public void setBlockAt(int x, int y, int z, short blockId) {
             block ^= clear << bitIndex;
             block |= (long) blockId << bitIndex;
 
+            final boolean isCurrentAir = Block.fromStateId(getBlockAt(x, y, z)).isAir();
```
TheMode, line 99: Hmm I am not really confident performance-wise. Would probably be better to retrieve the state id from the `block` value above

PR #302, opened 2021-05-21
Add resource loading API

### LINKED ISSUE #106: Resource management

### LINKED ISSUE #252: Get data folder for extension

mworzala:
This adds a simple api to read files from the extension jar resources, as well as preference over files in the extension directory.

The obvious use case here is config files.

(there was an issue with the origin repository and the pr needed to be recreated)

resolves #106 
resolves #252

### DIFF
--- a/src/main/java/net/minestom/server/extensions/DiscoveredExtension.java
+++ b/src/main/java/net/minestom/server/extensions/DiscoveredExtension.java
@@ -11,6 +11,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


kezz (approved):
LGTM!

PR #304, opened 2021-05-23
Food animation fix

Bloepiloepi:
This pull request fixes 2 issues related to food:

- Fixed other players still being able to see eating particles when a player stops eating
- Fixed eating animation being shown when PlayerPreEatEvent is cancelled

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -2299,7 +2299,10 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
      * @param allowFood true if food should be updated, false otherwise
      * @return the called {@link ItemUpdateStateEvent},
      * null if there is no item to update the state
+     *
+     * @deprecated Use {@link #callItemUpdateStateEvent(Hand)} instead
      */
+    @Deprecated
     public @Nullable ItemUpdateStateEvent callItemUpdateStateEvent(boolean allowFood, @Nullable Hand hand) {
         if (hand == null)
             return null;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1136,6 +1136,15 @@ public boolean isEating() {
         return eatingHand != null;
     }
 
+    /**
+     * Gets the hand which the player is eating from.
+     *
+     * @return the eating hand, null if none
+     */
+    public Hand getEatingHand() {
```
kezz, line 1144: This should be annotated with `@Nullable`

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2287,19 +2296,14 @@ public void refreshEating(@Nullable Hand eatingHand) {
      * Used to call {@link ItemUpdateStateEvent} with the proper item
      * It does check which hand to get the item to update.
      *
-     * @param allowFood true if food should be updated, false otherwise
      * @return the called {@link ItemUpdateStateEvent},
      * null if there is no item to update the state
      */
-    public @Nullable ItemUpdateStateEvent callItemUpdateStateEvent(boolean allowFood, @Nullable Hand hand) {
+    public @Nullable ItemUpdateStateEvent callItemUpdateStateEvent(@Nullable Hand hand) {
```
kezz, line 2302: Not sure about this breaking change, can we leave the method overloading this one but with a deprecation notice?

kezz (changes requested):
Not tested by me but this looks pretty good! Just left a few comments.

kezz (approved):

Issue #250, opened 2021-04-20
Split click types for dragging into multiple for different mouse buttons


stefvanschie:
Currently `ClickType.START_DRAGGING` and `ClickType.DRAGGING` are the only two click types related to dragging in an inventory. From these it's not clear which mouse button was used for the dragging (i.e., left mouse button, right mouse button or middle mouse button). The suggestion is to split up these two click types into multiple ones for each mouse button, similar to how there is a `ClickType.LEFT_CLICK` and a `ClickType.RIGHT_CLICK`. For example, `ClickType.START_LEFT_DRAGGING`, `ClickType.LEFT_DRAGGING`, `ClickType.START_RIGHT_DRAGGING`, `ClickType.RIGHT_DRAGGING`, `ClickType.START_MIDDLE_DRAGGING`, `ClickType.MIDDLE_DRAGGING`. This way servers and extensions can easily identify which type of dragging has or is occurring and respond appropriately to each one.

lexikiq:
I believe this issue is fixed

TheMode:
Indeed, closing.

PR #305, opened 2021-05-23
Changing item drag events

stefvanschie:
This addresses issue #250.

### Changes
This pull requests splits `ClickType.START_DRAGGING` and `ClickType.DRAGGING` into separate enum entries for both left and right mouse button. The old `ClickType.START_DRAGGING` and `ClickType.DRAGGING` have been marked as deprecated. These are also no longer used in any of Minestom; instead the new entries are used.

### Considerations
I have not added javadocs to any of the new enum entries, since this wasn't done for any of the other entries in the `ClickType` enum either. If wanted, javadocs can be added for those.

The old entries are marked as deprecated. This is to ensure that old code (e.g., custom servers, extensions) will still be compatible with this change and not give any missing fields errors. Nonetheless, code that relied upon these old click types will no longer encounter these, since only the newer values are used in, for example, events. This may lead to (unintended) changes. Other options include dropping the old fields entirely without deprecation; or firing multiple events, one with the old click type and one with the new one (this may as a side effect mean that certain actions are executed twice). The current approach is a middle ground I settled on, but can be changed if wanted.


src/main/java/net/minestom/server/inventory/click/InventoryClickProcessor.java
```diff
@@ -241,9 +241,13 @@ public InventoryClickResult dragging(@Nullable Inventory inventory, @NotNull Pla
             if (button == 0) {
                 // Start left
                 this.leftDraggingMap.put(player, new IntOpenHashSet());
+
+                clickResult = startCondition(inventory, player, slot, ClickType.START_LEFT_DRAGGING, clicked, cursor);
```
TheMode, line 245: Shouldn't all those new conditions check for cancellation?

stefvanschie: I'm not quite sure. The original code had this line at the declaration of the variable and doesn't seem to do any checks regarding cancellation either. Only when looping over the slots is a check done, which is still there.

The `startCondition` method does seem to do some checks itself regarding cancellation (namely setting cancel and refresh to `true` on the click result), so perhaps that is enough?

PR #308, opened 2021-06-01
Convenience toMilliseconds, add UpdateOption methods to TaskBuilder

LeoDog896:
This allows the usage of UpdateOption for delay and repeat settings in TaskBuilder.


kezz (approved):
LGTM. Personally, I don't think we should be adding *more* `UpdateOption` API but until a day when it gets replaced with `Duration` this seems fine.

mworzala (approved):

PR #313, opened 2021-06-07
Fix potion effect IDs.

Articdive:


code-generators/src/main/java/net/minestom/codegen/potion/PotionEffectGenerator.java
```diff
@@ -100,8 +100,8 @@ public void generate() {
                         .returns(potionEffectClassName)
                         .addAnnotation(Nullable.class)
                         .addParameter(TypeName.SHORT, "id")
-                        .beginControlFlow("if(id >= 0 && id < VALUES.length)")
-                        .addStatement("return VALUES[id]")
+                        .beginControlFlow("if(id >= 0 && id < VALUES.length + 1)")
```
kezz, line 103: Should this be `>= 1`?

Articdive: Eh, in this fix here it wasn't done: https://github.com/Minestom/Minestom/commit/38bcb755c284c891437c5da5b7026f373a962aec

PR #316, opened 2021-06-11
Updated Server List Ping

thiccaxe:
Added playersHidden field to ResponseData; shows "???" in Vanilla.
image

Added event for ping/pong packet, which is cancellable

### DIFF
--- a/src/main/java/net/minestom/server/ping/ResponseData.java
+++ b/src/main/java/net/minestom/server/ping/ResponseData.java
@@ -29,6 +29,7 @@ public class ResponseData {
     private int online;
     private Component description;
     private String favicon;
+    private boolean playersHidden;
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/net/minestom/server/ping/ResponseData.java
```diff
@@ -294,6 +296,25 @@ public void clearEntries() {
...
+    public void setPlayersHidden(boolean playersHidden) {
+        this.playersHidden = playersHidden;
+    }
+
+    /**
+     * Returns if the players are hidden or not.
+     *
+     * @return if the players are hidden
+     */
+    public boolean isPlayersHidden() {
```
kezz, line 314:
```suggestion
    public boolean arePlayersHidden() {
```

kezz (approved):
LGTM! Just a minor nitpick.

PR #318, opened 2021-06-12
Add an option to start the server with no terminal

mworzala:
Useful in cases where user input will never be provided. Setting the vm arg `minestom.terminal.disabled` will cause the server to start without a terminal.

The change is extremely minor (terminal defaults to on, just has a vm arg to disable it), I just want to confirm there are no huge objections for any reason.


kezz (approved):
LGTM. Does anything error out without the terminal being present? For example, sending a message to the console/etc. Don't see why it would though, but worth checking.

mworzala:
No, it just ignores messages.

Issue #314, opened 2021-06-10
Less generic name for Instance


xt449:
The naming for Instance object and those relating to it is ridiculous.
There a field for DimensionType yet the programmer has not been informed of any Dimension objects. There is also a WorldBorder field; so clearly the names here are not consistant.

To help those coming from other APIs we could use World or more accurately WorldContainer, similarly Dimension or DimensionContainer, or to accommodate the abstractness that is this class it could simply be a Region.

Whatever is chosen, please don't use the most vague possible class name.

xt449:
What about World? Is there some fundamental difference for why this name has not been used?

TheMode:
I feel that World could also become confusing.
To give a simple example, let's say you have a single map but divided into 5 to dispatch players (often seen in MMORPG). Would you say that there are 5 worlds or instances?

Not saying that our naming is perfect, we can still take feedback

xt449:
Although, Minestom doesn't use regions or the region file format so that seems to be a decent name. But since it is already defined as something else in the Minecraft community, I can understand that too.

How about RegionContainer? This would be troublesome with the current InstanceContainer, so its still not ideal, but getting there.

I think ***Universe*** has potential though since it does not have any other name conflicts.

xt449:
The name that Mojang uses (Level) is also available, albeit worse than Dimension.
Could also use Realm, but it is already a Minecraft term so I can understand not wanting to try that.

However, the more I've been thinking, World is just such a easy name, even if it's not a perfect description it fits well enough and has the bonus of being recognizable by developers from other Minecraft modding APIs

xt449:
Not sure if this is close enough on topic, but why is the block package inside the instance package?

xt449:
@TheMode how about Zone? That's what you mean by instance right?
I can't find a meaning or similar word that makes sense for instance, https://www.merriam-webster.com/dictionary/instance

TheMode:
instance, world, level, dimension, universe, space, zone

There are a lot of possible names but I do not think that we should focus on it. There are a lot of more important issues

xt449:
Okay

xt449:
I'll go work on those then

PR #320, opened 2021-06-13
Refactor while we still can

xt449:
Refactored all "Instance" classes to "World".

Moved net.minestom.server.instance.block to net.minestom.server.block
Moved net.minestom.server.instance to net.minestom.server.world

Refactored most if not all relevant fields, variables, methods, and Javadocs referencing Instance objects to now reference there respective World objects

EDIT:
Also replaced ambiguous Javadocs containing "... for instance ..." with "... for example ..."


xt449:
Resolves #314

kezz (changes requested):
I'm against this PR and the changes it makes. Both because `World` is, in my opinion, worse than `Instance` but also I think such a change would be better suited for a 1.0 to 2.0 version bump.

xt449:
Missing instance of instanced instance from instance manager: no instances found in Minestom instance of instances manager (for instance SharedInstance).

LeoDog896:
I also agree with @kezz on this one -- Instance is definitely not the best name, but I definitely do not think World is a better replacement.

PR #322, opened 2021-06-13
Added InventoryHolder interface which allows any class to contain an inventory

xt449:
This implements https://trello.com/c/b6Ofyz4b/99-entity-inventory


mworzala (commented):
I have no issue with this concept generally, but I think it would go better with the other inventory changes (sections) which are being considered.

As a general note for PRs, it is best to bring it up in an issue or Discord before you start work on it. We want everyone to be on the same page about additions so that nobody ends up wasting their time :)

PR #325, opened 2021-06-16
Switch internal ExtensionManager workings to NIO

LeoDog896:


src/main/java/net/minestom/server/extensions/ExtensionManager.java
```diff
@@ -324,27 +328,25 @@ private Extension loadExtension(@NotNull DiscoveredExtension discoveredExtension
...
-
-        if (fileList != null) {
-            // Loop through all files in extension folder
-            for (File file : fileList) {
-
-                // Ignore folders
-                if (file.isDirectory()) {
-                    continue;
+        try {
+            Files.walk(extensionFolder).forEach(file -> {
```
kezz, line 332: Do we want to walk this with a `maxDepth` of `1` to ensure we only take files in the root directory? Also, it might be worth using `walkFileTree` instead just to avoid the stream if we're not using it.

kezz (changes requested):
Definitely support the use of nio, just a few minor comments regarding implementation.

Issue #199, opened 2021-03-29
Open Deprecations


mworzala:
Below is a table of the deprecated methods or groups of methods and roughly when they are planned for removal.

This Table does not cover all deprecations, and other methods may still be removed without warning, for example on Minecraft version updates, as the API is not yet stable..

None of these features should be used, since they will be removed at some point. If you are unsure about what to replace your usage of these deprecated features with, come and chat to us in the [Discord](https://discord.gg/rV24CzG).

| Deprecation                                                           | Reason            | Replacement                            | Removal        | See Also                                                                                                                                        |
|-----------------------------------------------------------------------|-------------------|----------------------------------------|----------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
| Chat package (including `JsonMessage` and related methods/subclasses) | No longer needed  | Adventure API                          | Estimated 1.17 | https://wiki.minestom.com/feature/adventure                                                                                                     |
| `SoundCategory` enum                                                  | No longer needed  | Adventure API                          | Estimated 1.17 | https://wiki.minestom.com/feature/adventure#sound                                                                                               |
| Argument types `Long` and `Entities`                                  |                   |                                        | Estimated 1.17 |                                                                                                                                                 |
| `ResponseData#setName(String)`                                        | Misleading        | `#setVersion(String)`                  |                |                                                                                                                                                 |
| `ResponseDataConsumer`                                                | Replaced by event | ServerListPingEvent                    | 1.17           | [ServerListPingEvent](https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/event/server/ServerListPingEvent.java) |
| `ResponseData` methods for `Player` and `PingPlayer`                  | No longer needed  | `NamedAndIdentified` and entry methods | 1.17           |                                                                                                                                                 |
| `AdventureSerializer`                                                 | No longer needed  | The standard Adventure serializers     |                |                                                                                                                                                 |

Geolykt:
I would mark them depreated for removal via the Deprecation annotation since we are already using Java 11

mworzala:
They are, was mostly for us to keep track of what needs to be removed and when.

Geolykt:
They aren't in master, if you are not understanding what I mean, I mean the `@Depreacted(forRemoval = true)` annotation

mworzala:
> They aren't in master, if you are not understanding what I mean, I mean the `@Depreacted(forRemoval = true)` annotation

[...]

PR #326, opened 2021-06-16
Refactor and remove classes from entity package.

LeoDog896:


mworzala:
IMO this should be done with the other deprecations in one big removal. I think everything in #199 can go except maybe the sound category stuff (cc @kezz)?

PR #328, opened 2021-06-16
Update book meta to use components

kezz:
This PR updates both `WritableBookMeta` and `WrittenBookMeta`. The former has its title and author parts removed, as these do not exist in writable books. The latter is converted to use components.

This is a breaking change for both classes.


KrystilizeNevaDies:
This is not really relevant anymore, we have currect functionality for this. Feel free to reopen.

PR #331, opened 2021-06-16
Set default UUID provider to Mojang UUIDs

LeoDog896:
This sets the default UUID provider to use the UUIDs from MojangAuth.


LeoDog896:
Use MojangAuth instead

Issue #191, opened 2021-03-27
Cannot send singular title/subtitle/times with Adventure


kezz:
### What is the current behavior?
As `Title` does not accept null components, there is no way to set individual elements of a title without resending the whole set of title packets.

### Additional information / Possible thoughts on why this bug is happening
This is not intended behaviour - the vanilla client does not have the possibility to set individual parts of a title without setting the entire element. Additionally, in 1.17 this will be hard coded. It's perhaps best to leave this as it is and mark with `wontfix` to prep for 1.16 and make the legacy chat methods of sending titles not delegate to the Adventure method.

kezz:
Waiting on https://github.com/KyoriPowered/adventure/issues/314.

TheMode:
> This is not intended behaviour - the vanilla client does not have the possibility to set individual parts of a title without setting the entire element. Additionally, in 1.17 this will be hard coded. It's perhaps best to leave this as it is and mark with wontfix to prep for 1.16 and make the legacy chat methods of sending titles not delegate to the Adventure method.

After checking the snapshot code, it appears that the only protocol difference is that each title component (title/subtitle/actionbar/time) will get their own packets instead of sharing a single code. Meaning that the behavior is unlikely to change

PR #332, opened 2021-06-17
Remove deprecated JsonMessage classes

LeoDog896:
This finally removes all of the JsonMessage methods plaguing the codebase.


kezz (approved):
This PR looks good to me, assuming it all compiles/runs/etc. Before merging would we want a temporary fix for #191 implemented? This PR would remove the ability to send individual title parts without jumping into packets. There'll probably be an Adventure specific way of doing this in 4.9.0 so maybe we just leave it with packets for now anyway..

TheMode:
Removed in the current block branch

PR #333, opened 2021-06-19
Coordinate API

TheMode:
- Common interface for Position/BlockPosition/Vector
- Immutable, to become record and then primitive class


src/main/java/net/minestom/server/utils/incubator/Pos.java
```diff
@@ -0,0 +1,70 @@
...
+        this.yaw = yaw;
+        this.pitch = pitch;
+    }
+
+    public Pos(double x, double y, double z) {
+        this(x, y, z, 0, 0);
+    }
+
+    public Pos() {
+        this(0, 0, 0, 0, 0);
```
kezz, line 25: I'd prefer these to be in static methods so we can return `ZERO` in the case where the input represents the same position.

src/main/java/net/minestom/server/utils/incubator/Pos.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Contract(pure = true)
+    public @NotNull Pos withYaw(float yaw) {
```
kezz, line 29: Should probably have checks to see if the values are actually changing, otherwise we can just return the current pos.

kezz (changes requested):
I love the idea and immutable positions/vectors is brilliant. I wonder if we should instead be typing the vector class similar to other game engines (e.g. `Vec3i` for int, `Vec3d` for doubles, etc). Also, I'm not sure about the combination of interface with implementation and a standard class. I think it would be better for them both to be interfaces imo.

src/main/java/net/minestom/server/utils/incubator/Pos.java
TheMode: I could perhaps just remove this constructor. Though I think that we should decide by only having static methods or constructors, as it will affect this class being a `record` or not.

MrGazdag:
I would voice some concerns about positions being immutable, namely that vanilla Minecraft uses a mutable BlockPos class, but allocates new BlockPos instances too many times. Right now, I am trying to write optimized code which uses the fewest possible Vectors, to eliminate this issue. Part of this is reusing Vector instances. I would at least recommend a mutable variant of Point.

TheMode:
> I would voice some concerns about positions being immutable, namely that vanilla Minecraft uses a mutable BlockPos class, but allocates new BlockPos instances too many times. I think making it immutable would produce similarly high allocation counts. Right now, I am trying to write optimized code which uses the fewest possible Vectors, to eliminate this issue. Part of this is reusing Vector instances. I would at least recommend a mutable variant of Point.

Immutability provides the guarantee that the object can be reused without any side-effect, and can trigger some JVM optimizations (namely scalar replacement)

See https://www.reddit.com/r/java/comments/ocaj6j/128bit_floatingpoint_arithmetic_for_java/h3tdzmh?utm_source=share&utm_medium=web2x&context=3

In the future, these classes will become `primitive record` (once we finally get Valhalla) which will completely the potential performance penalty.

The codebase is currently filled by Vector/Position clones that cannot be optimized away, nothing is reused, and in case of entity having to check every tick if the position object has been modified.

TheMode:
See #355

PR #334, opened 2021-06-19
Ensure proper error when extension.json is missing

LeoDog896:
This ensures that a NullPointerException won't be thrown but rather a readable exception telling the user that they forgot an extension.json in their Extension jar.


mworzala (approved):
looks good

kezz (approved):

PR #335, opened 2021-06-21
Implementation of a PlayerPostChatEvent

Marlonzockt:
Adds a PlayerPostChatEvent which gets called after a chat message has been sent.

### DIFF
--- a/src/main/java/net/minestom/server/event/AbstractPlayerChatEvent.java
+++ b/src/main/java/net/minestom/server/event/AbstractPlayerChatEvent.java
@@ -1,0 +1,41 @@
+package net.minestom.server.event;
+
+import net.kyori.adventure.text.Component;
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.trait.PlayerEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public interface AbstractPlayerChatEvent extends PlayerEvent {
+
+    /**
+     * Those are the players who will receive the message.
+     * <p>
+     * It can be modified to add or remove recipient.
+     *
+     * @return a modifiable list of message targets
+     */
+    @NotNull Collection<Player> getRecipients();
+
+    /**
+     * Gets the message sent.
+     *
+     * @return the sender's message
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/event/AbstractPlayerChatEvent.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.trait.PlayerEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public interface AbstractPlayerChatEvent extends PlayerEvent {
```
KrystilizeNevaDies, line 13: Renaming to IPlayerChatEvent would probably be better for consistency

KrystilizeNevaDies (commented):
Looks good, just rename interface

src/main/java/net/minestom/server/event/AbstractPlayerChatEvent.java
TheMode: I personally prefer `AbstractPlayerChatEvent`, but I am not sure about inheritance in events. Seems like a huge mess and there is likely a better solution to handle Pre/Post listeners

Marlonzockt: So what should I do then?

kezz: Something we discussed in discord was the ability to register an event as a "post" listener. Basically the event is fired and then the result is returned to the caller, at the same time a new thread is spun up to call the event for anyone listening to it in a "post" way. This way, we would add support for async monitoring for stuff like chat, movement, etc - without having to block the actual action itself.

Marlonzockt: Which would make my PR useless?

TheMode: If we decide going this way yes, api changes require a bit more thought than implementation code

Marlonzockt: Yeah that's fine, I just wanted to give you any option

PR #336, opened 2021-06-22
Implement some Adventure methods into ConsoleSender

kezz:
As `ConsoleSender` is also an `Audience`, it should implement some other common Adventure methods to increase cross-platform compatibility and prep for potential upcoming Adventure Audience type checking.


src/main/java/net/minestom/server/command/CommandSender.java
```diff
@@ -1,6 +1,9 @@
 package net.minestom.server.command;
 
+import java.util.UUID;
```
TheMode, line 3: Unused import?

kezz: Appears to be, I'll update this PR tomorrow at some point.

PR #338, opened 2021-06-26
Changes to Hologram API

Moulberry:
Update Hologram.java to support subclasses overriding metadata & y offset
Create MarkerHologram.java, hologram that uses the marker flag in order to allow clients to interact through the Armor Stand


Moulberry:
Update BlockPlacementListener.java to allow block placement for marker Armor Stands

PR #343, opened 2021-06-28
Debug grid (/dg) command

Kebab11noel:
This PR adds a new command: **/dg [radius] [center]**, which generates a checkerboard. This is useful for testing entity movement since it makes easier to see block borders. For easy usage both parameters are optional, defaulting to 100 block radius and centering below the player's feet.
**The default checkerboard**
100x100 checkerboard

### DIFF
--- a/src/test/java/demo/Main.java
+++ b/src/test/java/demo/Main.java
@@ -55,6 +55,7 @@ public class Main {
         commandManager.register(new SummonCommand());
         commandManager.register(new RemoveCommand());
         commandManager.register(new GiveCommand());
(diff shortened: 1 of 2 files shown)


kezz (approved):

TheMode:
What about a new chunk generator instead?

Kebab11noel:
> What about a new chunk generator instead? Which could perhaps add various elements for testing purpose

I thought about it, but i think it can be irritating to look at it for long period if it isn't necessary, hence why i decided it's better to use a two character command.

KrystilizeNevaDies:
Updated: (#1809)

PR #345, opened 2021-06-29
Add ArgumentSoundEvent to ArgumentType

LeoDog896:
This adds a sound event argument to the ArgumentType list, to allow for selecting custom/normal sounds client side.

### DIFF
--- a/src/main/java/net/minestom/server/command/builder/arguments/minecraft/registry/ArgumentSound.java
+++ b/src/main/java/net/minestom/server/command/builder/arguments/minecraft/registry/ArgumentSound.java
@@ -41,6 +41,6 @@ public class ArgumentSound extends ArgumentRegistry<Sound.Type> {
 
     @Override
     public String toString() {
-        return String.format("SoundEvent<%s>", getId());
+        return String.format("Sound<%s>", getId());
     }
 }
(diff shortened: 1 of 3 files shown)


kezz:
If this argument accepts custom sounds, returning a `SoundEvent` is wrong as that only supports built-in sounds. Probably worth either changing this to return `Sound.Type`, which can either be a `SoundEvent` or just a custom sound with a key. This would allow people to `instanceof` to check if they have a `SoundEvent` or not.

src/main/java/net/minestom/server/command/builder/arguments/minecraft/registry/ArgumentSound.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minestom.server.command.builder.NodeMaker;
+import net.minestom.server.command.builder.arguments.minecraft.SuggestionType;
+import net.minestom.server.entity.EntityType;
+import net.minestom.server.network.packet.server.play.DeclareCommandsPacket;
+import net.minestom.server.registry.Registries;
+import net.minestom.server.sound.SoundEvent;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents an argument giving an {@link EntityType}.
```
kezz, line 14:
```suggestion
 * Represents an argument giving an {@link Sound.Type}.
 ```

kezz (changes requested):

TheMode:
> If this argument accepts custom sounds, returning a `SoundEvent` is wrong as that only supports built-in sounds. Probably worth either changing this to return `Sound.Type`, which can either be a `SoundEvent` or just a custom sound with a key. This would allow people to `instanceof` to check if they have a `SoundEvent` or not.

Should be fine once we are done replacing the enums

kezz (approved):

PR #346, opened 2021-06-29
Deprecate `UpdateOption`

### LINKED ISSUE #258: Replace `UpdateOption` with `Duration`
The `UpdateOption` class is essentially a worse version of the `Duration` class with fewer features and the inability to use it throughout other APIs.

The `UpdateOption` class should be deprecated and replaced with the `Duration` class. The same goes for the `TimeUnit` class which should be using Java's `ChronoUnit` class and potentially the tick class introduced in #205.

Kebab11noel:
This PR will resolve #258 by deprecating and replacing `net.minestom.server.utils.time.UpdateOption` with `java.time.Duration`.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -1530,9 +1530,7 @@ public class Entity implements Viewable, Tickable, EventHandler<EntityEvent>, Da
      * @param delay    the time before removing the entity,
      *                 0 to cancel the removing
      * @param temporalUnit the unit of the delay
-     * @deprecated Replaced by {@link #scheduleRemove(Duration)}
      */
-    @Deprecated
     public void scheduleRemove(long delay, @NotNull TemporalUnit temporalUnit) {
         scheduleRemove(Duration.of(delay, temporalUnit));
     }
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -45,14 +45,15 @@ import net.minestom.server.utils.chunk.ChunkUtils;
 import net.minestom.server.utils.entity.EntityUtils;
 import net.minestom.server.utils.player.PlayerUtils;
 import net.minestom.server.utils.time.Cooldown;
-import net.minestom.server.utils.time.TimeUnit;
-import net.minestom.server.utils.time.UpdateOption;
 import net.minestom.server.utils.validate.Check;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jglrxavpok.hephaistos.nbt.NBTCompound;
 
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -127,8 +128,8 @@ public class Entity implements Viewable, Tickable, EventHandler<EntityEvent>, Da
     protected EntityType entityType; // UNSAFE to change, modify at your own risk
 
     // Network synchronization, send the absolute position of the entity each X milliseconds
-    private static final UpdateOption SYNCHRONIZATION_COOLDOWN = new UpdateOption(1, TimeUnit.MINUTE);
-    private UpdateOption customSynchronizationCooldown;
+    private static final Duration SYNCHRONIZATION_COOLDOWN = Duration.of(1, ChronoUnit.MINUTES);
+    private Duration customSynchronizationCooldown;
     private long lastAbsoluteSynchronizationTime;
 
     // Events
@@ -1142,7 +1143,7 @@ public class Entity implements Viewable, Tickable, EventHandler<EntityEvent>, Da
      * Sets the entity in fire visually.
      * <p>
      * WARNING: if you want to apply damage or specify a duration,
-     * see {@link LivingEntity#setFireForDuration(int, TimeUnit)}.
+     * see {@link LivingEntity#setFireForDuration(int, TemporalUnit)}.
      *
      * @param fire should the entity be set in fire
      */
@@ -1528,18 +1529,18 @@ public class Entity implements Viewable, Tickable, EventHandler<EntityEvent>, Da
      *
      * @param delay    the time before removing the entity,
      *                 0 to cancel the removing
-     * @param timeUnit the unit of the delay
+     * @param temporalUnit the unit of the delay
      */
-    public void scheduleRemove(long delay, @NotNull TimeUnit timeUnit) {
+    public void scheduleRemove(long delay, @NotNull TemporalUnit temporalUnit) {
         if (delay == 0) { // Cancel the scheduled remove
             this.scheduledRemoveTime = 0;
             return;
         }
-        this.scheduledRemoveTime = System.currentTimeMillis() + timeUnit.toMilliseconds(delay);
+        this.scheduledRemoveTime = System.currentTimeMillis() + temporalUnit.getDuration().multipliedBy(delay).toMillis();
     }
 
     /**
-     * Gets if the entity removal has been scheduled with {@link #scheduleRemove(long, TimeUnit)}.
+     * Gets if the entity removal has been scheduled with {@link #scheduleRemove(long, TemporalUnit)}.
      *
      * @return true if the entity removal has been scheduled
      */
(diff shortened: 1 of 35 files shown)


src/main/java/net/minestom/server/utils/time/TimeUnit.java
```diff
@@ -2,6 +2,10 @@
 
 import net.minestom.server.MinecraftServer;
 
+/**
+ * @deprecated Replaced by {@link java.time.temporal.TemporalUnit}
+ */
+@Deprecated(forRemoval = true)
 public enum TimeUnit {
```
TheMode, line 9: I believe that we should keep this as a class to expose all our time units. Otherwise we will end up using `ChronoUnit` and `Tick` separately.

TheMode (changes requested):
`forRemoval` should not be necessary for now

src/main/java/net/minestom/server/utils/time/TimeUnit.java
kezz: This enum could be made to implement `TemporalUnit`, delegating the methods to both `Tick` and `ChronoUnit` respectively. However, I'm not a massive fan of this option as a long-term solution.

src/main/java/net/minestom/server/utils/time/Tick.java
```diff
@@ -0,0 +1,45 @@
+package net.minestom.server.utils.time;
+
+import net.minestom.server.MinecraftServer;
+
+import java.time.Duration;
+import java.time.temporal.Temporal;
+import java.time.temporal.TemporalUnit;
+
+public class Tick implements TemporalUnit {
```
kezz, line 9: I prefer my implementation of this class (in #205). Feel free to nab it to replace this if you want.

kezz (approved):

src/main/java/net/minestom/server/utils/time/TimeUnit.java
Kebab11noel: I think duplicating code (in this case keeping the enum instead of `ChronoUnit`) isn't a good thing, personally i think nothing wrong with using the two beside each other.

TheMode: The issue is that most people will not know that the two exist. I think that it makes sense to list everything somewhere as it would be officialy supported

src/main/java/net/minestom/server/utils/time/Tick.java
Kebab11noel: I just checked that pr after i pushed this, but i'll include it too or if it gets merged before this i'll rebase.

Kebab11noel: Well I can just copy-paste it and i'll have less things to do if i have to rebase... Soon i'll push it and resolve this

src/main/java/net/minestom/server/utils/time/TimeUnit.java
Kebab11noel: Yeah i see... Then can we settle with a compromise so that the `TimeUnit` isn't going to be for removal (but still deprecated) and in the deprecation notice include `Tick` and `ChronoUnit`?

Kebab11noel: Maybe add a wiki page too, long before removal, that way it really shouldn't cause issues.

TheMode: I don't really see why adding constants in TimeUnit is an issue, it is in the utils package and could be used to list all the officially supported units

Kebab11noel: Oh wait... I thought about a different implementation of this... nevermind; so you mean like this?
```java
public enum TimeUnit {
    DAY(ChronoUnit.DAYS);
    //...

    private final TemporalUnit temporalUnit;

    TimeUnit(TemporalUnit temporalUnit) {
        this.temporalUnit = temporalUnit;
    }

    public TemporalUnit getTemporalUnit() {
        return temporalUnit;
    }
}
```

TheMode: I said class constants, so
```java
public static final TemporalUnit DAY = ChronoUnit.DAYS;
public static final TemporalUnit TICK = Tick.TICKS;
```

Kebab11noel: Ah okay, i don't know why i wanted to stick with the enum... I'm fine with constants

Kebab11noel: Oh i know why i wanted to stick with the enum, because of deprecation so it won't break existing projects, but if we going to use class instead of enum there is no point of keeping deprecated methods i marked for removal previously.

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1528,18 +1529,18 @@ public boolean isRemoved() {
      *
      * @param delay    the time before removing the entity,
      *                 0 to cancel the removing
-     * @param timeUnit the unit of the delay
+     * @param temporalUnit the unit of the delay
      */
-    public void scheduleRemove(long delay, @NotNull TimeUnit timeUnit) {
+    public void scheduleRemove(long delay, @NotNull TemporalUnit temporalUnit) {
```
kezz, line 1534: I think if we're moving to use more `java.time` stuff, methods like this where we pass a number and a time unit should just accept `Duratiom` instead.

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1609,8 +1610,20 @@ public void askSynchronization() {
      * Set custom cooldown for position synchronization.
      *
      * @param cooldown custom cooldown for position synchronization.
+     * @deprecated Replaced by {@link #setCustomSynchronizationCooldown(Duration)}
      */
-    public void setCustomSynchronizationCooldown(@Nullable UpdateOption cooldown) {
+    @SuppressWarnings("removal")
+    @Deprecated(forRemoval = true)
+    public void setCustomSynchronizationCooldown(@Nullable net.minestom.server.utils.time.UpdateOption cooldown) {
```
kezz, line 1617: There's so many breaking changes in this PR atm. I think we should either break them all or try and not break anything - no need for half measures.

src/main/java/net/minestom/server/entity/ItemEntity.java
```diff
@@ -225,10 +240,10 @@ public long getPickupDelay() {
      * Sets the pickup delay of the ItemEntity.
      *
      * @param delay    the pickup delay
-     * @param timeUnit the unit of the delay
+     * @param temporalUnit the unit of the delay
      */
-    public void setPickupDelay(long delay, @NotNull TimeUnit timeUnit) {
-        this.pickupDelay = timeUnit.toMilliseconds(delay);
+    public void setPickupDelay(long delay, @NotNull TemporalUnit temporalUnit) {
```
kezz, line 245: Same here, can just take `Duration`

src/main/java/net/minestom/server/instance/block/CustomBlock.java
```diff
@@ -172,14 +172,14 @@ public boolean enableMultiPlayerBreaking() {
...
      */
     public boolean hasUpdate() {
-        final UpdateOption updateOption = getUpdateOption();
-        if (updateOption == null)
+        final Duration updateFrequency = getUpdateFrequency();
+        if (updateFrequency == null)
             return false;
 
-        return updateOption.getValue() > 0;
+        return !updateFrequency.isZero();
```
kezz, line 182: This isn't quite right - you want to check for a negation of `isNegative`.

src/main/java/net/minestom/server/item/metadata/PotionMeta.java
```diff
@@ -98,7 +98,7 @@ public void read(@NotNull NBTCompound nbtCompound) {
                 for (NBTCompound potionCompound : customEffectList) {
                     final byte id = potionCompound.getAsByte("Id");
                     final byte amplifier = potionCompound.getAsByte("Amplifier");
-                    final int duration = potionCompound.containsKey("Duration") ? potionCompound.getNumber("Duration").intValue() : (int) TimeUnit.SECOND.toMilliseconds(30);
+                    final int duration = potionCompound.containsKey("Duration") ? potionCompound.getNumber("Duration").intValue() : (int) TimeUnit.getMillis(30, TimeUnit.SECOND);
```
kezz, line 101: We should probably be using `ChronoUnit` instead of the concurrent-based `TimeUnit`

src/main/java/net/minestom/server/entity/Entity.java
Kebab11noel: Correct me if i'm wrong, but this pr won't break a single project (exept of those which rely on `TimeUnit.toMilliseconds`).

src/main/java/net/minestom/server/entity/Entity.java
Kebab11noel: Fair point, i'll deprecate these and add methods that take `Duration`

src/main/java/net/minestom/server/item/metadata/PotionMeta.java
Kebab11noel: Well I used it because it's much shorter than writing `ChronoUnit.SECONDS.getDuration().multipliedBy(30).toMillis()` also i didn't want to mix the two like `TimeUnit.getMillis(30, ChronoUnit.SECONDS)`.

PR #347, opened 2021-06-30
Fix automatic component translation 

GreenpixDev:
I removed the "final" modifier from the field "AUTOMATIC_COMPONENT_TRANSLATION" to enable component translation.

### DIFF
--- a/src/main/java/net/minestom/server/adventure/MinestomAdventure.java
+++ b/src/main/java/net/minestom/server/adventure/MinestomAdventure.java
@@ -26,7 +26,7 @@ public final class MinestomAdventure {
     /**
      * If components should be automatically translated in outgoing packets.
      */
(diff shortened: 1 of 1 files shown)


kezz:
Not 100% sure about this one. The reason we kept it final was so a lot of stuff get cleverly compiled away. I really doubt the performance hit from making this non-final is meaningful, however.

NukeKM:
It really doesn't make sense though. Why add automatic component translation in the first place? As far as I know it can't be manually set to true, so even if I'd need to use it, I won't be able to. In my opinion, making it non-final is the best option here. (Sorry for my english)

ifxory-plugins:
Hey, why do you keep this field in class, if in this case, you could be remove it? Meaning from it, if it cannot be changed?

Faintof:
Люди с снс собрались чтобы дружно апнуть квешен, уже поздно, но + up

Faintof:
хуй

PR #349, opened 2021-07-04
copy changes to Vector from #348

MrGazdag:
see #348 for more details


MrGazdag:
since changes from Vector are not needed (i am dumb) i'm gonna create another PR without the changes to Vector

PR #351, opened 2021-07-04
Click callback

### LINKED ISSUE #321: Click event callback commands
Similar to Sponge's command callbacks, it would be nice to have some sort of utility method to temporarily create a configurable command with a simple callback that would be stored and expired over time.

Kebab11noel:
This pull request will add a utility class, that handles click callbacks, also closes #321.
**Features**
- [ ] expire callback after specified time (or never)
- [x] usage limit (or unlimited)
- [x] check executor identity (optional, security feature*)
- [ ] customizable error messages (default is none)
- [ ] multiple method overrides - for easier usage
- [ ] callback for executor identity violation (optional, can be used to flag malicious users)

**Example usage**
```java
// This will create a never-expiring, insecure* callback that can be used indefinitely
player.sendMessage(Component.text("ping").clickEvent(
    TextAction.runCallback(() -> player.sendMessage("pong"), null, null, null)
));
```
<hr>
* Behind the curtains the client executes a command on click and that command contains a UUID to identify the callback, however the command can also be executed in the regular way, which means it's possible to brute-force the ids, although the likelyhood of success is low for sensitive operation the check should be used to ensure only one person can execute the callback.

### DIFF
--- a/src/main/java/net/minestom/server/utils/TextAction.java
+++ b/src/main/java/net/minestom/server/utils/TextAction.java
@@ -1,0 +1,106 @@
+package net.minestom.server.utils;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import net.kyori.adventure.text.event.ClickEvent;
+import net.minestom.server.MinecraftServer;
+import net.minestom.server.command.CommandSender;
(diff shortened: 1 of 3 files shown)


KrystilizeNevaDies:
Closed due to inactivity.

PR #357, opened 2021-07-06
Increase rate limit to support requestful mods.

LeoDog896:
This mainly makes it where mods such as AntiGhost won't immideately kick you from the server.


Kebab11noel:
Isn't [Acknowledge Player Digging](https://wiki.vg/Protocol#Acknowledge_Player_Digging) packet's successful field there to prevent ghosting?

PR #358, opened 2021-07-07
Command API improvement

Kebab11noel:
# New features
* automatically unregister commands after
   * specific time - `#setTtl(Duration)`
   * usage limit is reached - `#setMaxUsage(Integer)`
* white- and blacklist command senders to limit access
* hide commands - `#setHidden(boolean)`

# Possible features
The following features are just proposals, they might not make it into the final PR.
## Command builder (Status: proposed, implemented)
A command builder can be used to inline command creation, this is useful for simple commands. The same command implemented in three different ways:
**1. Using the command builder**
```java
final String timeCommand = MinecraftServer.getCommandManager().register(
        Command.builder((command, sender, context) -> sender.sendMessage(timeUntilNextHour())).build());

player.sendMessage(Component.text("Time until the next hour")
        .clickEvent(ClickEvent.runCommand("/"+timeCommand)));
```
2. Using the `Command` class
```java
final Command timeCommand = new Command(UUID.randomUUID().toString());
timeCommand.setDefaultExecutor(((sender, context) -> sender.sendMessage(timeUntilNextHour())));
timeCommand.setHidden(true);
MinecraftServer.getCommandManager().register(timeCommand);

player.sendMessage(Component.text("Time until the next hour")
        .clickEvent(ClickEvent.runCommand("/"+timeCommand.getName())));
```
3. Extending `Command` class
```java
// TimeCommand.java
public class TimeCommand extends Command {
    public TimeCommand() {
        super(UUID.randomUUID().toString());
        
        setDefaultExecutor(this::execute);
        setHidden(true);
    }

    private void execute(@NotNull CommandSender commandSender, @NotNull CommandContext context) {
        commandSender.sendMessage(timeUntilNextHour());
    }
}

// other file
final String timeCommand = MinecraftServer.getCommandManager().register(new TimeCommand());
player.sendMessage(Component.text("Time until the next hour")
        .clickEvent(ClickEvent.runCommand("/"+timeCommand)));
```
## Multiple condition with priority (Status: proposed)
Allowing multiple `CommandCondition`s can be useful to break down a big condition check into smaller chunks that could be reused among other commands. This would also mean that white- and blacklist, expiration and usage checks become automatically handled conditions.<br>
_No code examples yet_

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandManager.java
+++ b/src/main/java/net/minestom/server/command/CommandManager.java
@@ -3,6 +3,7 @@ package net.minestom.server.command;
 import it.unimi.dsi.fastutil.Pair;
 import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.ints.IntList;
+import net.minestom.server.MinecraftServer;
 import net.minestom.server.command.builder.*;
 import net.minestom.server.command.builder.arguments.Argument;
 import net.minestom.server.command.builder.arguments.minecraft.SuggestionType;
@@ -47,8 +48,9 @@ public final class CommandManager {
      *
      * @param command the command to register
      * @throws IllegalStateException if a command with the same name already exists
+     * @return the registered command's name
      */
-    public synchronized void register(@NotNull Command command) {
+    public synchronized String register(@NotNull Command command) {
         Check.stateCondition(commandExists(command.getName()),
                 "A command with the name " + command.getName() + " is already registered!");
         if (command.getAliases() != null) {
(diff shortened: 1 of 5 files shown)


mworzala:
What is the difference between white/blacklisting and permissions?

Kebab11noel:
White/blacklist is more bound to the command it self than to the player. For example in the case of a temporary command it makes more sense to say "this cammand can be executed by this player" than "this player can execute this specific command", this should also make cleanup easier, on the other hand permissions make more sense for normal commands like `/give`.

PR #360, opened 2021-07-08
Fix extension resource load in Windows

djytw:
In windows, `Path#toString` will use `\` as the separator, but `ClassLoader#getResource` needs `/`. This will cause resources in subfolders to fail to load in Windows.

### DIFF
--- a/src/main/java/net/minestom/server/extensions/Extension.java
+++ b/src/main/java/net/minestom/server/extensions/Extension.java
@@ -151,7 +151,18 @@ public abstract class Extension {
      * @return The file contents, or null if there was an issue reading the file.
      */
     public @Nullable InputStream getPackagedResource(@NotNull String fileName) {
-        return getPackagedResource(Paths.get(fileName));
+        try {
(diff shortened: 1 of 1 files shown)


kezz:
Do you think it would make more sense for this method to just accept a string instead?

mworzala:
> Do you think it would make more sense for this method to just accept a string instead?

There is a string variant as well ([here](https://github.com/Minestom/Minestom/blob/694d05e78aa9ed34561e3836eb5c14c2f33d3c09/src/main/java/net/minestom/server/extensions/Extension.java#L153)).

kezz:
Maybe we bring it the other way around, make the string the default and the path one the overload - going from string to path and back to string feels a bit unnecessary.

djytw:
> Maybe we bring it the other way around, make the string the default and the path one the overload - going from string to path and back to string feels a bit unnecessary.

Good idea but more lines need to be changed, including `savePackagedResource` and `getResource` - they all take `Path` as the default parameter. If you guys agreed on this, I can make the change.

kezz (approved):

PR #361, opened 2021-07-08
Add ComponentHolding to packets for fully support automatic translation

GreenpixDev:
The "automatic component translation" option only works for packets that inherit **ComponentHoldingServerPacket** interface, so I added this interface to all packets that contain components and items. Also I added inheritance **ComponentHolder** for **ItemStack**.

### DIFF
--- a/src/main/java/net/minestom/server/network/packet/server/play/ResourcePackSendPacket.java
+++ b/src/main/java/net/minestom/server/network/packet/server/play/ResourcePackSendPacket.java
@@ -66,11 +66,11 @@ public class ResourcePackSendPacket implements ComponentHoldingServerPacket {
 
     @Override
     public @NotNull Collection<Component> components() {
-        List<Component> components = new ArrayList<>();
         if (forcedMessage != null) {
-            components.add(forcedMessage);
+            return Collections.singleton(forcedMessage);
+        } else {
+            return Collections.emptyList();
         }
-        return components;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


src/main/java/net/minestom/server/network/packet/server/play/ResourcePackSendPacket.java
```diff
@@ -56,4 +63,23 @@ public void read(@NotNull BinaryReader reader) {
...
         return ServerPacketIdentifier.RESOURCE_PACK_SEND;
     }
+
+    @Override
+    public @NotNull Collection<Component> components() {
+        List<Component> components = new ArrayList<>();
+        if (forcedMessage != null) {
+            components.add(forcedMessage);
+        }
+        return components;
```
kezz, line 73: Saves an allocation I suppose!
```suggestion
        if (forcedMessage != null) {
            return Collections.singleton(forcedMessage);
        } else {
            return Collections.emptyList();
        }
```

kezz (changes requested):
Could we do away with `ComponentHoldingServerPacket` entirely and just use `ComponentHolder`?

TheMode:
A bit off-topic but I am personally still against the whole adventure translation API. They prevent a lot of optimization we are trying to push using immutable data.
Even if it gets merged, we will not be able to enforce it. (an example would be caching window items packets)

GreenpixDev:
Indeed they prevent a lot of optimization but some users also need server flexibility. Everyone can enable or disable automatic translation or caching of packets.
If the automatic translation is not planned to be supported in the future, I suggest making it deprecated.

TheMode:
> Indeed they prevent a lot of optimization but some users also need server flexibility. Everyone can enable or disable automatic translation or caching of packets.
> If the automatic translation is not planned to be supported in the future, I suggest making it deprecated.

The issue is that it drastically complicates the implementation of those optimizations.

We are not sure what direction to take with it, my take would be to handle all those packet transformations on a proxy. https://github.com/TheMode/Paxy

GreenpixDev:
Ok i got it, thanks!
I close this pull request.

PR #368, opened 2021-07-12
Implement EntityBlockCollisionEvent

KrystilizeNevaDies:
This PR implements a method to detect and cancel collisions, this allows for logic such as proper liquid handling and collision events such as fireball explosions.


KrystilizeNevaDies:
Closing to push to new block api

PR #371, opened 2021-07-17
Add getTags to ItemMeta

Moulberry:


LeoDog896:
Maybe this would be better on the Tag Holder itself?

TheMode:
I think that it would make more sense to have such a method inside `TagReadable`, though we need to be careful about what we add there. As some tags do not necessarily map to a nbt key

PR #376, opened 2021-07-23
Fix deserialization issues w/ ItemMeta had Name/Lore containing lit…

Moulberry:
Fix deserialization issues when ItemMeta had Name/Lore containing literal quote (") characters

It seems as though the call to unescapeJavaString is unnecessary, and in fact can cause issues with literal quote characters.


Moulberry:
Issue doesn't appear to be fixable in Minestom.
https://github.com/jglrxavpok/Hephaistos/issues/8

PR #382, opened 2021-07-27
Add more name cases to ArgumentEnum

MrGazdag:
Adds `camelCase`, `PascalCase`, `Title Case`, and `snake_case` to the supported formats on an [ArgumentEnum](https://github.com/Minestom/Minestom/blob/new-block-api/src/main/java/net/minestom/server/command/builder/arguments/ArgumentEnum.java), including examples.

### DIFF
--- a/src/main/java/net/minestom/server/command/builder/arguments/ArgumentEnum.java
+++ b/src/main/java/net/minestom/server/command/builder/arguments/ArgumentEnum.java
@@ -57,9 +57,153 @@ public class ArgumentEnum<E extends Enum> extends Argument<E> {
     }
 
     public enum Format {
+        /**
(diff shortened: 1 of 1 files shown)


TheMode:
I am personally not convinced, I feel that using those will always result in a poorly designed command. Messing with upper/lower case will just make the client confused

LeoDog896:
I personally would love camelCase or snake_case, just lowercase can get confusing really quickly.

MTGPROD:
I personally prefer camelCase, snake_case is ulgy

MrGazdag:
> Literals are case-sensitives, wouldn't it just make the syntax hard to use?

Tab-completion helps with that

PR #383, opened 2021-07-27
Set time to negative when timeRate is 0

MrGazdag:
Modifies the `createTimePacket()` to set the time to negative is the `timeRate` is 0, to disable the sun moving

Reference: [wiki.vg](https://wiki.vg/Protocol#Time_Update)

Current code produces this (recorded for 2 seconds, notice how the clouds do not reset):
sun bug

### DIFF
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -405,7 +405,7 @@ public abstract class Instance implements BlockGetter, BlockSetter, Tickable, Ta
      * @return the {@link TimeUpdatePacket} with this instance data
      */
     private @NotNull TimeUpdatePacket createTimePacket() {
-        return new TimeUpdatePacket(worldAge, timeRate == 0 ? -time : time);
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -405,7 +405,7 @@ public void setTimeUpdate(@Nullable Duration timeUpdate) {
      * @return the {@link TimeUpdatePacket} with this instance data
      */
     private @NotNull TimeUpdatePacket createTimePacket() {
-        return new TimeUpdatePacket(worldAge, time);
+        return new TimeUpdatePacket(worldAge, timeRate == 0 ? -time : time);
```
Eoghanmc22, line 408: Maybe do `-Math.abs(time)` because projects probably did this previously with negative time

MrGazdag: > Maybe do `-Math.abs(time)` because projects probably did this previously with negative time

Alright, modifying

PR #387, opened 2021-07-31
Changed how the position turned into a short

sunarya-thito:
Using modulo is not the proper way to get the relative position in a chunk because it creates issues for negative coordinates.
This change will fix the issue with negative coordinates.


TheMode:
Sorry for the late answer, I would perhaps suggest `ChunkUtils#getChunkCoordinate` instead

PR #389, opened 2021-08-03
Switch to NIO

TheMode:
Experiment using NIO directly instead of Netty. Goal being to have more control over the pipeline and memory allocation (reusing buffers while avoiding pooling)

* Future support for unix-domain socket https://openjdk.java.net/jeps/380
* Future buffers using the new off-heap memory API https://openjdk.java.net/jeps/412
* Simplify the integration of a packet transformation API https://github.com/TheMode/Paxy


MrGazdag:
Don't forget to rename packages and remove/rename stuff which mentions netty (like NettyPlayerConnection)

PR #391, opened 2021-08-06
Allow player to use creative mode features in creative mode

AwesomestCode:
Disables client side prediction of water buckets being emptied when placed in creative mode. wiki.vg labels this flag incorrectly as instant-break: https://wiki.vg/Protocol#Player_Abilities_.28clientbound.29

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -1331,8 +1331,14 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
             sendChangeGameStatePacket(ChangeGameStatePacket.Reason.CHANGE_GAMEMODE, gameMode.getId());
 
             PlayerInfoPacket infoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.UPDATE_GAMEMODE);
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1331,8 +1331,14 @@ public void setGameMode(@NotNull GameMode gameMode) {
             sendChangeGameStatePacket(ChangeGameStatePacket.Reason.CHANGE_GAMEMODE, gameMode.getId());
 
             PlayerInfoPacket infoPacket = new PlayerInfoPacket(PlayerInfoPacket.Action.UPDATE_GAMEMODE);
+            PlayerAbilitiesPacket abilitiesPacket = new PlayerAbilitiesPacket();
+            if (gameMode.equals(GameMode.CREATIVE)) {
+                abilitiesPacket.flags = 0b1101;
```
mworzala, line 1336: I haven't looked, but I assume there are constants somewhere for this which could be used. If not they should be added. At minimum a comment here is needed to say which flags are being set/what they do.

MrGazdag: i made those constants, its in PlayerAbilitiesPacket

PR #393, opened 2021-08-06
Add Tab Completion to Console w/ JLine

LeoDog896:
This adds basic tab completion to minestom.


MrGazdag:
The name of this commit is a bit confusing, it implies tab completing not being in Minestom, which is not the case

You should rename it to something like, "Add Tab Completion to Console/JLine"

Other than that, it would be cool to tab-complete from console. +1

KrystilizeNevaDies:
Closed due to inactivity.

PR #394, opened 2021-08-07
Fix a bug with the map api

Eoghanmc22:
When creating maps directly from a LargeFramebuffer without the use of sub-views, preparePacket didn't work correctly.

Current behavior:
The entire map is the color of the first pixel in the chosen region

Expected behavior:
Each pixel has their assigned color

Solution:
Use the x and y variables from the loop instead of the left and top arguments which were passed in to the function.


jglrxavpok (approved):
Looks good to me

PR #397, opened 2021-08-09
Fix player stop sprinting when changing attributes

Bloepiloepi:
This pull request fixes an issue I encountered: the player stops sprinting when you change attributes.
The fix is to send only the required attributes in the packet, instead of all.

This pull request also contains a fix for `CrossbowMeta`, because the fix is very small and it didn't make sense to put it in a separate pull request.


Marlonzockt:
_because the fix is very small and it didn't make sense to put it in a separate pull request_ Looks at #395

PR #406, opened 2021-08-11
Switch annotations from `@ApiStatus.Internal` to `@ApiStatus.Experimental` for utility classes.

BitCrack:
`@ApiStatus.Experimental` implies that the class is usable, but is expected to change often and break code.
This annotation in my opinion fits utility classes better than the current `@ApiStatus.Internal` which implies that the class is used only for internal code and shouldn't be used as a public API, but in my opinion, the utility classes affected are okay to be included as apart of the API.


TheMode:
I am personally not convinced. As those methods are unlikely to ever be considered "stable"

PR #414, opened 2021-08-15
Don't enforce automatic scheduling of `Task`  within `TaskBuilder`

BitCrack:
This PR allows you to build a `Task` from `TaskBuilder` if you need to schedule it at a later time.
You could do the same by constructing the `Task` by yourself, but there were no methods inside the builder to do that without it scheduling it (as the only method available for that is `TaskBuilder#schedule`).

This adds a `TaskBuilder#build` method and moves practically all functionality from `TaskBuilder#schedule` to `Task#schedule` and a new method, `Task#scheduleForShutdown`.

### DIFF
--- a/src/main/java/net/minestom/server/timer/Task.java
+++ b/src/main/java/net/minestom/server/timer/Task.java
@@ -100,12 +100,26 @@ public class Task implements Runnable {
      * Sets up the task for correct execution.
      */
     public void schedule() {
+        Int2ObjectMap<Task> tasks = this.schedulerManager.tasks;
+        synchronized (tasks) {
+            tasks.put(getId(), this);
+        }
         if(owningExtension != null) {
             this.schedulerManager.onScheduleFromExtension(owningExtension, this);
         }
         this.future = this.repeat == 0L ?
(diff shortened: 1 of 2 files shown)


mworzala (commented):
No issue with the idea, however there is one inconsistency.

The task does not keep track of whether its a shutdown task or not so the following snippets behave differently.
```java
scheduleManager.buildShutdownTask().schedule();
scheduleManager.buildShutdownTask().build().schedule();
```

While there is of course a separate `#scheduleForShutdown` this seems rather confusing to me.

BitCrack:
> No issue with the idea, however there is one inconsistency.
> 
> The task does not keep track of whether its a shutdown task or not so the following snippets behave differently.
> 
> ```java
> scheduleManager.buildShutdownTask().schedule();
> scheduleManager.buildShutdownTask().build().schedule();
> ```
> 
> While there is of course a separate `#scheduleForShutdown` this seems rather confusing to me.

I have merged `Task#scheduleForShutdown` into the existing method and used an already present variable `shutdown` to determine how to schedule that task

mworzala (approved):

PR #416, opened 2021-08-18
Air drag fix

Bloepiloepi:
This pull request fixes the air drag of some entities being off.

Living entities have an air drag of 0.91, and other entities have 0.98. Minestom does not provide any rules for entities, so maybe it should be customizable? (Like `Entity.setAirDrag(0.98)`)


Kebab11noel (approved):
LGTM! However I think too that the airDrag should be a field to allow for customization.

PR #417, opened 2021-08-19
Free event calling

TheMode:
Introduce `ListenerHandle<Event>` representing a direct access to an `EventNode` listener list. Useful when an event is often called and map lookup should be avoided.

Listeners propagation also changed, each `EventNode` will cache a list of consumers for each event type. The list will be updated whenever a new listener is added to a child or itself.

This is far from being remotely usable, but this is much more efficient than the current approach (I even noticed that no event allocation is performed at all when the event does not have any listener on Hotspot JDK 16)


mworzala (approved):
Happy with public API, did not review implementation

PR #419, opened 2021-08-19
Rewrote tags API to allow for better access and checking

BomBardyGamer:
This should probably be 2 PRs, but I can sort that out afterwards.

This PR is a rewrite of the old tags API, to be much more open to checking tags against their native types, to allow adding custom tag types for modified clients, and also, to support the constants (there are 5 classes with the built-in tags as constants in them).

Things to address:
- [ ] The new game event object I had to add to make this work
- [x] The choice to make `Fluid` implement `ProtocolObject` and changing the generators to match
- [ ] Whether the 5 classes with the constants for the built-in tags should be auto generated or not


mworzala:
Before a "formal" review, some notes:
 - Making `Fluid` a `ProtocolObject` seems reasonable.
 - The constant classes should be generated if its reasonable, they will always need to match the game when it updates anyway.

Someone else will also need to take a look here, I am not terribly familiar with the current system (or even the game data), so I am mostly looking at structure.

BomBardyGamer:
Rebased this to target the 1.18 branch, going to re-open this.

PR #420, opened 2021-08-20
Fast exit collision handling

Kebab11noel:


src/main/java/net/minestom/server/collision/CollisionCache.java
```diff
@@ -0,0 +1,57 @@
...
+            return new Pair<>(false, null);
+        }
+        if (wasOnGround && velocity.y() <= 0) {
+            // Ignore negative/zero y
+            velocity = velocity.withY(0);
+        }
+
+        final Pos position = entity.getPosition();
+        final int cornerHash = hashCorners(position.x(), position.y(), position.z(), entity.getBoundingBox());
+        return new Pair<>(this.cornerHash == cornerHash, velocity);
```
kezz, line 43: Would be better to use something specific here (possibly that can be turned into a record later.

KrystilizeNevaDies:
Closed due to inactivity. Feel free to reopen.

PR #421, opened 2021-08-20
Accurate player velocity

Bloepiloepi:
This pull request makes the server calculate player velocity using their previous position.
Which means:
- Always the right value in `Entity.getVelocity()`
- A little less computation, because collision and physics for players is skipped
- Methods like `Entity.setVelocity()` still work fine


Kebab11noel (approved):
This also looks good to me, tested with multiple players and didn't find anything.

PR #428, opened 2021-08-22
Entity#getLineOfSightEntity and BoundingBox#intersect(Point, Point)

RinesThaix:
Relocated line-of-sight methods from `LivingEntity` to `Entity`;
Implemented `BoundingBox#intersect(Point, Point)` function to check whether bounding box intersects given segment of a line;
Implemented `Entity#getNearbyEntities` (probably not as a final solution, but as a dependency for the next method) and `Entity#getLineOfSightEntity`

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -35,6 +35,7 @@ import net.minestom.server.tag.Tag;
 import net.minestom.server.tag.TagHandler;
 import net.minestom.server.thread.ThreadProvider;
 import net.minestom.server.utils.async.AsyncUtils;
+import net.minestom.server.utils.block.BlockIterator;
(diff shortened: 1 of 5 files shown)


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1452,6 +1455,122 @@ public void takeKnockback(final float strength, final double x, final double z)
         }
     }
 
+    public Collection<Entity> getNearbyEntities(double range) {
```
TheMode, line 1458: Should probably be in `Instance`

TheMode (changes requested):
LGTM, just need to move the `getNearbyEntities` method

src/main/java/net/minestom/server/entity/Entity.java
RinesThaix: Did it and also added synchronization over entities lock for the whole cycle

PR #430, opened 2021-08-24
Close AnvilLoader Region files when unregistering instance - Fixes #423

### LINKED ISSUE #423: AnvilLoader does not release world files after unregistering instance
After saving instance files (world regions, level...), the AnvilLoader does not close some files

Code:
```java
instance.saveInstance()
    .thenCompose(unused -> instance.saveChunksToStorage())
    .thenRun(() -> manager.unregisterInstance(instance));
```

Locked folder:
image

The folder is unlocked after closing/killing the server

AlexDicy:
Allows an `IChunkLoader` to clear previously loaded data from an instance.

Closes and clears regions loaded previously in AnvilLoader when `MinecraftServer.getInstanceManager().unregisterInstance(InstanceContainer)` is called.

With the current AnvilLoader implementation, once an `InstanceContainer` is unregistered it can be registered again.


src/main/java/net/minestom/server/instance/SharedInstance.java
```diff
@@ -104,6 +104,11 @@ public boolean isInVoid(@NotNull Point point) {
         return instanceContainer.isInVoid(point);
     }
 
+    @Override
+    protected void setRegistered(boolean registered) {
+        instanceContainer.setRegistered(registered);
+    }
+
```
jglrxavpok, line 111: @TheMode is that correct? The SharedInstance could be unregistered but the InstanceContainer can still live right?
To me, it should keep its chunk loader alive.

jglrxavpok (changes requested):

src/main/java/net/minestom/server/instance/SharedInstance.java
AlexDicy: I agree, it should be kept alive. It's a misunderstanding from me

AlexDicy:
Renaming the branch just 'caused a load of issues. Created a new PR with the correct branch to start with

PR #443, opened 2021-08-31
Add cursor position to BlockPlacementRule

MrGazdag:
This PR adds the cursor position parameter to BlockPlacementRule's `blockPlace` method.

BlockPlacementRule implementations currently cannot be used to create functioning slab/stair placement, as they depend on the Player's cursor (if the cursor is above 0.5 Y, then a top slab/stair block will be placed). This would require the exact cursor position.
A possible alternative is to raycast server-side, but because the player's yaw can slightly differ server-side, than client-side, it might give incorrect results.

In order to not break existing BlockPlacementRule implementations, the existing method was left in, it was marked as deprecated, and a new one with the new parameter was added. Both methods have received a default implementation, again, to not break existing BlockPlacementRule implementations (and to not require implementing the old method on newer implementations).
The old method can be:
 - left as is,
 - marked for removal, then removed later, or
 - removed completely.
 
 This PR does not yet include new implementations for the existing built-in BlockPlacementRules.

### DIFF
--- a/src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
+++ b/src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
@@ -34,10 +34,32 @@ public abstract class BlockPlacementRule {
      * @param blockPosition the block position
      * @param pl            the player who placed the block
      * @return the block to place, {@code null} to cancel
+     * 
+     * @deprecated use {@link #blockPlace(Instance, Block, BlockFace, Point, Player, Point)} instead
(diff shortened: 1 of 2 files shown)


luzzardik:
Bump! This PR helps with a lot of placement rules issues like stairs, would be lovely to merge and fix once for all the stairs, slabs and other cursor facing related placements !

GreatWyrm:
This was added in 128c525b7cee3ae998c3b55edec73a6cf88878ee.

PR #445, opened 2021-09-04
Add MessageConsumer to ConsoleSender

LeoDog896:
This allows developers to hook into ConsoleSender without permissions, for example:

> In-game console sending
> Rendering components in console
> Redirecting to a live console


kezz:
Needs a default value and a not-null check on set to ensure this doesn't throw a bazillion NPEs.

PR #450, opened 2021-09-07
Chunk-related improvements

RinesThaix:
1. `InstanceChunkLoadEvent` and `InstanceChunkUnloadEvent` now have exact references to a chunk (previous there were only coordinates), because listener would like to retrieve `Chunk` instance itself to do something about it pretty much always.
2. There are `Instance#getChunks` and `Instance#getChunksAt` methods which decrease synchronizations count in `InstanceContainer` drastically whenever you need to get (or check whether they're loaded) multiple chunks at once.


RinesThaix:
I've just realized the code does wrong things, working on it to fix

Issue #415, opened 2021-08-17
When spawning multiple FakePlayers they will all listen to to first spawn event


gabber235:
# Description
When I am trying to spawn multiple `FakePlayer` when a real player joins, all of the `spawnCallback`'s are called since they don't compare for the correct `uuid`. 

## Example
> 1. `Player` tries to join the server
> 2. Code spawns a `FakePlayer`
> 3. `Player` is set to a world
> 4. The player's on spawn event is ran
> 5. `FakePlayer` `spawnCallback` is ran while it has not yet spawned.

Results in a fake player that is not yet active while still having called its callback.

## Current Solution
Looking at the current solution it will not work by just adding check. Since the event is only listened to one.
https://github.com/Minestom/Minestom/blob/adf9a91d7a9e8ead9f4753bf6bab6c89f7972a54/src/main/java/net/minestom/server/entity/fakeplayer/FakePlayer.java#L58-L66

PR #451, opened 2021-09-08
Safe FakePlayer SpawnCallback

gabber235:
# Description
When I am trying to spawn multiple `FakePlayer` when a real player joins, all of the `spawnCallback`'s are called since they don't compare for the correct `uuid`.
For more information about this bug look at #415 

# Solution
To both compare the `uuid` in the spawn event until the correct player is found, and unregister the event listener when it has been found. I used a `EventNode` which is registered when the `FakePlayer` is created. Once the `FakePlayer` has been spawned. The `EventNode` will be unregistered from the global event manager and the callback will be called.


KrystilizeNevaDies:
LGTM, just fix the conflicts and double check that this fixes issue #415 (extra points for a unit test for that)

gabber235:
Looking at the current version of the `FakePlayer` it looks like this is already implemented. This line fixes the issue that I had originally: https://github.com/Minestom/Minestom/blob/0b77fb805b5034c4fdf6a7c81fd635daa2de825d/src/main/java/net/minestom/server/entity/fakeplayer/FakePlayer.java#L62

PR #460, opened 2021-09-12
Make loadDynamicExtension and loadExtensionList provide their Extensions instead of boolean

LeoDog896:
This allows the function's return value to be used to add a string to any command (check LoadExtensionCommand)


mworzala:
Is there any reason to return the `Extension` instead of the `DiscoveredExtension`? People shouldn't really be interacting with the `Extension` instance on their own.

LeoDog896:
No particular reason, unfortunately. I can migrate it once I'm home.

PR #465, opened 2021-09-15
Additional ways to interact with extensions

new-maxim724:
As I wrote in the discord, this will help to decouple the implementation and only give access to the abstraction. It also saves us from casting the type to the required one.

Why am I not using Class<? extends Extension> ? Since we cannot inherit the interface from Extension, we have to do this.
On the one hand, this is not entirely secure, but on the other hand, we know that the ExtensionManager only loads Extension subclasses.

As example:
image

### DIFF
--- a/src/main/java/net/minestom/server/extensions/ExtensionManager.java
+++ b/src/main/java/net/minestom/server/extensions/ExtensionManager.java
@@ -612,8 +612,21 @@ public class ExtensionManager {
         return extensions.get(name.toLowerCase());
     }
 
+    @Nullable
+    @SuppressWarnings("unchecked")
+    public <E> E getExtension(@NotNull Class<E> type) {
+        return (E) extensions.values().stream()
+                .filter(extension -> extension.getClass().isAssignableFrom(type))
+                .findAny().orElse(null);
+    }
+
     public boolean hasExtension(@NotNull String name) {
         return extensions.containsKey(name);
+    }
+
+    public boolean hasExtension(@NotNull Class<?> type) {
+        return extensions.values().stream()
+                .anyMatch(extension -> extension.getClass().isAssignableFrom(type));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/extensions/ExtensionManager.java
```diff
@@ -612,10 +612,23 @@ public Extension getExtension(@NotNull String name) {
         return extensions.get(name.toLowerCase());
     }
 
+    @Nullable
+    @SuppressWarnings("unchecked")
+    public <E> E getExtension(@NotNull Class<E> type) {
+        return (E) extensions.values().stream()
+                .filter(extension -> extension.getClass().isAssignableFrom(type))
```
jglrxavpok, line 619: This won't work if different extensions try to access the same types: 
Extension A will have a different value when loading 'TestExtension.class' than the value Extension B will have.

src/main/java/net/minestom/server/extensions/ExtensionManager.java
```diff
@@ -612,10 +612,23 @@ public Extension getExtension(@NotNull String name) {
...
+                .findAny().orElse(null);
+    }
+
     public boolean hasExtension(@NotNull String name) {
         return extensions.containsKey(name);
     }
 
+    public boolean hasExtension(@NotNull Class<?> type) {
+        return extensions.values().stream()
+                .anyMatch(extension -> extension.getClass().isAssignableFrom(type));
```
jglrxavpok, line 629: Same goes.

jglrxavpok (changes requested):
Probably won't work as-is due to different extensions having separate extensions.
Have you checked in "real" setting with separate jars for two extensions and a third jar with a Minestom implementation?

Extension A will have a different value when loading 'TestExtension.class' than the value Extension B will have.

mworzala:
I'm not really sure the goal here, but if it is still relevant after #537 please re open as an issue or PR to that branch. In the mean time, closing due to inactivity and in favor of #537

PR #468, opened 2021-09-19
Add `getCoordinateType` to RelativeVec

LeoDog896:
For serialization purposes, CoordinateType should be exposed.

### DIFF
--- a/src/main/java/net/minestom/server/utils/location/RelativeVec.java
+++ b/src/main/java/net/minestom/server/utils/location/RelativeVec.java
@@ -27,6 +27,10 @@ public final class RelativeVec {
         this.relativeZ = relativeZ;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/utils/location/RelativeVec.java
```diff
@@ -27,6 +27,10 @@ public RelativeVec(@NotNull Vec vec, @NotNull CoordinateType coordinateType, boo
         this.relativeZ = relativeZ;
     }
 
+    public @NotNull CoordinateType getCoordinateType() {
```
TheMode, line 30: `coordinateType` would be better, in case we make it a record

LeoDog896: :+1: fixed

PR #470, opened 2021-09-19
Added isChunkLoaded to Instance.java

MrBretze:


TheMode:
I do not really see the point of this shortcut, what is your use-case?

MrBretze:
I use that for a RayTracing system with the BlockIterator, is just a shortcut, i can remove that if you want

TheMode:
There is also no reason for those methods to be abstract

PR #491, opened 2021-10-15
fix 'seeInvisiblePlayers' friendly flag

yusshu:
https://wiki.vg/Protocol#Teams

"Bit mask. 0x01: Allow friendly fire, 0x02: can see invisible players on same team."


vCamilx:
😍

PR #496, opened 2021-10-17
EntityProjectile improvements

RinesThaix:
1. Projectiles can hurt entities other than shooter in first 3 ticks of their life.
2. They don't stuck in non-solid blocks such as torches, high grass.
3. Previously all projectiles were treated as arrows whilst hitting an entity: they could hit only living ones and always increased victim's arrow count by one. Now this can be configured, also `EntityArrow` class is invented to provide an out-of-the-box way to achieve previous behaviour for those projectiles that are arrows.

### DIFF
--- a/src/main/java/net/minestom/server/entity/EntityProjectile.java
+++ b/src/main/java/net/minestom/server/entity/EntityProjectile.java
@@ -153,8 +153,7 @@ public class EntityProjectile extends Entity {
             } else {
                 pos = pos.add(direction);
             }
-            Block block = instance.getBlock(pos);
-            if (!block.isAir() && !block.isLiquid()) {
+            if (instance.getBlock(pos).isSolid()) {
                 teleport(pos);
                 return true;
             }
--- a/src/main/java/net/minestom/server/entity/EntityProjectile.java
+++ b/src/main/java/net/minestom/server/entity/EntityProjectile.java
@@ -18,6 +18,7 @@ import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Class that allows to instantiate entities with projectile-like physics handling.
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/entity/EntityProjectile.java
```diff
@@ -111,26 +150,22 @@ public void tick(long time) {
...
             onStuck();
         } else {
-            if (!super.onGround) {
-                return;
-            }
-            super.onGround = false;
-            setNoGravity(false);
-            onUnstuck();
+            onHit(((State.HitEntity) state).entity);
+            remove();
```
Bloepiloepi, line 154: I think whether the projectile is removed here should depend on which projectile it is. For example, it doesn't apply to fishing hooks (which should stay inside the entity) and piercing arrows (which should go through the entity). Subclasses can call `remove()` in `onHit(Entity)` if needed.

src/main/java/net/minestom/server/entity/EntityProjectile.java
```diff
@@ -111,26 +150,22 @@ public void tick(long time) {
...
      * @param posNow position after current tick.
-     * @return if an arrow is stuck in block / hit an entity.
+     * @return current state of the projectile.
      */
     @SuppressWarnings("ConstantConditions")
-    private boolean isStuck(Pos pos, Pos posNow) {
+    private State getState(Pos pos, Pos posNow) {
         if (pos.samePoint(posNow)) {
-            return true;
+            return State.StuckInBlock;
```
Bloepiloepi, line 168: This shouldn't always return `State.StuckInBlock`, because projectiles can never unstuck this way. It should check if the block is still solid, and if it isn't, return `State.Flying`.

src/main/java/net/minestom/server/entity/EntityProjectile.java
```diff
@@ -152,38 +187,93 @@ private boolean isStuck(Pos pos, Pos posNow) {
...
+            final Pos finalPos = pos;
+            Stream<Entity> victims = entities.stream().filter(entity -> entity.getBoundingBox().intersect(finalPos));
+
             /*
-              We won't check collisions with entities for first ticks of arrow's life, because it spawns in the
-              shooter and will immediately damage him.
+              We won't check collisions with self for first ticks of projectile's life, because it spawns in the
+              shooter and will immediately be triggered by him.
              */
             if (getAliveTicks() < 3) {
```
Bloepiloepi, line 213: In my experience 3 ticks isn't always enough, especially for slower projectiles like splash potions. I think 6 would be a better amount.

Bloepiloepi (changes requested):
Just a few changes I would do

KrystilizeNevaDies:
Closed due to inactivity and conflicts. Feel free to reopen.

PR #508, opened 2021-10-30
Allow custom socket addresses

TheMode:
- Remove deprecated `ResponseDataConsumer`
- Allow arbitrary `SocketAddress` mostly meant at adding support for unix domain sockets (allowing better communication with proxies)


TheMode:
Added in e0e5fe049da32b73f0819389f3a8a984fd3f6698

PR #509, opened 2021-10-30
Use NIO for extension manager instead of IO

LeoDog896:
Since Minestom bumped to JDK 17, this means that the FileSystem API (introduced in Java 14) can be used, thus the entire extension management can be switched to using NIO instead of IO.

However, this does require https://github.com/Minestom/DependencyGetter/pull/3

### DIFF
--- a/src/main/java/net/minestom/server/extensions/ExtensionManager.java
+++ b/src/main/java/net/minestom/server/extensions/ExtensionManager.java
@@ -354,10 +354,10 @@ public class ExtensionManager {
             LOGGER.info("Found indev folders for extension. Adding to list of discovered extensions.");
             final String extensionClasses = System.getProperty(INDEV_CLASSES_FOLDER);
             final String extensionResources = System.getProperty(INDEV_RESOURCES_FOLDER);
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/extensions/ExtensionManager.java
```diff
@@ -321,42 +324,41 @@ private Extension loadExtension(@NotNull DiscoveredExtension discoveredExtension
...
         List<DiscoveredExtension> extensions = new LinkedList<>();
 
-        File[] fileList = extensionFolder.listFiles();
+        Stream<Path> fileList = Files.list(extensionFolder);
 
-        if (fileList != null) {
-            // Loop through all files in extension folder
-            for (File file : fileList) {
+        // Loop through all files in extension folder
+        for (Path file : fileList.collect(Collectors.toList())) {
```
jglrxavpok, line 333: Why not use Stream#forEach?

jglrxavpok (commented):
Have you checked that loading an extension still works?
It should, but it should be tested anyway.

src/main/java/net/minestom/server/extensions/ExtensionManager.java
LeoDog896: Fixed :+1:

PR #517, opened 2021-11-04
Allow option to disable dummy block handler warning

LBuke:
`(BlockManager.getHandlerOrDummy) - WARN - Block minecraft:daylight_detector does not have any corresponding handler, default to dummy.`


LBuke:
Ignore the multiple commits, those have been reverted.

TheMode:
Couldn't you just configure the logger?

LBuke:
Yeah I guess the logger could be configured though i feel this would be nice for those who don't know how to configure the logger

PR #519, opened 2021-11-06
Fix CommandManager javadocs

ZakShearman:
This fixes some javadoc mistakes and an annotation placement issue within the CommandManager class.

There are also some ,more javadocs that are weirdly worded or maybe not marked properly:
- [ ] `executeServerCommand(String command)` I can't tell if the end is a to do note or genuinely meant to be there and is just weirdly worded.
- [ ] `commandExists(String commandName)` Not sure what the `or name` is on the end of this.

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandManager.java
+++ b/src/main/java/net/minestom/server/command/CommandManager.java
@@ -92,7 +92,7 @@ public final class CommandManager {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


TheMode:
`executeServerCommand` is probably just badly worded. This can be used to run a silent command (nothing printed in the console.

As for `commandExists` I probably meant "or not"

ZakShearman:
👍🏻 ill change those

PR #520, opened 2021-11-07
Area effect cloud holograms

Remynfv:
Changes holograms from invisible armor stands to area effect clouds with radius 0. This is much much faster for client side rendering, and increases client side performance drastically when there are lots of holograms present. 

Also adds a basic /hologram command to the test demo. Usage: /hologram <text> [count]


ZakShearman:
Is there a way to change the render distance of area of effect clouds?

PR #523, opened 2021-11-08
Investigate Better Physics

iam4722202468:


iam4722202468:
Block physics PR moved to https://github.com/Minestom/Minestom/pull/730

PR #526, opened 2021-11-11
Remove direct Mixin and ASM support

mworzala:
Removes direct support for Mixin and ASM in core Minestom or extensions.

### Goals
Below are the overarching goals of this PR. In addition the extension classloader isolation should remain functionally equivalent to the current system.
- [x] Remove Mixin
- [x] Remove ASM
- [x] Remove Bootstrap
- [x] Remove extension unload capability
- [x] Call preInitialize while server is still in the starting phase
- [x] ~~Create an official library which can add support back for these subsystems~~

### Motivation
Mixin is intended to modify code in environments where you do not have access to the source code, or you *must* do modification only at runtime. Neither of those cases are true in Minestom and I feel that without the crutch of "you can just use a Mixin" we will be forced to focus on better API design.

Additionally, we can remove the dependency on the Sponge maven repo :)

##### Extension Unload
This features is not directly related to Mixin/ASM, however, removing it at the same time makes sense. The motivation is that extension unload/reload is an extremely unsafe process without a good way to get it right. There are two major problematic use cases:
 - Callbacks passed into Minestom, such as event nodes. When we unload the extension those must go away as well or it cannot be fully unloaded, leading to issues when it is reloaded. This can be partially solved by tracking the class loader of these items, however it is unreliable. For example if the origin is a dependency.
 - Extension dependencies. If extension A depends on extension B and uses some of its apis, then we reload extension B there is a problem. Extension B is unloaded however the classes cannot be completely released because they are in use by extension A. This means that even after extension B comes back, A is referencing the old (now disabled) version of B. This problem is mitigated by unloading all of the dependents of an extension when it is unloaded, but that compounds with the first issue to create a very unstable environment.

Finally, we believe that the startup time of Minestom is fast enough that it will not make a huge difference in development.

### API Changes
* Removed `Bootstrap`, `MinestomRootClassLoader`. In addition, it is possible that other code will be broken by the lack of `Bootstrap`. It used to call `MinecraftServer#init` before any user code was executed, meaning it was valid to use `MinecraftServer` before calling `#init` yourself. This will no longer be valid.
* Isolation is strictly enforced now, meaning if you want to use a class from another extension it _must_ be in your `extension.json` file.


Articdive (approved):
Looks good on my end.

ZakShearman:
The following line still references the Bootstrap method https://github.com/Minestom/Minestom/blob/ad533533a4d66203978ed50b0c1bc6bd3d00875d/src/main/java/net/minestom/server/extensions/ExtensionManager.java#L282

mworzala:
> The following line still references the Bootstrap method
> 
> https://github.com/Minestom/Minestom/blob/ad533533a4d66203978ed50b0c1bc6bd3d00875d/src/main/java/net/minestom/server/extensions/ExtensionManager.java#L282

Fixed, thank you!

PR #527, opened 2021-11-12
Build system update

Articdive:
Updates the current build-system to be a little more modern. Mainly lining up to many Minecraft projects such as Krypton, ViaVersion and Adventure.

The changes include:
- Usage of Gradle's build-logic module (not fully utilized yet)
- Usage of version catalogues and type-safe project accessories (could be utilized more)
- Decluttering of the build scripts
- Creation of a demo subproject and subsequent separation of the original (demo and test)
- As a result of the previous point: the "core" test source-set can and should only contain JUnit Tests.
- Removal of the LWJGL source-set and dependencies (which will be an os-specific Extension once #529 is implemented)

This does also seem to be the direct in which Gradle wants to extend it's technology, so it should most likely be compatible with Gradle 8, whenever that happens.


BomBardyGamer:
Can I recommend the use of two things:
- version catalogues and type-safe project accessories
- build-logic module to isolate build logic

Good example projects for these:
- [Adventure](https://github.com/KyoriPowered/adventure) (doesn't use version catalogues yet, but uses build logic)
- [ViaVersion](https://github.com/ViaVersion/ViaVersion)
- [Krypton](https://github.com/KryptonMC/Krypton)

I am aware that version catalogues and type safe accessories are still previews, but I believe they are stable enough for usage.

The benefit to version catalogues is that your dependencies are easier to manage, as they are separated from the project, in a separate TOML file.
The benefit to build-logic is that your logic can easily be reused between modules, and it helps declutter your build scripts.

build-logic/src/main/kotlin/minestom.common-conventions.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+    // We only use Jetbrains Annotations
+    exclude("org.checkerframework", "checker-qual")
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
+    }
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
```
BomBardyGamer, line 15: Pretty sure you can't have both a toolchain and the source/target compatibility. I'm like 99% sure that's meant to thrown an error.

build-logic/src/main/kotlin/minestom.common-conventions.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+    withJavadocJar()
+    withSourcesJar()
+}
+
+tasks {
+    withType<JavaCompile> {
+        sourceCompatibility = JavaVersion.VERSION_17.toString()
+        targetCompatibility = JavaVersion.VERSION_17.toString()
```
BomBardyGamer, line 23: Don't you already set these above in the `java` configuration? Pretty sure you don't need to repeat yourself here.

BomBardyGamer (commented):
Overall, looks fine to me. Just a few things about the compatibility settings that confuse me a bit.

mworzala (commented):
Looks good to be, but it does seem Indra would simplify a bit.

PR #531, opened 2021-11-14
Use StringReaders for Commands/Arguments

GoldenStack:
This PR intends to make commands use `StringReader`s. This makes them much more configurable and removes the need for properties of arguments such as `allowSpace` and `useRemaining`.
This also makes it easier to get more accurate error messages, since the entire command and the point to which the dispatcher had read up to before an error occurred can be retrieved and displayed.

### Changes:
Command parsers and arguments will use string readers.
Command-related exceptions must be completely reworked so they can store the extra context from string readers.

### API changes
If someone is adding a new argument or modifying the suggestion callback, they will have to use string readers instead of the strings that are currently provided.
If they are using the current argument callback, they will have to use `CommandException`s instead of `ArgumentSyntaxException`s.
ArgumentTime returns a Long (meaning the number of ticks) instead of a duration
ArgumentResourceLocation returns the resulting NamespaceID instead of a String
The Command class now uses lists for its names and aliases instead of arrays
CommandSyntax is now a record and a few methods that used `Argument[]` now use `List<Argument>`
SuggestionEntry is now a record
Conditions (`net.minestom.server.command.builder.condition.Conditions`) prints slightly different messages

This list of API changes will be updated once the PR is complete


GoldenStack:
This is outdated, not very well-designed, overly ambitious, and it contains unnecessary API breakages.

PR #536, opened 2021-12-03
Add Potion flags

MrGazdag:
Adds the flags used in the `flags` parameter of the constructor and the byte returned by `flags()`.
This PR also includes a bit of documentation.

### DIFF
--- a/src/main/java/net/minestom/server/potion/Potion.java
+++ b/src/main/java/net/minestom/server/potion/Potion.java
@@ -53,6 +53,30 @@ public record Potion(PotionEffect effect, byte amplifier, int duration, byte fla
     public byte flags() {
         return flags;
     }
+    
+    /**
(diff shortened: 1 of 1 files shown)


mworzala:
I would prefer to have methods such as `isAmbient`, `hasParticles`, etc instead of exposing the flags field directly

MrGazdag:
> I would prefer to have methods such as `isAmbient`, `hasParticles`, etc instead of exposing the flags field directly

Honestly i would prefer both, i'm going to add the methods

mworzala:
Not sure what I was talking about, its a record we can't not have the `flags` method

MrGazdag:
> Not sure what I was talking about, its a record we can't not have the `flags` method

We can have methods to query the individual flag values though

mworzala (approved):

PR #537, opened 2021-12-04
Extension Improvements

### LINKED ISSUE #529: Extension Improvements
Some general improvements which would be good for extensions:
 - [x] Soft dependencies (eg named as a dependency for load order, but allow the extension to start without it)
 - [ ] Version matching (eg depend on LuckPerms 5.x.x specifically)
 - [x] Move to NIO
 - [x] Share external dependencies of the same version between extensions (otherwise there can be conflicts)
 - [x] Common repo names as strings (or just added automatically?), eg
```json
{
    "externalDependencies": {
        "repositories": [
            "MavenCentral",
            "JitPack",
            {"name": "sponge", "url": "..."}
        ],
        "artifacts": ["..."]
    }
}
```
 - [x] Remove `get` prefixes when there is no `set` equivalent
 - [x] Platform dependent external dependencies, see [comment](https://github.com/Minestom/Minestom/issues/529#issuecomment-980143828)
 - [x] ~Revisit extension callback order (`preInitialize` should be called in a phase when startup tasks can still be done)~ Done in #526 
 - [x] "autoscan" mechanism to detect an `extension.json` in the root class loader and automatically load it as an extension. This means if the test server is running with your project classpath (either a main file in `test` or in the future a gradle plugin) the extension will load automagically.
 - [x] Allow `preInitialize` to return a status enum to indicate whether an extension has loaded correctly. If not, it will be unloaded and any dependents will not be loaded
 - [x] ~Registry modification (depends on an internal version of this)~ (for another PR, once the system is closer to finalized)
 - [ ] Internal dependencies (only maven), which are added directly to the extension class loader, instead of a child. (fixes a library loading issue with LWJGL. The two dependencies are isolated meaning it can never find the libraries)

If anything is missed or you disagree, feel free to mention.

Note: Unrelated to #526

mworzala:
See related issue for description.

Depends on #526. Currently based on that PR

Closes #529


ZakShearman:
Don't think this spelling mistake was introduced by this branch but `DiscoveredExtension.java`, `/** Version of this extension, highly reccomended to set it. */` should be recommended

mworzala:
Very dead

PR #544, opened 2021-12-20
Sync Player#allowFlying when Player#gameMode changes.

DigitalJokerMan:


mworzala:
I would be curious how this plays with manually setting the flight ability.

eg: in survival with flight true, then set to creative and back to survival. can the client still fly?

TheMode:
Done in 6a83fd6ac8eb521ff053a36a627b6260da6e4746

PR #546, opened 2021-12-22
Fix entity metadata

R0bbyYT:
This pull request fixes in the entity metadata that in some setters the index was retrieved instead of set.


mworzala:
Very cool

PR #555, opened 2021-12-26
Make PlayerUseItemOnBlockEvent.getBlockFace() return a BlockFace

Marlonzockt:
Currently PlayerUseItemOnBlockEvent.getBlockFace() returns a Direction instead of a BlockFace. Since there's a BlockFace.toDirection() method you can easily convert a BlockFace back to a Direction if you really need to, however there's no such easy way of getting a BlockFace from a Direction.


TheMode:
Oops

PR #552, opened 2021-12-26
Fix IndexOutOfBoundsException from getSection

RealGatt:
DynamicChunk's `getSection` was not checking for an out of bounds exception. This caused chunks to load indefinitely/not load at all. Solution checks and captures the exception and returns a new empty Section.

PR #561, opened 2021-12-28
check for an outofbounds exception from getSection

RealGatt:
Duplicate of #552 due to me doing a boo-boo and making this change in my primary 1.18 branch 🙃 

See the original ticket for explanation/etc (apologies)

### DIFF
--- a/src/main/java/net/minestom/server/instance/DynamicChunk.java
+++ b/src/main/java/net/minestom/server/instance/DynamicChunk.java
@@ -102,7 +102,11 @@ public class DynamicChunk extends Chunk {
 
     @Override
     public @NotNull Section getSection(int section) {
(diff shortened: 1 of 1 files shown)


TheMode:
I still do not believe it to be a proper fix, we are increasing the risk of people modifying the section without understanding that it is a placeholder for "not present". Not to mention initialization cost.

Kebab11noel:
You should find the root cause of the issue instead of making a change which just makes it fail silently, it's the caller's responsibility to not request a section outside of the world.

RealGatt:
Looking into the root cause of the issue now. Keeping this open (unless you want it closed) so I remnember to work on this branch more

RealGatt:
Issue has been resolved w/ latest /1.18 commits

PR #562, opened 2021-12-28
Fix DeclareRecipePacket

RealGatt:
Packet requires the recipes to include the recipe type. This commit fixes that inconsistency.

Makes RecipeType a public enum

Allows for registering multiple recipes at the one time


TheMode:
Seems to have a lot of style change, also you may be able to write the type in the packet `#write` method directly, as it is common to all recipes.

RealGatt:
Fixed the style changes. Not sure how I would go about implementing the writing of per-recipe recipe-types within the packet `#write` method directly, without exploding out the function to take more effort than what it is probably worth.

RealGatt:
Forgot to attach this initially. Just proof of changes functioning

Granite -> Spruce Wood is a custom recipe used for testing.

image

RealGatt:
Branch now also fixes Shaped Recipes from crashing clients when sent.

PR #569, opened 2021-12-30
Fixed ArrayIndexOutOfBoundsException when placing or breaking a block close to max / min InstanceContainer height

### LINKED ISSUE #568: ArrayIndexOutOfBoundsException on block update at vertical instance limit
When breaking or placing a block at the world min/max y in an InstanceContainer, an ArrayIndexOutOfBoundsException gets thrown

Marlonzockt:
Fixed #568


Marlonzockt:
For some reason this pull request contains 4 commits although only one of them actually does anything. How do I avoid that?

src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -488,7 +488,7 @@ private void executeNeighboursBlockPlacementRule(@NotNull Point blockPosition) {
         for (int offsetX = -1; offsetX < 2; offsetX++) {
             for (int offsetY = -1; offsetY < 2; offsetY++) {
                 for (int offsetZ = -1; offsetZ < 2; offsetZ++) {
-                    if (offsetX == 0 && offsetY == 0 && offsetZ == 0)
+                    if (offsetX == 0 && offsetY == 0 && offsetZ == 0 || offsetY < getDimensionType().getMinY() || offsetY > getDimensionType().getTotalHeight())
```
GreatWyrm, line 491: I believe you need to compare against the neighborY number, you're just comparing against the offset, not the true position

GreatWyrm (changes requested):

src/main/java/net/minestom/server/instance/InstanceContainer.java
Marlonzockt: you're right, my bad

GreatWyrm (approved):
Looks good to me now, may want to reorder the `neighborX, neighborY, neighborZ` values so you only calculate `neighborY` value before the out of bounds check, but that probably doesn't matter

PR #571, opened 2022-01-01
Low level world generation API

Kebab11noel:
This pr changes the low level API to allow better separation of abstraction layers and changes the generation to section based while providing full backwards compatibility.


Kebab11noel:
Close in favour of #574

Issue #534, opened 2021-11-20
Non solid blocks get replaced with air


emortaldev:
Non solid blocks get replaced with air when a player attempts to place a block on the block face behind it.

[Video for better explaination](https://cdn.discordapp.com/attachments/715174453398601748/911708895670583326/2021-11-20_20-05-04.mp4) (I am right clicking with the TNT)

Eoghanmc22:
this is probably caused by BlockPlacementListener never checking if the block at placementPosition is non air. also how are you canceling block placements? i can only reproduce with the block getting replace with what im holding or if i cancel the event no change (the behavior shown might also be possible if minestom thinks the player is in adventure mode but the player thinks it is in survival, possibly by setting the gamemode in the login event). also idk if a simple air check is good enough to fix this because some blocks (ie tall grass) should be able to be overridden but some blocks with solid = false like saplings shouldn't be overridden (at least if you want to preserve vanilla behavior).

emortaldev:
I am cancelling the block placements by cancelling the block break and block place event when the player is in adventure mode

CatDevz:
+1 for this issue. Occurs in both Survival and Adventure

TheMode:
Fixed by 517d6a3b7c6640082716a144a24f91bab6c349be

emortaldev:
Reopening because the issue still appears

CatDevz:
This issue is different

PR #583, opened 2022-01-10
Fix issue #534

CatDevz:
This PR fixes issue #534 

The client was always being sent air blocks to "keep them in sync", when this often wouldn't keep them in sync. This PR changes that to send the block actually in the instance.

### DIFF
--- a/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
+++ b/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
@@ -92,10 +92,11 @@ public class BlockPlacementListener {
         final Point placementPosition = blockPosition.add(offsetX, offsetY, offsetZ);
 
         if (!canPlaceBlock) {
-            // Send a block change with AIR as block to keep the client in sync,
+            // Send a block change with the real block in the instance to keep the client in sync,
             // using refreshChunk results in the client not being in sync
             // after rapid invalid block placements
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/listener/BlockPlacementListener.java
```diff
@@ -92,10 +92,11 @@ public static void listener(ClientPlayerBlockPlacementPacket packet, Player play
         final Point placementPosition = blockPosition.add(offsetX, offsetY, offsetZ);
 
         if (!canPlaceBlock) {
-            // Send a block change with AIR as block to keep the client in sync,
+            // Send a block change with the real block in the instance to keep the client in sync,
             // using refreshChunk results in the client not being in sync
             // after rapid invalid block placements
-            player.getPlayerConnection().sendPacket(new BlockChangePacket(placementPosition, Block.AIR));
+            final Block block = interactedChunk.getBlock(placementPosition);
```
TheMode, line 98: I would suggest retrieving the block from the instance, as it comes with additional synchronization not present in the chunk method.

TheMode (changes requested):

src/main/java/net/minestom/server/listener/BlockPlacementListener.java
CatDevz: Got it, just changed it.

PR #586, opened 2022-01-12
Allowed for 'f' to be pressed in inventories (Fixes #585)

### LINKED ISSUE #585: Pressing 'F' in inventory will swap item to crafting area (but not in creative...)
[https://www.youtube.com/watch?v=4toi6mHPWOg](https://www.youtube.com/watch?v=4toi6mHPWOg)

This applies only to Survival and Adventure, swapping works as expected in creative

CatDevz:
Like the title says, this fixes issue #585


TheMode:
You probably need to handle click events if you do not want players to bypass them

CatDevz:
I couldn't find a way to do it in the `InventoryClickProcessor::changeHeld`, as it doesn't seem to handle the location items are placed. However I removed the duplicate code.

PR #620, opened 2022-01-29
Fix player settings

emortaldev:


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2078,6 +2080,10 @@ public MainHand getMainHand() {
             return mainHand;
         }
 
+        public boolean getTextFilteringEnabled() { return enableTextFiltering; }
```
TheMode, line 2083: `enableTextFiltering`

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2078,6 +2080,10 @@ public MainHand getMainHand() {
...
+
+        public boolean getAllowsListing() { return allowServerListings; }
```
TheMode, line 2085: `allowServerListings`

TheMode (changes requested):

PR #621, opened 2022-01-29
Add setCooldown

emortaldev:
Updated version of https://github.com/Minestom/Minestom/pull/434

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -1948,6 +1948,35 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
         inventory.setBoots(itemStack);
     }
 
+    /**
+     * Sets a cooldown on a material (like when an enderpearl is used)
+     *
+     * @param material the material to add a cooldown to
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1948,6 +1948,35 @@ public void setBoots(@NotNull ItemStack itemStack) {
...
+
+    /**
+     * Sets a cooldown on a material (like when an enderpearl is used)
+     *
+     * @param material the material to add a cooldown to
+     * @param ticks how long the material will be on cooldown
+     * @param lagCompensation whether another packet is sent later to remove cooldown (incase the server lags)
+     */
+    public void setCooldown(Material material, int ticks, boolean lagCompensation) {
+        final SetCooldownPacket packet = new SetCooldownPacket(material.id(), 0);
```
GreatWyrm, line 1969: Shouldn't the second argument be `ticks` rather than 0?

emortaldev: whoops yep

GreatWyrm (approved):

PR #691, opened 2022-02-17
Shut down tinylog during Minestom shutdown hook

LeoDog896:
For tinylog configurations that handle files, shutting down a Minestom server normally means the log file won't be able to write the output of the shutdown. This moves the shutdown hook for tinylog to Minestom's shutdown hook, shutting down the logger when Minestom is shut down.


TheMode:
Is there no other way? I am not really for making the codebase depends on tinylog.

KrystilizeNevaDies:
Closed due to inactivity. Feel free to reopen.

PR #693, opened 2022-02-18
Make argb values with no alpha return MapColors#NONE

LooFifteen:
This PR allows `MapColors#closestColor` to return colour `NONE` when no alpha value is provided.

[Related Discussion](https://github.com/Minestom/Minestom/discussions/692)


src/main/java/net/minestom/server/map/MapColors.java
```diff
@@ -201,6 +201,7 @@ private static void fillRGBArray() {
     }
 
     public static PreciseMapColor closestColor(int argb) {
+        if (argb << 24 == 0) return new PreciseMapColor(NONE, Multiplier.x1_00);
```
jglrxavpok, line 204: That's not ARGB but RGBA.

src/test/java/map/TestMapColors.java
```diff
@@ -0,0 +1,20 @@
...
+public class TestMapColors {
+
+    @Test
+    public void closestColor() {
+        assertEquals(MapColors.closestColor(0x539D87FF).getBaseColor(), MapColors.ICE);
+    }
+
+    @Test
+    public void closestColorNoAlpha() {
+        assertEquals(MapColors.closestColor(0xFFFFFF00).getBaseColor(), MapColors.NONE);
```
jglrxavpok, line 17: Ditto

jglrxavpok (changes requested):
While I agree with the intent of this PR, you risk breaking all current code using this API.
I do think alpha support is interesting, but we should not change too much from the current format (RGB -> RGBA requires a shift in addition to specifying the alpha bits, while RGB->ARGB only requires the addition of alpha bits)

src/main/java/net/minestom/server/map/MapColors.java
SLLCoding: Oh, whoops

KrystilizeNevaDies:
Closed due to no longer being relevant. Feel free to reopen.

PR #699, opened 2022-02-19
Make MapColors a registry object and auto generated.

Articdive:
Title does say it all. Depends on https://github.com/Minestom/MinestomDataGenerator/pull/6


jglrxavpok:
As it is breaking, I would merge this PR at the same time as https://github.com/Minestom/Minestom/pull/693 
Other than that, I don't think I have any complaint

PR #707, opened 2022-02-21
Resetting player flying speed after changing gamemode

Pelfox:
Hello. Today I found a bug, that causes the player not to reset flying speed after changing the gamemode. So, I decided to fix it.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -1252,6 +1252,7 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
                 this.flying = false;
             }
         }
+        this.flyingSpeed = 0.05f; // reset flying speed
     }
 
     /**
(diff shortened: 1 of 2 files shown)


TheMode:
Could you test it? In this case I assume in https://github.com/Minestom/Minestom/blob/master/src/test/java/net/minestom/server/entity/player/PlayerIntegrationTest.java#L30

Pelfox:
This there any CONTRIBUTING.md or something like this? I don't know how to test it. Will something like
```java
        player.setGameMode(GameMode.SPECTATOR);
        player.setFlyingSpeed(1.0f);
        player.setGameMode(GameMode.CREATIVE);
        assertEquals(player.getFlyingSpeed(), 0.05f);
```
good?

TheMode:
Those are normal junit tests - but with a small API to retrieve game objects.

The general idea is to add tests that fail on master, but that your changes fix. It has the advantages of:
- Properly explaining what you are fixing
- Ensure that future contributors do not break those changes without knowing

Pelfox:
So, this RP does nothing (as I and Moulberry saw on his stream), so I'll close it

PR #710, opened 2022-02-22
Resetting player flying speed after changing gamemode

Pelfox:
Hi. This is an improved version of my PR #707. This PR contains a fix, that resets the player's speed after changing the gamemode.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -1252,6 +1252,10 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
                 this.flying = false;
             }
         }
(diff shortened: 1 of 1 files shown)


Moulberry:
Make sure to check `isActive()` so that you don't send the packet if the player isn't in the PLAY state

TheMode:
Would be worth an explanatory comment

TheMode:
I was mostly referring to `!isActive()` as it does not make a ton of sense to me.

Pelfox:
OMG, I’m so stupid. Sorry

TheMode:
No problem!

And you generally want comments to either explain the unsaid, or ease the reading. In this case it could be (given as example) `Sync abilities with the client to avoid X issue, we must therefore also ensure that the modified abilities are correct)`

PR #401, opened 2021-08-11
Add basic permission manager

LeoDog896:
This adds a permission manager capable of registering string permissions, useful for tab-completing in permission plugins

PR #719, opened 2022-02-27
Add wildcard support to no-verifier hasPermission

LeoDog896:
This adds wildcard support to the #hasPermission(Permission permission) method.

EX:

If a player has the permission `test.*`, `test.one` and `test.two` would count as valid permissions when passed through the above hasPermission method.

This is part of one of the two PRs (#401) to help bring better support for permission plugins.

### DIFF
--- a/src/main/java/net/minestom/server/permission/PermissionHandler.java
+++ b/src/main/java/net/minestom/server/permission/PermissionHandler.java
@@ -67,6 +67,15 @@ public interface PermissionHandler {
             if (permissionLoop.equals(permission)) {
                 return true;
             }
+
+            if (permissionLoop.getPermissionName().endsWith(".*")) {
+
+                // Remove (.*) from the end of the string
+                String permissionPrefix = permissionLoop.getPermissionName().substring(0, permission.getPermissionName().length() - 2);
+
+                if (permission.getPermissionName().startsWith(permissionPrefix))
+                    return true;
+            }
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


TheMode:
Could you test it?

LeoDog896:
Added. Works fine on my machine, just have to wait for the CI builds here.

src/test/java/net/minestom/server/permission/TestPermissions.java
```diff
@@ -62,6 +64,15 @@ public void hasPermissionClass() {
         assertTrue(player.hasPermission(permission2));
     }
 
+    @Test
+    public void hasPermissionWildcard() {
+        assertFalse(player.hasPermission(permission3));
+
+        player.addPermission(permission3);
+
+        assertTrue(player.hasPermission(permission3));
```
Marlonzockt, line 73: That doesn't really test the wildcard feature, does it?

src/main/java/net/minestom/server/permission/PermissionHandler.java
```diff
@@ -67,6 +67,15 @@ default boolean hasPermission(@NotNull Permission permission) {
             if (permissionLoop.equals(permission)) {
                 return true;
             }
+
+            if (permissionLoop.getPermissionName().endsWith(".*")) {
+
+                // Remove (.*) from the end of the string
+                String permissionPrefix = permissionLoop.getPermissionName().substring(0, permission.getPermissionName().length() - 2);
```
mworzala, line 74: This is not correct. It will fail on `*`.

mworzala (commented):
+1 to this not being tested very well. I would expect **at minimum** tests for wildcarding multiple levels deep, just * (see other comment), and unhappy/weird cases

KrystilizeNevaDies:
Closed due to no longer being relevant. Feel free to reopen.

Issue #599, opened 2022-01-24
Multiple AdvancementTabs Exception


EpicPlayerA10:
When player leaves with more than one viewable advancement tabs then I get this exception:
```
java.util.ConcurrentModificationException
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1597)
	at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1620)
	at net.minestom.server.entity.Player.remove(Player.java:465)
	at net.minestom.server.entity.Entity.lambda$scheduleNextTick$5(Entity.java:221)
	at net.minestom.server.timer.Task$Builder$1.get(Task.java:68)
	at net.minestom.server.timer.Task$Builder$1.get(Task.java:59)
	at net.minestom.server.timer.SchedulerImpl.handleTask(SchedulerImpl.java:111)
	at net.minestom.server.timer.SchedulerImpl.lambda$processTick$3(SchedulerImpl.java:61)
	at org.jctools.queues.MessagePassingQueueUtil.drain(MessagePassingQueueUtil.java:39)
	at org.jctools.queues.BaseMpscLinkedArrayQueue.drain(BaseMpscLinkedArrayQueue.java:638)
	at org.jctools.queues.MpscUnboundedArrayQueue.drain(MpscUnboundedArrayQueue.java:23)
	at org.jctools.queues.MpscUnboundedArrayQueue.drain(MpscUnboundedArrayQueue.java:63)
	at net.minestom.server.timer.SchedulerImpl.processTick(SchedulerImpl.java:58)
	at net.minestom.server.timer.SchedulerImpl.processTick(SchedulerImpl.java:44)
	at net.minestom.server.entity.Entity.tick(Entity.java:528)
	at net.minestom.server.thread.TickThread.tick(TickThread.java:64)
	at net.minestom.server.thread.TickThread.run(TickThread.java:42)
```

PR #724, opened 2022-03-02
Fix exception on player disconnect

EpicPlayerA10:
Fix exception when player disconnects with more than one registered advancement tabs
Related issue: #599


TheMode:
You should fix the issue by replacing the hash set to a copy CopyOnWriteSet in the advancement code. Manually copying it shouldn’t be needed.

also those tests involving the server state should end by IntegrationTest, not only Test

TheMode:
LGTM! And thanks for including tests :)

PR #743, opened 2022-03-09
Attempt to Fix Pathfinding

iam4722202468:


iam4722202468:
Current pathfinding uses full block steps. Fixing bounds on blocks doesn't fix pathfinding issues.

iam4722202468:
Current pathfinding uses full block steps. Fixing bounds on blocks doesn't fix pathfinding issues.

PR #745, opened 2022-03-10
Entity Pathfinding API + Stable Implementation

KrystilizeNevaDies:
This pathfinding api aims to flesh out the control and live data that developers have access to while pathfinding.
Previously, pathfinding was limited to simply setting the target position.

Within this new api, developers will be able to specify movement options, manually create pathfinding tasks, and handle their own threading options. Additionally, developers will be able to access live data via task execution objects, enabling features such as pathfinding visualizers, and dynamic pathfinding tasks.

Example of the task/execution api:
```java
        EntityCreature entity = ...;
        double speed = ...;
        Point target = ...;

        // Create the task to move the entity.
        // The static walk task walks the entity to a static position at a static speed, without jumping.
        StaticWalkTask walkTask = PathfindTask.walkTo(target, speed);

        // Now schedule this task.
        var execution = entity.schedulePathfind(walkTask);

        // The task is now scheduled to run in the future.
        // Now lets look at what path the entity will take.
        
        // The execution object provides many ways to observe the task.
        // Some execution objects may even support dynamically changing the path.
        execution.result().thenAccept(result -> {
            Queue<Point> path = result.fullPath().join();
            System.out.println("Entity Path: " + path);
        });
```

There is also a new a-star pathfinding engine implementation used by default. This implementation is not focussed on speed, but on stability.

### DIFF
--- a/src/main/java/net/minestom/server/entity/ai/goal/RandomStrollGoal.java
+++ b/src/main/java/net/minestom/server/entity/ai/goal/RandomStrollGoal.java
@@ -8,6 +8,10 @@ import org.jetbrains.annotations.NotNull;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 public class RandomStrollGoal extends GoalSelector {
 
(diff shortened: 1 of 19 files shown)


src/main/java/net/minestom/server/entity/ai/goal/RandomStrollGoal.java
```diff
@@ -32,17 +43,26 @@ public boolean shouldStart() {
...
-        while (remainingAttempt-- > 0) {
-            final int index = random.nextInt(closePositions.size());
-            final Vec position = closePositions.get(index);
-
-            final var target = entityCreature.getPosition().add(position);
-            final boolean result = entityCreature.getNavigator().setPathTo(target);
-            if (result) {
-                break;
+        final int remainingAttempt = closePositions.size();
+        CompletableFuture.runAsync(() -> {
```
BomBardyGamer, line 47: Got a few things I want to say here:
1. Shouldn't the future use an existing execution pool? Because otherwise, it'll use `ForkJoinPool.commonPool()`, which may not be desired, as we already have existing scheduling pools.
2. Couldn't this be a scheduled task using the scheduler so it can be tracked more easily?
3. If we are going to stick with futures for these things, I think the method should return the future, so that users can call back to it.
4. I think it might even be better if the goals weren't the ones handling the concurrency at all, and instead that was left to the thing managing all of the AI processing.
5. Why is this the only goal that uses a future to compute the logic? Am I missing something here?

I may be talking out of my ass here, idk, but I think these are valid points.

BomBardyGamer (commented):
Overall, looks like a really well designed API! If I was making a Minestom-based server, I'd really be looking forward to this. Just the one comment from me.

PR #753, opened 2022-03-12
Applying block placement rules right after chunks loading

RinesThaix:


KrystilizeNevaDies:
Is this relevant anymore?

Also are we sure that we should be running this placement rules on world load? It might be better to keep the blockstates, as arbitrary blockstates might be what the user wants.

mworzala:
Second what Krystilize said, arbitrary block states are very relevant.

PR #762, opened 2022-03-13
Remove faces and cartesian product

iam4722202468:


iam4722202468:
Definitely good doesn't have any problems

PR #768, opened 2022-03-14
Sort command registration in Demo/Main

RickHeadle:
Sorting command registration in Demo.
It changes absolutely nothing in functionality. Just formatting.


TheMode:
As you said it does not change anything, but it can create unnecessary conflicts. 
Furthermore we do keep the right to add various commands there in the future, and I doubt that anyone will take time to maintain such ordering

PR #817, opened 2022-03-27
Standardization start logic for providers

TheMeinerLP:
This allows to provider to override via environment variables port and host.
Also to use unix socket paths. 
Documentation describes how it works

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -355,7 +355,19 @@ public final class MinecraftServer {
      * Starts the server.
      * <p>
      * It should be called after {@link #init()} and probably your own initialization code.
-     *
+     * <br>
+     * Can be controlled/overridden via jvm properties.<br>
+     * <strong>Unix</strong><br>
+     * -Dunix=/var/run/minestom/minestom.sock
+     * <br>
+     * <strong>Host and Port</strong><br>
+     * -Dhost=127.0.0.1<br>
+     * -Dport=255565<br>
+     *<br>
+     * <strong>Full commandline example</strong><br>
+     * java -Dhost=127.0.0.1 -Dport=25565 -jar minestom.jar<br>
+     * or <br>
+     * java -Dunix=/var/run/minestom/minestom.sock -jar minestom.jar<br>
      * @param address the server address
      * @throws IllegalStateException if called before {@link #init()} or if the server is already running
      */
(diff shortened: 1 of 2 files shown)


Marlonzockt:
I don't think you should remove the old methods

TheMeinerLP:
Why ?
I know breaking changes. But i don't the point currently. 
Only if i add fallback support for old apis

Marlonzockt:
you can literally just recreate the old method and make it a shorthand for your builder syntax

TheMode:
As previously mentioned you may want to look into `ServerProcess` instead.

There is absolutely no reason to break every single project for a niche utils

TheMeinerLP:
@TheMode What you preferre a builder pattern with old methods they use internal the builder or applyed direcly at start on the ServerProcessImpl?

TheMode:
Is there even a need for a builder? Seems like you could achieve the same with a single `MinecraftServer#start()` method retrieving the environment variables.

Your env keys are also way too generic, perhaps that you meant jvm properties instead.

TheMeinerLP:
Okay i do that

TheMeinerLP:
@TheMode Hopefully its better

TheMode:
I meant a new `start` method, overriding the address does not make much sense.

And in any case I am not entirely convinced by its usefulness.

TheMeinerLP:
The main behavor is here to integrated cloud provider to make finish develope software easy to use.
No 30 diffrent config system or integration. Thats the point of this pr.
here the discussion at general on discord.
image
https://discord.com/channels/706185253441634317/706186227493109860/957571625748930640

TheMeinerLP:
@TheMode Please give me some feedback what your best practicing for this integration and  keep in mind the point of this project. 
The plan is currently make a stanalizion for **Administartors** they get from a developer a finish product to control port and host. 
Worst exmaple is here spring boot.
But as a developer and a administartor i like to have one schema for change this. And not 30 differen systems from each developer.

TheMode:
This is what I had in mind
`MinecraftServer#start()`
```java
public void start() {
    final String unixPath = System.getProperty("unix");
    if (unixPath != null) start(UnixDomainSocketAddress.of(unixPath));
    final String host = System.getProperty("host", "0.0.0.0");
    final int port = Integer.getInteger("port", 25565);
    start(new InetSocketAddress(host, port));
}
```

TheMeinerLP:
Problem here is, if the developer use the other one with port and host. How you can garantee a provider they can control it ?
Like for security laws

TheMode:
If the developer explicitly wants to hardcode the address I do not see why you would want to prevent it, overriding provided addresses is just asking for confusion. The method has great defaults and could become the first choice for both flexibility and simplicity.

TheMeinerLP:
And again is a project open not to create standars they useful for non developers.

PR #831, opened 2022-03-30
Fix InventoryClickProcessor

jhooc77:
Fixed some bugs related to the stacking rule.
`(int) ((float) cursorAmount / (float) slotCount);` causes floating point error.
Both problems also need to be fixed in #610.

And It looks better to use the stacking rule's amount in block placing.


TheMode:
Master currently does not have exhaustive click tests (I have just pushed a few from the inventory rework branch) so I am not comfortable merging it now. This PR will likely have to wait until I or you write these.

PR #855, opened 2022-04-03
Use toUpperCase(Locale.ROOT) instead of toLowerCase

LeoDog896:
Loading dependencies if they arent uppercase doesn't work. This PR fixes that issue.

For example, if you had the dependency "Test", the code would expect "TEST" but it instead would transform it to "test".


BitCrack:
I had to fix that too, though personally instead of switching from lower case to upper case everywhere, I just switched the casing here https://github.com/Minestom/Minestom/blob/db6e1e5b3d656eea45697a624152ff9171393668/src/main/java/net/minestom/server/extensions/ExtensionManager.java#L130

Issue #808, opened 2022-03-25
The sign's not loaded below y = 0


DustKing5:
Minestom version: 8dcb3191f6

The world created in 1.18.2

![2022-03-25_22 31 37](https://user-images.githubusercontent.com/67860797/160203391-58006c05-cc65-434f-8256-81544107741a.png)

Marlonzockt:
Have you tried the fix discussed in #661 already? (Part is in the question, part in the answer). That might fix it

DustKing5:
Yes, that's the first thing what I tried

DustKing5:
After a few experiments, I noticed that I experience this error under y = 0

TheMode:
Could you provide a reproducer?

DustKing5:
`InstanceManager instanceManager = MinecraftServer.getInstanceManager();`
`instanceContainer = instanceManager.createInstanceContainer(new AnvilLoader("world"));`

Compressed world folder:
[world.zip](https://github.com/Minestom/Minestom/files/8355087/world.zip)

TheMode:
Should be fixed by 740ffc884684e37bdbfa90f1c904a38c0e2ed7ce

PR #861, opened 2022-04-03
Fix bad conversion of block index using negative Y

AlexanderMandera:
Fixes issue #808.

We got the same problem and I debugged everything.
I found out that the NBT is sent right but the X/Y/Z coordinates inside the ChunkData packet is wrong:
image

The expected coordinates for this would be -72 -42 -10.

I found out that the block index does not respect negative values (I think because it was missed during 1.18 Minestom update). It did work pre-1.18 because there weren't any tile entities on Y < 0, right?

But now when there are negative integers, the two-complement value does not convert back from the block index the right way.

I am no expert at binary arithmetics but I tried
* to use the 28th bit of the index to make a negative flag
* use the other bits to store the absolute value of Y and
* check for the negative flag in the `getBlockPosition` function to negate the value again.

I have tested it and it works fine now (signs appear again on negative Y values).

I don't know if this is the best solution for this problem, so I appreciate some hints to do it better. :)

### DIFF
--- a/src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
+++ b/src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
@@ -238,7 +238,7 @@ public final class ChunkUtils {
      * @return the chunk position Y of the specified index
      */
     public static int blockIndexToChunkPositionY(int index) {
-        int y = index >>> 4 & 0xFF;
+        int y = (index & 0x07FFFFF0) >>> 4;
         if(((index >>> 27) & 1) == 1) y = -y; // Sign bit set, invert sign
         return y; // 4-28 bits
     }
--- a/src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
+++ b/src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
@@ -216,9 +216,7 @@ public final class ChunkUtils {
      */
     public static @NotNull Point getBlockPosition(int index, int chunkX, int chunkZ) {
         final int x = blockIndexToChunkPositionX(index) + Chunk.CHUNK_SIZE_X * chunkX;
-        int y = index >>> 4 & 0xFF;
-        if(((index >>> 27) & 1) == 1) y = -y; // Sign bit set, invert sign
-
+        final int y = blockIndexToChunkPositionY(index);
         final int z = blockIndexToChunkPositionZ(index) + Chunk.CHUNK_SIZE_Z * chunkZ;
         return new Vec(x, y, z);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
```diff
@@ -198,7 +198,12 @@ public static int getBlockIndex(int x, int y, int z) {
         z = z % Chunk.CHUNK_SIZE_Z;
 
         int index = x & 0xF; // 4 bits
```
Eoghanmc22, line 200: I dont think this is needed

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
```diff
@@ -211,7 +216,9 @@ public static int getBlockIndex(int x, int y, int z) {
      */
     public static @NotNull Point getBlockPosition(int index, int chunkX, int chunkZ) {
         final int x = blockIndexToChunkPositionX(index) + Chunk.CHUNK_SIZE_X * chunkX;
-        final int y = index >>> 4 & 0xFF;
+        int y = index >>> 4 & 0xFF;
```
Eoghanmc22, line 219: i think this should be `final int y = index << 4 >> 8 & 0xFF`

Eoghanmc22: actually, why is this mask `& 0xFF` maybe this should use blockIndexToChunkPositionY and then change that to use the change i mentioned?

TheMode:
Could you test it in `CoordinateTest`?

AlexanderMandera:
@TheMode I found that the affected functions are not covered by any tests.
Everything in CoordinateTest passed though.

TheMode:
Yes, you have to create new tests

AlexanderMandera:
Ok, got it, will do it later

src/test/java/net/minestom/server/coordinate/CoordinateTest.java
```diff
@@ -100,4 +100,18 @@ public void toSectionRelativeCoordinate() {
         assertEquals(0, ChunkUtils.toSectionRelativeCoordinate(32));
         assertEquals(1, ChunkUtils.toSectionRelativeCoordinate(33));
     }
+
+    @Test
+    public void testBlockIndex() {
```
TheMode, line 105: You need to test every possibility (ensuring that all coordinates give a different index, prevent potential overflow). There is otherwise no way to know if you broke something.

There is also no need to use `Pos`

AlexanderMandera: I have tried to get suitable test cases. I don't know if that are all but I also checked for Y overflow.

Y values between -2^23+1 to 2^23-1 convert correctly.
Everything above and below that results in a wrong conversion.

I don't know if I should catch this case and throw an exception, what do you think?
I think it is unlikely that someone wants and can use such high or low Y coordinates?

TheMode: Seems good to me

PR #865, opened 2022-04-04
Add optional outgoing packet manipulation

erikzimmermann:
Sometimes it's necessary to either block or replace packets. With the current API it's possible to see when a packet is about to be sent but it could not be intercepted.

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerPacketOutEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerPacketOutEvent.java
@@ -1,26 +1,32 @@
 package net.minestom.server.event.player;
 
+import net.minestom.server.Viewable;
+import net.minestom.server.entity.Entity;
 import net.minestom.server.entity.Player;
-import net.minestom.server.event.trait.CancellableEvent;
 import net.minestom.server.event.trait.PlayerEvent;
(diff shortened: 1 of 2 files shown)


TheMode:
Ideally we would want the method to allow one packet being transformed into multiple.

Also worth reiterating that `PlayerPacketOutEvent` currently (and probably ever) does not support viewable packets. It is therefore a bit hard to rely on it.

CodingAir:
> Ideally we would want the method to allow one packet being transformed into multiple.
> 
> Also worth reiterating that `PlayerPacketOutEvent` currently (and probably ever) does not support viewable packets. It is therefore a bit hard to rely on it.

Thanks for your response. 

I changed the structure a bit to allow an appendix of packets instead of just one replacement.
Also, I updated the java doc for this event to inform other developers regarding the context of viewables.

outroddet:
Exactly what I need!

KrystilizeNevaDies:
Could this get updated? I'd like to look into getting it merged.

mworzala:
We have discussed this a bunch of times and avoided it for the following general reasons:
* Prevents some optimization on packet writing
* Generally the need for outgoing packet editing signifies a missing api

PR #891, opened 2022-04-07
Fix StairsPlacementRule

luzzardik:
This PR merges #443 adding cursor position to BlockPlacementRule, adding half property calculation through it, and fixing properties mapping.


luzzardik:
Found issues afterwards, fixing them before opening another PR :D

PR #895, opened 2022-04-09
Enhanced packet manipulation

outroddet:


TheMode:
Isn't it a duplicate of #865 ?

PR #900, opened 2022-04-09
Use eye height difference when setting direction using Entity#lookAt(Entity)

AlexDicy:
This change fixes the direction an Entity is looking at when using `Entity#lookAt(Entity)` and the two entities don't have the same eye height

List of changes in this PR:
- `Entity#lookAt(Entity)` now uses the entities eye height difference
- Added tests for `Entity#setView`, `Entity#lookAt(Pos)`, `Entity#lookAt(Entity)`
- `Entity#lookAt(Pos)` (and `#lookAt(Entity)`) do not update the entity's position if the target position is the same

Before PR - the wolf is looking at player's feet
image

After PR - the wolf is looking at player's eyes
image

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -348,7 +348,8 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
      */
     public void lookAt(@NotNull Entity entity) {
         Check.argCondition(entity.instance != instance, "Entity can look at another entity that is within it's own instance");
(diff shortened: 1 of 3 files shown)


TheMode:
Could you test it?

AlexDicy:
> Could you test it?

What do you mean?
(Just to make sure: I did test the code, last screenshot was taken with the updated code)

TheMode:
I mean writing tests, like the many present in https://github.com/Minestom/Minestom/tree/master/src/test/java/net/minestom/server/entity

PR #917, opened 2022-04-13
Fixed expanded bounding box

RinesThaix:
Currently players can't pickup items on half-blocks or staircases, because their expanded bounding box still starts at entity's bottom Y and not lower.


KermanIsPretty:
Should be already fixed from #1216, but this has a better fix is in #2614 that follows the same idea.

kermandev:
Seems to be fixed on latest release, as you can pickup blocks below you on a slab.

PR #921, opened 2022-04-14
Add onGround to PlayerMoveEvent

Bloepiloepi:
It is not possible to just use `Player#isOnGround()` because the field is updated after the event call.


TheMode:
I am not against, however it could quickly return outdated information if `newPosition` gets modified

PR #935, opened 2022-04-17
Item metadata fixes

Bloepiloepi:
This pull request fixes a few issues with item metadata and tags:
- Default values for primitive and always present metadata
- Nullable annotation for optional metadata
- Exception when reading a mapped tag with default value null

### DIFF
--- a/src/main/java/net/minestom/server/item/metadata/CrossbowMeta.java
+++ b/src/main/java/net/minestom/server/item/metadata/CrossbowMeta.java
@@ -12,7 +12,7 @@ import java.util.List;
 
 public record CrossbowMeta(TagReadable readable) implements ItemMetaView<CrossbowMeta.Builder> {
     private static final Tag<List<ItemStack>> PROJECTILES = Tag.ItemStack("ChargedProjectiles").list();
-    private static final Tag<Boolean> CHARGED = Tag.Boolean("Charged");
+    private static final Tag<Boolean> CHARGED = Tag.Boolean("Charged").defaultValue(false);
 
     public @NotNull List<ItemStack> getProjectiles() {
         return getTag(PROJECTILES);
--- a/src/main/java/net/minestom/server/item/metadata/CrossbowMeta.java
+++ b/src/main/java/net/minestom/server/item/metadata/CrossbowMeta.java
@@ -11,7 +11,7 @@ import org.jetbrains.annotations.UnknownNullability;
 import java.util.List;
 
 public record CrossbowMeta(TagReadable readable) implements ItemMetaView<CrossbowMeta.Builder> {
-    private static final Tag<List<ItemStack>> PROJECTILES = Tag.ItemStack("ChargedProjectiles").list();
+    private static final Tag<List<ItemStack>> PROJECTILES = Tag.ItemStack("ChargedProjectiles").list().defaultValue(List.of());
     private static final Tag<Boolean> CHARGED = Tag.Boolean("Charged").defaultValue(false);
 
     public @NotNull List<ItemStack> getProjectiles() {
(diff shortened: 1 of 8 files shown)


src/main/java/net/minestom/server/item/metadata/CrossbowMeta.java
```diff
@@ -12,7 +12,7 @@
 
 public record CrossbowMeta(TagReadable readable) implements ItemMetaView<CrossbowMeta.Builder> {
     private static final Tag<List<ItemStack>> PROJECTILES = Tag.ItemStack("ChargedProjectiles").list();
```
TheMode, line 14: This should also have a default. `List.of()`

src/main/java/net/minestom/server/item/metadata/WrittenBookMeta.java
```diff
@@ -23,7 +23,7 @@ public record WrittenBookMeta(TagReadable readable) implements ItemMetaView<Writ
                     textComponent -> LegacyComponentSerializer.legacySection().serialize(textComponent))
             .list().defaultValue(List.of());
 
-    public boolean isResolved() {
+    public @Nullable Boolean isResolved() {
```
TheMode, line 26: What about defaulting to `false`?

TheMode (changes requested):

src/main/java/net/minestom/server/item/metadata/WrittenBookMeta.java
Bloepiloepi: I agree that it would make more sense

src/main/java/net/minestom/server/item/metadata/FireworkMeta.java
```diff
@@ -17,11 +18,11 @@ public record FireworkMeta(TagReadable readable) implements ItemMetaView<Firewor
             .path("Fireworks").list().defaultValue(List.of());
     private static final Tag<Byte> FLIGHT_DURATION = Tag.Byte("Flight").path("Fireworks");
 
-    public List<FireworkEffect> getEffects() {
+    public @NotNull List<FireworkEffect> getEffects() {
         return getTag(EFFECTS);
     }
 
-    public byte getFlightDuration() {
+    public @Nullable Byte getFlightDuration() {
```
TheMode, line 25: Shouldn't this be defaulted to `0`? Should check vanilla.

src/main/java/net/minestom/server/item/metadata/PotionMeta.java
```diff
@@ -41,15 +41,15 @@ public void write(@NotNull TagWritable writer, @NotNull CustomPotionEffect value
...
         return getTag(POTION_TYPE);
     }
 
-    public List<CustomPotionEffect> getCustomPotionEffects() {
+    public @NotNull List<CustomPotionEffect> getCustomPotionEffects() {
         return getTag(CUSTOM_POTION_EFFECTS);
     }
 
-    public Color getColor() {
+    public @Nullable Color getColor() {
```
TheMode, line 52: Same. I think that we need parity with the client.

TheMode (changes requested):

src/main/java/net/minestom/server/item/metadata/FireworkMeta.java
Bloepiloepi: No, it is defaulted to null. The client will even display lore on the firework item if the flight duration is set

src/main/java/net/minestom/server/item/metadata/PotionMeta.java
Bloepiloepi: This is also supposed to be nullable, and if it is null, the client will calculate the color of the potion item based on the potion effects. If we default it every potion will be the same color.

TheMode:
LGTM. Just need to test the `Tag` default value change (probably in `TagTest`)

PR #943, opened 2022-04-18
Fix Entity#takeKnockback and gravity

Bloepiloepi:
This pull request fixes the `Entity#takeKnockback` method. Previously, blocks/tick and blocks/second were used interchangeably, and the tps was divided by 2. The method is now also slightly more readable.


Kebab11noel (approved):
The knockback calculation matches vanilla, however test values are a bit different compared to what i got from vanilla but the difference is negligible

PR #945, opened 2022-04-18
Add Pos#withLookAt

ZakShearman:
Add a withLookAt method to Pos based on code from Entity#lookAt

### DIFF
--- a/src/main/java/net/minestom/server/coordinate/Pos.java
+++ b/src/main/java/net/minestom/server/coordinate/Pos.java
@@ -120,10 +120,8 @@ public record Pos(double x, double y, double z, float yaw, float pitch) implemen
         }
 
         Vec delta = Vec.fromPoint(point.sub(this)).normalize();
-        return new Pos(x, y, z,
-                PositionUtils.getLookYaw(delta.x(), delta.z()),
-                PositionUtils.getLookPitch(delta.x(), delta.y(), delta.z())
-        );
+        return withView(PositionUtils.getLookYaw(delta.x(), delta.z()),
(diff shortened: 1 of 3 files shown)


ZakShearman:
Not sure if I got the unit test naming right and also, should the Entity lookAt unit tests be changed/removed since they are almost the same as the new Pos ones.

src/main/java/net/minestom/server/coordinate/Pos.java
```diff
@@ -112,6 +113,19 @@ public Pos(@NotNull Point point) {
         return new Pos(x, y, z, yaw, pitch);
     }
 
+    @Contract(pure = true)
+    public @NotNull Pos withLookAt(@NotNull Pos position) {
```
Marlonzockt, line 117: The parameter could be a Point instead of a Pos, couldn't it?

ZakShearman: Indeed, changed this now.

src/main/java/net/minestom/server/coordinate/Pos.java
```diff
@@ -112,6 +113,19 @@ public Pos(@NotNull Point point) {
...
     }
 
+    @Contract(pure = true)
+    public @NotNull Pos withLookAt(@NotNull Point point) {
+        if (this.samePoint(point)) {
+            return this;
+        }
+
+        Vec delta = Vec.fromPoint(point.sub(this)).normalize();
+        return new Pos(x, y, z,
```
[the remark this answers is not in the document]
ZakShearman: Thanks I've changed this 👍🏻

PR #960, opened 2022-04-21
Rename outdated LivingEntityMeta methods

### LINKED ISSUE #712: Miss named LivingEntityMeta methods
`setHealthAddedByAbsorption` and `getHealthAddedByAbsorption` are named incorrectly as these set and get the bee stinger count on the entity which displays a small bee stinger randomly on the player like the arrow. 

Methods in question: 
https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/entity/metadata/LivingEntityMeta.java#L83

Wiki vg link of what it is:
index 13 is the stinger one which corresponds to the 2 methods.
https://wiki.vg/Entity_metadata#Living_Entity

GreatWyrm:
Fixes #712 .

The absorption heart count was moved to PlayerMeta, and bee stinger count took the place of meta index 13, but these methods were never updated to match.

This PR renames the methods to solve the issue.

### DIFF
--- a/src/main/java/net/minestom/server/entity/metadata/LivingEntityMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/LivingEntityMeta.java
@@ -76,10 +76,40 @@ public class LivingEntityMeta extends EntityMeta {
         super.metadata.setIndex(OFFSET + 4, Metadata.VarInt(value));
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


WeiiswurstDev:
Is there anyone using the old method (names)? Should we deprecate them instead of removing them?

PR #961, opened 2022-04-21
Fix unknown repository check

VendoAU:
Predefined repositories weren't being checked properly, giving an error when using the correct repos.


mworzala:
This is why you write tests (me) >:(

PR #962, opened 2022-04-22
Fix #942

### LINKED ISSUE #942: Using setNotifyAboutChanges appears to ignore some entity meta
[Video for example](https://cdn.discordapp.com/attachments/706185253974442006/965626658323300432/2022-04-18_15-51-40.mp4)
[Code to reproduce](https://bin.bloom.host/uniroyesuz.swift)

In the video you can see when I type reset2 (the one without `setNotifyAboutChanges`) everything works fine, but when I do reset it does not send certain information at all (e.g. setGravity) or delays them until another action (e.g. setInvisible seemed to fix itself after I crouched)

WeiiswurstDev:
The bug was caused because the list of changes was cleared before it was sent to the players. It will now get cleared after it is sent.


WeiiswurstDev:
I tested the fix locally and the code for reproducing the bug does not reproduce it anymore. Feel free to test it yourself with the code in the original issue. For my testing, I simply overwrote the main method in the example (however, this is not part of the commit / the PR) with the test code.

Therefore, this PR fixes #942

TheMode:
Could you write a test case for it?

WeiiswurstDev:
Ok, will push it soon™

src/test/java/net/minestom/server/entity/EntityMetaTest.java
```diff
@@ -21,4 +30,76 @@ public void ensureRegistration() throws IllegalAccessException {
         }
         assertTrue(list.isEmpty(), "Missing meta for: " + list);
     }
+
+
+    @Test
+    public void notifyAboutChanges(Env env) {
```
TheMode, line 36: This should be put in a dedicated `EntityMetaIntegrationTest` file (and allow the changes above to be reverted)

WeiiswurstDev: Alright, will move it into a new test

WeiiswurstDev: Should I amend the commit?

TheMode: branch are squashed before merge anyway. so don't worry about the git history

WeiiswurstDev: ok

src/test/java/net/minestom/server/entity/EntityMetaTest.java
```diff
@@ -1,11 +1,18 @@
...
+import net.minestom.server.network.packet.server.play.EntityMetaDataPacket;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Consumer;
 
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
TheMode, line 15: Can you remove the line changes from this file?

TheMode: (There are still multiple line changes)

WeiiswurstDev: ok

src/test/java/net/minestom/server/entity/EntityMetaTest.java
```diff
@@ -21,4 +21,4 @@ public void ensureRegistration() throws IllegalAccessException {
         }
         assertTrue(list.isEmpty(), "Missing meta for: " + list);
     }
-}
+}
```
TheMode, line 24: Still a line change

TheMode:
Thanks for the fix & awesome test ;)

WeiiswurstDev:
no problem

PR #964, opened 2022-04-23
MultipleItemEvent

### LINKED ISSUE #950: InventoryPreClickEvent and PlayerSwapItemEvent don't extend ItemEvent
ItemEvent only provides one item.

But InventoryPreClickEvent and PlayerSwapItemEvent needs 2 Items so they don't extend ItemEvent. Understand.

But maybe add an MultiItemEvent or so with an List<ItemStack> and make it work with EventFilter.ITEM

DasLixou:
This adds support for `EventFilter.ITEM` for Events with more than one ItemStack.

before, even:

```java
EventNode.event("multipleItemEventTestlol", EventFilter.ITEM, itemEvent -> {
        return test(itemEvent.getItemStack());
});
```
wasn't possible when using multiple item events like PlayerSwapItemEvent.

now, with this Pull Request, this is possible:

```java
EventNode.event("multipleItemEventTestlol", EventFilter.ITEM, itemEvent -> {
        if(!(itemEvent instanceof MultipleItemEvent)) return test(itemEvent.getItemStack());
        boolean valid = false;
        for (ItemStack item : ((MultipleItemEvent)itemEvent).getItemStacks()) {
                if(test(item)) {
                        valid = true;
                        break;
                }
        }
        return valid;
});
```

This fixes #950 

Feedback welcome - Thanks 😃

### DIFF
--- a/src/main/java/net/minestom/server/event/inventory/InventoryClickEvent.java
+++ b/src/main/java/net/minestom/server/event/inventory/InventoryClickEvent.java
@@ -3,6 +3,7 @@ package net.minestom.server.event.inventory;
 import net.minestom.server.entity.Player;
 import net.minestom.server.event.trait.EntityInstanceEvent;
 import net.minestom.server.event.trait.InventoryEvent;
+import net.minestom.server.event.trait.MultipleItemEvent;
 import net.minestom.server.event.trait.PlayerEvent;
 import net.minestom.server.inventory.Inventory;
 import net.minestom.server.inventory.click.ClickType;
@@ -14,14 +15,13 @@ import org.jetbrains.annotations.Nullable;
  * Called after {@link InventoryPreClickEvent}, this event cannot be cancelled and items related to the click
  * are already moved.
  */
-public class InventoryClickEvent implements InventoryEvent, PlayerEvent, EntityInstanceEvent {
+public class InventoryClickEvent implements MultipleItemEvent, InventoryEvent, PlayerEvent, EntityInstanceEvent {
 
     private final Inventory inventory;
     private final Player player;
     private final int slot;
     private final ClickType clickType;
-    private final ItemStack clickedItem;
-    private final ItemStack cursorItem;
+    private final ItemStack[] itemStacks;
 
     public InventoryClickEvent(@Nullable Inventory inventory, @NotNull Player player,
                                int slot, @NotNull ClickType clickType,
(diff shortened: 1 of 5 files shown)


src/main/java/net/minestom/server/event/inventory/InventoryClickEvent.java
```diff
@@ -80,11 +79,14 @@ public ItemStack getClickedItem() {
...
     public ItemStack getCursorItem() {
-        return cursorItem;
+        return itemStacks[1];
     }
 
     @Override
     public @Nullable Inventory getInventory() {
         return inventory;
     }
+
```
WeiiswurstDev, line 89: Consider adding some javadoc here (size of the array, indexes of the contents)

src/main/java/net/minestom/server/event/inventory/InventoryItemChangeEvent.java
```diff
@@ -54,11 +53,14 @@ public int getSlot() {
...
     public @NotNull ItemStack getNewItem() {
-        return newItem;
+        return itemStacks[1];
     }
 
     @Override
     public @Nullable Inventory getInventory() {
         return inventory;
     }
+
```
WeiiswurstDev, line 63: Consider adding some javadoc here (size of the array, indexes of the contents)

src/main/java/net/minestom/server/event/inventory/InventoryClickEvent.java
WeiiswurstDev: Ok for some reason the preview does not show the right lines lol

WeiiswurstDev:
Generally getItemStacks() needs javadoc wherever it is implemented because it is unclear what is in that array

src/main/java/net/minestom/server/event/inventory/InventoryItemChangeEvent.java
DasLixou: Sure I’ll do that tomorrow. Thanks for the feedback 😄

DasLixou:
> Generally getItemStacks() needs javadoc wherever it is implemented because it is unclear what is in that array

Yeah I’ll make javadoc for them tomorrow :)

TheMode:
I am personally against the change. It does not seem to add much except confusion (also, arrays are very unsafe to expose like that since you cannot guarantee immutability)

WeiiswurstDev:
I think it would work better as a 'tag' interface of sorts. 

Since you have different values in the array every time anyways, they cant really be used without checking the type of event first. And if the type of the event has to be checked anyways, then you can just use the existing methods of the events.

tldr the only reason to use it is to cancel events with the filter, and we dont need to expose an additional method for that purpose.

PR #975, opened 2022-04-25
hasVelocity is now false if the entity just has gravity

### LINKED ISSUE #967: Entities always have some downward velocity
What the title says — stationary entities (standing on a block) have a velocity of [0, -1.568, 0]. Testing with a field watchpoint and logging the value showed that the `velocity` field is constantly (every tick) being set to this value by `velocityTick()`. 

Here's the code I used to spawn the entity:

```java
MinecraftServer.getGlobalEventHandler().addListener(PlayerChatEvent.class, event -> {
    String msg = event.getMessage();
    Player player = event.getPlayer();
    Instance instance = player.getInstance();

    if(instance != null) {
        switch (msg) {
            case "ZZZ" -> {
                EntityCreature creature = new EntityCreature(EntityType.ZOMBIE);
                creature.setInstance(instance, player.getPosition().add(5, 0, 0));
            }
        }
    }
});
```

I was using 765d6057da0bb9882ce31973e6f78be32f340c4d, though I'm technically running a minor fork that only implements a few (unrelated) changes. I tested on again on e2dac1c46ee01d99e50aab3deef99127424a9adb, right before the `takeKnockback` and gravity fixes, and velocity reads as expected. So I'm assuming it's something to do with how gravity is applied, though I don't know exactly what.

WeiiswurstDev:
Fixes #967 
I also added a new test case and modified an existing case to ensure that the code works, and updated the javadoc of hasVelocity.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -914,7 +914,7 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     public boolean hasVelocity() {
         if (isOnGround()) {
             // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
-            return !(velocity.x() == 0 && velocity.z() == 0 && velocity.y() < 0);
+            return velocity.x() != 0 || velocity.z() != 0 || velocity.y() > 0;
         } else {
             // The entity does not have velocity if the velocity is zero
             return !velocity.isZero();
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -914,7 +914,7 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     public boolean hasVelocity() {
         if (isOnGround()) {
             // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
-            return velocity.x() != 0 || velocity.z() != 0 || velocity.y() > 0;
+            return Double.compare(velocity.x(), 0) != 0 || Double.compare(velocity.z(), 0) != 0 || Double.compare(velocity.y(), 0) > 0;
         } else {
             // The entity does not have velocity if the velocity is zero
             return !velocity.isZero();
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -909,10 +909,11 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     /**
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0
+     * @return true if velocity is not set to 0 or the default velocity.
      */
     public boolean hasVelocity() {
-        return !velocity.isZero();
+        return !(velocity.isZero() ||
+                (velocity.x() == 0 && velocity.z() == 0 && Math.abs(velocity.y() + 1.568) < 0.001));
     }
 
     /**
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -914,7 +914,7 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     public boolean hasVelocity() {
         if (isOnGround()) {
             // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
-            return Double.compare(velocity.x(), 0) != 0 || Double.compare(velocity.z(), 0) != 0 || Double.compare(velocity.y(), 0) > 0;
+            return Double.compare(velocity.x(), 0) != 0 || Double.compare(velocity.z(), 0) != 0 || velocity.y() > 0;
         } else {
             // The entity does not have velocity if the velocity is zero
             return !velocity.isZero();
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -909,13 +909,16 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     /**
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0 or the default velocity.
+     * @return Only false if the velocity is zero or the entity is on the ground and only moves downwards (gravity).
      */
     public boolean hasVelocity() {
-        final double defaultYGravity = -gravityAcceleration * MinecraftServer.TICK_PER_SECOND * (1 - gravityDragPerTick);
-        return !(velocity.isZero() || // The entity does not have velocity if the velocity is zero
-                // or the entity does not have velocity if the velocity is equal to the default (gravity) velocity
-                velocity.samePoint(0, defaultYGravity, 0));
+        if (isOnGround()) {
+            // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
+            return !(velocity.x() == 0 && velocity.z() == 0 && velocity.y() < 0);
+        } else {
+            // The entity does not have velocity if the velocity is zero
+            return !velocity.isZero();
+        }
     }
 
     /**
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -912,9 +912,10 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
      * @return true if velocity is not set to 0 or the default velocity.
      */
     public boolean hasVelocity() {
+        final double defaultYGravity = -gravityAcceleration * MinecraftServer.TICK_PER_SECOND * (1 - gravityDragPerTick);
         return !(velocity.isZero() || // The entity does not have velocity if the velocity is zero
                 // or the entity does not have velocity if the velocity is equal to the default (gravity) velocity
-                (velocity.x() == 0 && velocity.z() == 0 && Math.abs(velocity.y() + 1.568) < 0.001));
+                velocity.samePoint(0, defaultYGravity, 0));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -909,10 +909,11 @@ public void setVelocity(@NotNull Vec velocity) {
     /**
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0
+     * @return true if velocity is not set to 0 or the default velocity.
      */
     public boolean hasVelocity() {
-        return !velocity.isZero();
+        return !(velocity.isZero() ||
+                (velocity.x() == 0 && velocity.z() == 0 && Math.abs(velocity.y() + 1.568) < 0.001));
```
TheMode, line 916: Could you explain this (both here and in comment)? Look very magical to me.

WeiiswurstDev: the entity does not have velocity if:
- the velocity is zero or
- the velocity is the default velocity that entities get from gravity (x=0=z, y=-1.568)

I invert that or statement to get the final result: Does the entity have velocity.
I'll add some comments to the statements as well

TheMode: I mean, how did you get this number?

WeiiswurstDev: I printed out the velocity of the entity

WeiiswurstDev: The number was also mentioned in the issue

TheMode: So it does not account for the gravity settings of the entity? it does not sound very robust

WeiiswurstDev: Wait, entities can have gravity settings?

TheMode: drag is indeed not a global constant

WeiiswurstDev: Ok, I'll try to find the formula for calculating the velocity from the gravity settings and use that.

WeiiswurstDev: https://minecraft.fandom.com/wiki/Entity#Motion_of_entities

None of the numbers here match what the code produces. Is our gravity calculation different?

Bloepiloepi: The y velocity when onGround is equal to `-gravityAcceleration * tps * (1 - gravityDragPerTick)`

WeiiswurstDev:
Should I also check if the entity is on ground?
Or is that unnecessary?

WeiiswurstDev:
That makes sense, I'll add the onground check

src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
```diff
@@ -102,6 +105,30 @@ public void doubleKnockback(Env env) {
...
+
+        entity.setInstance(instance, new Pos(0, 42, 0)).join();
+        env.tick();
+
+        // Should be true: The entity is currently falling (in the air), so it does have a velocity.
+        // Only entities on the ground should ignore the default velocity.
+        assertTrue(entity.hasVelocity());
+
+        // Test until the entity is on the ground
+        boolean tickedSuccessfully = env.tickWhile(() -> !entity.isOnGround(), Duration.ofMillis(50));
```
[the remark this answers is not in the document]
WeiiswurstDev: I'm ticking multiple times to test regular gravity. Setting the velocity is another test (but I can add that as well)

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -909,10 +909,16 @@ public void setVelocity(@NotNull Vec velocity) {
...
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0
+     * @return Only false if the velocity is zero or the entity is on the ground and only moves downwards (gravity).
      */
     public boolean hasVelocity() {
-        return !velocity.isZero();
+        if (isOnGround()) {
+            // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
+            return !(velocity.x() == 0 && velocity.z() == 0 && velocity.y() < 0);
```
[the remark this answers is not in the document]
WeiiswurstDev: In my opinion, if an entity has a positive velocity, hasVelocity should return true immediatly and not just after one tick. 

The smaller check you suggested would make it inconsistent (People could set a velocity, check hasVelocity in the same method and would get false because the server didnt tick yet)

src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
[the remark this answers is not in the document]
iam4722202468: gravity is already being tested in `void gravity(Env env)`. I also don't see how this is testing gravity, this is just checking if an entity will be on the ground after 50ms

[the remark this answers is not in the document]
WeiiswurstDev: Makes sense, I'll change it.

WeiiswurstDev:
Alright, I removed the gravity check. 

If there is anything else you still want changed before you merge just let me know :)

src/main/java/net/minestom/server/entity/Entity.java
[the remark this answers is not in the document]
iam4722202468: should be `return velocity.x() != 0 || velocity.z() != 0 || velocity.y() > 0;`

currently when x, y and z of velocity are zero returns true

src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
```diff
@@ -102,6 +105,27 @@ public void doubleKnockback(Env env) {
...
+    @Test
+    public void testHasVelocity(Env env) {
+        var instance = env.createFlatInstance();
+        loadChunks(instance);
+
+        var entity = new Entity(EntityType.ZOMBIE);
+        // Should  be false because the new entity should have no velocity
+        assertFalse(entity.hasVelocity());
+
+        entity.setInstance(instance, new Pos(0, 40.1, 0)).join();
```
iam4722202468, line 117: Tick once after setting the instance. Entity should actually have no velocity because it just spawned and hasn't been ticked. This is returning true because of my comment above.

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -909,10 +909,16 @@ public void setVelocity(@NotNull Vec velocity) {
     /**
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0
+     * @return Only false if the velocity is zero or the entity is on the ground and only moves downwards (gravity).
```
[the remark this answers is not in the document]
WeiiswurstDev: Yeah thats easier to understand.
I guess if someone wants a more precise answer they can always look at the code

src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
```diff
@@ -102,6 +105,30 @@ public void doubleKnockback(Env env) {
...
     }
 
+    @Test
+    public void testHasVelocity(Env env) {
+        var instance = env.createFlatInstance();
+        loadChunks(instance);
+
+        var entity = new Entity(EntityType.ZOMBIE);
+        // Should  be false because the new entity should have no velocity
+        assertFalse(entity.hasVelocity());
```
iam4722202468, line 115: Don't need this check anymore, since we're checking after the entity joins

WeiiswurstDev: Yeah I didnt look at all the other checks.
But I dont really see the problem here: If the check fits in two places, why not have it in both?

iam4722202468: technically there isn't an issue, but there also isn't a point in checking twice since within this test the value cannot change.

Having two checks here would be like testing if setting the instance of an entity also sets the entity velocity, which is not the goal of this test. It's always better to test specific events and actions, instead of writing one test to check multiple events.

WeiiswurstDev: makes sense, got rid of the check

iam4722202468:
@TheMode this pr looks good to me

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -909,10 +909,16 @@ public void setVelocity(@NotNull Vec velocity) {
...
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0
+     * @return true if the entity is moving
      */
     public boolean hasVelocity() {
-        return !velocity.isZero();
+        if (isOnGround()) {
+            // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
+            return velocity.x() != 0 || velocity.z() != 0 || velocity.y() > 0;
```
TheMode, line 917: You should use `Double.compare`

WeiiswurstDev:
I had some problems signing the last change, the last commit fixes this (it only adds one newline, so I gave it the ci skip flag)

src/main/java/net/minestom/server/entity/Entity.java
WeiiswurstDev: Changed it.

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -909,10 +909,16 @@ public void setVelocity(@NotNull Vec velocity) {
...
      * Gets if the entity currently has a velocity applied.
      *
-     * @return true if velocity is not set to 0
+     * @return true if the entity is moving
      */
     public boolean hasVelocity() {
-        return !velocity.isZero();
+        if (isOnGround()) {
+            // if the entity is on the ground and only "moves" downwards, it does not have a velocity.
+            return Double.compare(velocity.x(), 0) != 0 || Double.compare(velocity.z(), 0) != 0 || Double.compare(velocity.y(), 0) > 0;
```
TheMode, line 917: The last `compare` is unnecessary

PR #984, opened 2022-04-26
Make PermissionHandler independent instance in Entity / ConsoleSender

LeeGodSRC:
This allows users to create their own custom permission handler implementation by Entity#setPermissionHandler() and ConsoleSender#setPermissionHandler()
Allowing more flexibility for users to create their own permission system


TheMode:
We are aware that the current permission system is flawed, we are however not ready to change it before having a clear goal in mind. And I am not convinced that a forwarding interface is the only thing we can do.

LeeGodSRC:
Understood, IMO this change can improve a lot of flexibility without disadvantage
This allows users to build luckperms for minestom for example which is what I'm planning to do
Can this be a temporary solution for the permission system until the new solution comes?

PR #1005, opened 2022-04-29
Heavily reduce memory consumption of AnvilLoader

jglrxavpok:
/!\ Requires building and publishing Hephaistos master to local Maven for now

Hephaistos is currently being refactored to allow for reads/writes of chunk & chunk section data without the overhead of ChunkColumn or ChunkSection.

Memory difference ("old" -> this PR, "new" -> state of master branch)
memorycomp

CPU usage difference ("old" -> this PR, "new" -> state of master branch)
cpucomp

For now, CPU performance seems lower with 36% lost performance. That will be addressed later before merging this PR.


demo/src/main/java/net/minestom/demo/MainDemo.java
```diff
@@ -1,20 +1,27 @@
...
+import net.minestom.server.world.biomes.Biome;
+import net.minestom.server.world.biomes.BiomeManager;
 
 public class MainDemo {
 
     public static void main(String[] args) {
         // Initialization
         MinecraftServer minecraftServer = MinecraftServer.init();
 
+        MinecraftServer.getCommandManager().register(new GamemodeCommand());
```
jglrxavpok, line 23: Needs to be removed before merge.

settings.gradle.kts
```diff
@@ -5,6 +5,7 @@ dependencyResolutionManagement {
     repositories {
         maven("https://jitpack.io")
         mavenCentral()
+        mavenLocal()
```
jglrxavpok, line 8: Needs to be removed before merge.

jglrxavpok:
New profiling with IDEA 2022:

Memory usage down by 91%
memorycomp

CPU usage down by 86%
cpucomp

Protocol:
1. Launching the same map with 'master' and this branch
2. Connect with vanilla game
3. Wait 10s for chunks to load
4. Stop server

jglrxavpok:
When saving:
CPU usage down by ~95% (saving is noticeably faster)
image

Memory usage down by 79%
image

TheMode (approved):

PR #1047, opened 2022-05-09
Fix entities falling through tall blocks

### LINKED ISSUE #990: Entities can fall through fences and walls
Tall blocks appear to have no collision when an entity attempts to pass through from above. Should be pretty easy to reproduce; just spawn any kind of entity on top of any number of fences/walls. It should fall all the way through. Tested on commit c8ae194f7d2eb4508df1aa29623d92100b22f809.

As far as I can tell, this occurs because `CollisionUtils#handlePhysics(Entity, Vec, PhysicsResult)` (as called in `Entity#velocityTick()`) is "assuming" that block collision shapes fit within a 1x1x1 cube, and is choosing candidates for detailed collision checks based on that assumption. This causes it to miss unusually tall blocks, like walls and fences.

I would try to fix this myself but I don't understand the collision code well enough to feel confident doing so.

Steanky:
Previously, entities could fall through tall blocks instead of standing on them, if their gravitational velocity was sufficiently close to zero. This was because the collision algorithm only sampled blocks in the range traversed by an entity as it's moving. Since fences and walls extend past a 1x1x1 cube, they would not be sampled unless the velocity had a magnitude greater than 0.5 (but even then, still could fail to be picked up on depending on entity positioning and velocity!)

The solution is to sometimes collision check against the bounds of the block beneath the one intersecting the moving bounding box, if the lower block has a height greater than 1. 

Caveat: The solution assumes there are no collision shapes greater than 1.5 blocks in height. 

Should fix #990, and a related problem with horizontal movement where the tall part of the fence would not be taken into account.

Also includes relevant integration tests.


iam4722202468:
Could you add another check similar to `entityPhysicsCheckFallCarpetFence`, where the entity hits both the fence and the carpet?

```
        entity.setInstance(instance, new Pos(0.5, 54.0625, 0.5)).join();
        assertEquals(instance, entity.getInstance());

        PhysicsResult res = CollisionUtils.handlePhysics(entity, new Vec(0, -11.03, 0));
        assertEqualsPoint(new Pos(0.5, 43.5, 0.5), res.newPosition());```

**Comment by Steanky** [comment]
I made the suggested changes, and also took the opportunity to add in an optimization + a few extra unit tests.

File: src/main/java/net/minestom/server/collision/BlockCollision.java [hunk 1]
```diff
@@ -375,12 +375,63 @@ static boolean checkBoundingBox(int blockX, int blockY, int blockZ,
...
+            pos = entityPosition.z();
+            block = blockZ;
+        }
+        else {
+            // if moving straight up, don't bother checking for tall solids beneath anything
+            // if moving straight down, only check for a tall solid underneath the last block
+            return yVelocity < 0 && blockY == Math.floor(entityPosition.y() + yVelocity);
+        }
+
+        return (entityVelocity.y() / v) * (block - pos) + entityPosition.y() >= blockY;
```
iam4722202468, line 435: I have no idea what you're doing here, could you put a comment

iam4722202468:
Your optimization causes this test to fail

```    @Test
    public void entityPhysicsCheckFallHitFenceLongMove(Env env) {
        var instance = env.createFlatInstance();
        instance.setBlock(0, 42, 0, Block.OAK_FENCE);
        instance.setBlock(0, 43, 0, Block.BROWN_CARPET);

        var entity = new Entity(EntityType.ZOMBIE);
        entity.setInstance(instance, new Pos(0.5, 54.0625, 0.5)).join();
        assertEquals(instance, entity.getInstance());

        PhysicsResult res = CollisionUtils.handlePhysics(entity, new Vec(0, -21.03, 0));
        assertEqualsPoint(new Pos(0.5, 43.5, 0.5), res.newPosition());
    }
```

From what I can understand, your optimization relies on the entity either moving a small amount or ending at the block with the fence, and these cases are not guaranteed.

Steanky:
It looks like the collider is hitting the carpet first, and then breaking out of the loop in RayUtils#raycastCollision. If there were no collision it'd eventually hit the fence.

I think I've got a relatively simple fix for this

Steanky:
Your test passes now, and I realized I could optimize the algorithm further (before, it would sometimes check blocks it didn't need to check).

src/main/java/net/minestom/server/collision/BlockCollision.java
```diff
@@ -375,12 +375,102 @@ static boolean checkBoundingBox(int blockX, int blockY, int blockZ,
...
+
+            final Vec currentPos = new Vec(blockX, blockY, blockZ);
+            if(belowShape.relativeEnd().y() > 1) {
+                // we should always check both shapes, so no short-circuit here, to handle cases where the bounding box
+                // hits the current solid but misses the tall solid
+                return belowShape.intersectBoxSwept(entityPosition, entityVelocity, belowPos, boundingBox, finalResult)
+                        | (currentCollidable && currentShape.intersectBoxSwept(entityPosition, entityVelocity,
+                        currentPos, boundingBox, finalResult));
+            }
+            else {
```
TheMode, line 399: Could you inline `else` expressions? It is very confusing to read. Would also suggest using the `.editorconfig` file present in the repo root.

Steanky: Done on latest commit. I was using the .editorconfig from the beginning, but I manually set the line endings to CRLF before pushing since that's what the original files appeared to be using. BlockCollision should be on LF now.

emortaldev:
I don't know if it's related, but falling sand blocks can fall through a solid block completely. I think it happens when the falling sand is 2 blocks above a solid block

Steanky:
> I don't know if it's related, but falling sand blocks can fall through a solid block completely. I think it happens when the falling sand is 2 blocks above a solid block

Probably unrelated unless it only happens with tall blocks. I tried spawning a bunch of falling blocks at different heights and was unable to get any to fall through (on a build with my fence collision fix). On an older build, the fence issue does impact falling blocks too.

PR #1049, opened 2022-05-10
Some tweaks to AbstractInventory/PlayerInventory, as well as various Tests

Moulberry:
Some tweaks to AbstractInventory/PlayerInventory, as well as various
tests to ensure the validity of these changes

- When calling Inventory#setItemStack, packets are not sent if previous
  == newItem
- When clearing an Inventory, only WindowItem & EntityEquipment packets
  are sent. Previously a large number of SetSlot packets were being sent
- Fixed bug with setCursorItem that would result in a packet *not* being
  sent if the item count changes


src/main/java/net/minestom/server/inventory/PlayerInventory.java
```diff
@@ -128,17 +127,16 @@ public void update() {
      * @param cursorItem the new cursor item
      */
     public void setCursorItem(@NotNull ItemStack cursorItem) {
-        final boolean similar = this.cursorItem.isSimilar(cursorItem);
+        if (this.cursorItem == cursorItem) return;
```
TheMode, line 130: Should use `equals`

src/main/java/net/minestom/server/inventory/AbstractInventory.java
```diff
@@ -75,7 +75,8 @@ protected final void safeItemInsert(int slot, @NotNull ItemStack itemStack) {
                     slot
             );
             previous = itemStacks[slot];
-            UNSAFE_itemInsert(slot, itemStack);
+            if (previous == itemStack) return; // Avoid sending updates if the item has not changed
```
TheMode, line 78: Same, `equals`

TheMode (changes requested):
Nitpick with item comparison, but otherwise LGTM

TheMode (approved):

PR #1051, opened 2022-05-10
Automatically generate dye colors.

Articdive:


code-generators/src/main/java/net/minestom/codegen/color/DyeColorGenerator.java
```diff
@@ -0,0 +1,152 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.StreamSupport;
+
+public class DyeColorGenerator extends MinestomCodeGenerator {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FluidGenerator.class);
```
Braayy, line 25: Shouldn't this be DyeColorGenerator.class instead of FluidGenerator.class?

Cooldude1235:
crfrfr

PR #1091, opened 2022-05-18
A simple Completer for MinestomTerminal & Fix LineReader visual issue

HSGamer:
image

Need testing & reviewing before merging


src/main/java/net/minestom/server/terminal/MinestomCompleter.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minestom.server.command.builder.suggestion.SuggestionEntry;
+import net.minestom.server.listener.TabCompleteListener;
+import org.jline.reader.Candidate;
+import org.jline.reader.Completer;
+import org.jline.reader.LineReader;
+import org.jline.reader.ParsedLine;
+
+import java.util.List;
+
+public class MinestomCompleter implements Completer {
```
TheMode, line 14: You can put this as an inner class in `MinestomTerminal`, there is also no reason to expose it

src/main/java/net/minestom/server/listener/TabCompleteListener.java
```diff
@@ -15,13 +16,27 @@
...
+                                suggestion.getStart(),
+                                suggestion.getLength(),
+                                suggestion.getEntries().stream()
+                                        .map(suggestionEntry -> new TabCompletePacket.Match(suggestionEntry.getEntry(), suggestionEntry.getTooltip()))
+                                        .toList())
+                        )
+                );
+    }
 
+    public static Optional<Suggestion> getSuggestion(CommandSender commandSender, String text) {
```
TheMode, line 39: I'd personally just return a nullable suggestion. Not much reason to deal with the overhead of `Optional` (+lambda) here since it is purely internal

HSGamer:
I've created a custom `Writer` to fix an issue with `LineReader` being overridden when other threads log messages.

https://user-images.githubusercontent.com/25300232/169654680-a051075e-ff62-4c22-9c48-824b5301080d.mp4

Should be enough to review and then merge @TheMode

src/main/java/net/minestom/server/terminal/MinestomConsoleWriter.java
```diff
@@ -0,0 +1,31 @@
+package net.minestom.server.terminal;
+
+import java.util.Map;
+
+import org.tinylog.core.LogEntry;
+import org.tinylog.writers.AbstractFormatPatternWriter;
+
+public final class MinestomConsoleWriter extends AbstractFormatPatternWriter {
```
TheMode, line 8: Again no reason to expose it, could be in `MinestomTerminal`

src/main/java/net/minestom/server/terminal/MinestomTerminal.java
```diff
@@ -58,7 +68,39 @@ public static void stop() {
             } catch (IOException e) {
                 e.printStackTrace();
             }
+            reader = null;
         }
     }
 
+    @Nullable
+    public static LineReader getReader() {
```
TheMode, line 76: What is this getter for?

TheMode (changes requested):

src/main/java/net/minestom/server/terminal/MinestomConsoleWriter.java
HSGamer: > Again no reason to expose it, could be in `MinestomTerminal`

@TheMode 
It does need exposing.
It's not for `MinestomTerminal`, but for the tinylog's `Writer` that is defined in the `META-INF/services`, which is higher than `MinestomTerminal`.

src/main/java/net/minestom/server/terminal/MinestomTerminal.java
HSGamer: @TheMode 
For `MinestomConsoleWriter`.
I explained in the above review.

src/main/java/net/minestom/server/terminal/MinestomConsoleWriter.java
HSGamer: It's a hacky workaround.
I didn't find any other way to fix the `LineReader` issue.
This is the best that I could do.
Maybe one day another solid fix can be used instead.

src/main/java/net/minestom/server/terminal/MinestomTerminal.java
HSGamer: Switched to `package-private`

PR #1095, opened 2022-05-20
Added support for setting blocks without updating.

DasLixou:
This can be used for example a clearer `Batch`, because the current batches just sets the blocks in the chunk and then resends the chunk, but when the blocks are getting setted, things like nearby block updates are ignored.

### DIFF
--- a/src/main/java/net/minestom/server/instance/InstanceContainer.java
+++ b/src/main/java/net/minestom/server/instance/InstanceContainer.java
@@ -89,6 +89,10 @@ public class InstanceContainer extends Instance {
 
     @Override
     public void setBlock(int x, int y, int z, @NotNull Block block) {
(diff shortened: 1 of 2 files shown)


TheMode:
The solution to this problem is definitely not to create a new `setBlock` method with one additional parameter. The batching API require some work, and we will have to move the block update logic somewhere else.

DasLixou:
Yeah but with this pr its also handy for other people that work not inside minestom

TheMode:
It is why I mentioned moving the update logic. I do not want to end up with a new `setBlock` method each time we think about a new use case

DasLixou:
So what would you recommend? The reason i made this pr is because I (and maybe others) need a way to set blocks without sending an update. Ok I understand new method is not the best idea but what else?

PR #1099, opened 2022-05-20
Use predicate for entity targeting

johnnywoof:
Use a predicate for checking if an entity can be targeted instead of using class types. This allows for greater flexibility in deciding the closet entity to target.

- `range` uses double instead of float
- `ClosestEntityTarget(@NotNull EntityCreature entityCreature, float range, @NotNull Class<? extends LivingEntity>... entitiesTarget)` constructor is deprecated in favor of new constructor with predicate
- Added unit test


KrystilizeNevaDies (approved):
LGTM
A predicate is a nice improvement, the unit tests are great too <3

PR #1111, opened 2022-05-22
Account for flying in velocity calculation

TogAr2:
It's a bit unclean to do `instanceof Player` but I think it's better than extending the method in `Player`, which results in duplicate code


TheMode:
Could you add a test?

ZakShearman:
Appears to still have weird behaviour where when flying up from the ground, your velocity starts negative and gets closer to 0. After a short time, it then goes back to -8.58 again whilst flying upwards. https://youtu.be/XVXZGDsKPHo

PR #1114, opened 2022-05-23
Added Facing to BlockFace

DasLixou:
This adds Facing property to blockface


TheMode:
The current block property API only work with string on purpose. A future auto-generated class may come later for static analysis/compile-time check.

PR #1119, opened 2022-05-24
Change assert in NamespacedID.from() to throw IllegalArgumentException instead

### LINKED ISSUE #1097: Illegal characters allowed in namespaces
When registering stuff, Minestom doesn't throw an exception if the namespace you provide contains invalid characters. This shows an error on a vanilla client and a cryptic "failed to join" on Fabric / Quilt.

To reproduce:
1. register a dimension with capital letters in the name
2. try to join

What happens:
you get an error message 

What should happen:
minestom throws an exception at step 1

GreatWyrm:
Fixes #1097.

Java programs have assertions turned off by default, and therefore you can register invalid namespaces with Minestom if you do not have assertions on.

This PR fixes it by replacing the assertions with exceptions instead.

### DIFF
--- a/src/main/java/net/minestom/server/utils/NamespaceID.java
+++ b/src/main/java/net/minestom/server/utils/NamespaceID.java
@@ -49,9 +49,15 @@ public final class NamespaceID implements CharSequence, Key {
         this.full = full;
         this.domain = domain;
         this.path = path;
-        assert !domain.contains(".") && !domain.contains("/") : "Domain cannot contain a dot nor a slash character (" + full + ")";
-        assert domain.matches(legalLetters) : "Illegal character in domain (" + full + "). Must match " + legalLetters;
-        assert path.matches(legalPathLetters) : "Illegal character in path (" + full + "). Must match " + legalPathLetters;
+        if(domain.contains(".") || domain.contains("/")) {
+            throw new IllegalArgumentException("Domain cannot contain a dot nor a slash character (" + full + ")");
+        }
+        if(!domain.matches(legalLetters)) {
+            throw new IllegalArgumentException("Illegal character in domain (" + full + "). Must match " + legalLetters);
+        }
+        if(!path.matches(legalPathLetters)) {
+            throw new IllegalArgumentException("Illegal character in path (" + full + "). Must match " + legalPathLetters);
+        }
     }
 
     public @NotNull String domain() {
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/utils/NamespaceID.java
```diff
@@ -49,9 +49,15 @@ private NamespaceID(String full, String domain, String path) {
         this.full = full;
         this.domain = domain;
         this.path = path;
-        assert !domain.contains(".") && !domain.contains("/") : "Domain cannot contain a dot nor a slash character (" + full + ")";
-        assert domain.matches(legalLetters) : "Illegal character in domain (" + full + "). Must match " + legalLetters;
-        assert path.matches(legalPathLetters) : "Illegal character in path (" + full + "). Must match " + legalPathLetters;
+        if(domain.contains(".") || domain.contains("/")) {
+            throw new IllegalArgumentException("Domain cannot contain a dot nor a slash character (" + full + ")");
+        }
+        if(!domain.matches(legalLetters)) {
```
TheMode, line 55: regex checks are fairly expensive, it is the original reason why these were assertions

GreatWyrm: True, but simply never checking means you can register illegal names.

I could try to split it up into different checks and avoid using regex, would that be what's needed?

MeFisto94:
Re: 
>  Java programs have assertions turned off by default
> it is the original reason why these were assertions

I also think that it depends a lot on the context, if we're witnessing a misunderstanding about Assertions or not (but I also don't know Minestom's Codebase enough):
Assertions are off by default so that they show now runtime overhead, but assert semantic failures in development builds (e.g. similar to how other languages have Debug and Run configurations)

It now depends on the context of this failure on whether we should rather push Assertions and basically build some documentation, showing how to run the server with assertions enabled (and to use that on QA/Testing environments, in case you don't run the server from the IDE).

Since using an invalid namespace does not cause any major fuck-up in unrelated parts (??), one could argue that an assert is still the right thing. You'll also immediately get feedback by the client when using such a namespaced key illegally.
But then, those IDs are probably cached, so the code path is not too hot and Minestom doesn't seem to use asserts very widely anyway.

Pick your Poison: Learning Curve of running with `-ea` or a runtime overhead in code bases that do it right.

src/main/java/net/minestom/server/utils/NamespaceID.java
TheMeinerLP: Pre Compile with `Pattern#compile` the regex and the regex get less expensive

TheMeinerLP (changes requested):

TheMeinerLP:
Also update the PR

GreatWyrm:
Closing, as the NamespaceID class is planned for removal in a future Minestom version

PR #1163, opened 2022-06-13
update adventure & use ComponentLogger

HSGamer:
Nothing special here.
Just switch the Logger to ComponentLogger so that we can send Component directly and/or allow modifying loggers via META-INF services (with color support??)


Xstoudi:
lgtm

PR #1172, opened 2022-06-18
Implement chat changes

Kebab11noel:


src/main/java/net/minestom/server/network/packet/server/play/SetChatPreviewPacket.java
```diff
@@ -7,6 +7,9 @@
 import org.jetbrains.annotations.NotNull;
 
 public record SetChatPreviewPacket(boolean enable) implements ServerPacket {
+    public static final SetChatPreviewPacket ENABLE = new SetChatPreviewPacket(true);
+    public static final SetChatPreviewPacket DISABLE = new SetChatPreviewPacket(true);
```
GreatWyrm, line 11: Should this be false?

Kebab11noel: Oops.. Will fix it in the next commit

src/main/java/net/minestom/server/adventure/audience/PacketGroupingAudience.java
```diff
@@ -58,7 +60,11 @@ default void sendGroupedPacket(@NotNull ServerPacket packet) {
 
     @Override
     default void sendMessage(@NotNull Identity source, @NotNull Component message, @NotNull MessageType type) {
-        Messenger.sendMessage(this.getPlayers(), message, ChatPosition.fromMessageType(type), source.uuid());
+        if (type == MessageType.SYSTEM)
+            Messenger.sendSystemMessage(getPlayers(), message, ChatType.SYSTEM);
+        else
+            Messenger.sendUnsignedMessage(getPlayers(), new MessageSender(Component.text("UNKNOWN SENDER",
+                    NamedTextColor.RED), null), message);
```
BomBardyGamer, line 67: This should be bracketed imo

src/main/java/net/minestom/server/command/ArgumentsSignature.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public record ArgumentsSignature(long timestamp, long salt, Map<String, byte[]> signatures, boolean signedPreview)
+implements Writeable {
+    public ArgumentsSignature(long timestamp, long salt, Map<String, byte[]> signatures, boolean signedPreview) {
```
BomBardyGamer, line 20: Can't you just remove all those parameters and only have `this.signatures = Map.copyOf(signatures)`?

src/main/java/net/minestom/server/crypto/SignatureValidator.java
```diff
@@ -0,0 +1,94 @@
...
+import net.kyori.adventure.text.serializer.gson.GsonComponentSerializer;
+import net.minestom.server.entity.Player;
+import net.minestom.server.utils.crypto.KeyUtils;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
+import java.security.*;
```
BomBardyGamer, line 13: Star imports no good

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -663,16 +667,24 @@ public void sendPluginMessage(@NotNull String channel, @NotNull String message)
 
     @Override
     public void sendMessage(@NotNull Identity source, @NotNull Component message, @NotNull MessageType type) {
-        Messenger.sendMessage(this, message, ChatPosition.fromMessageType(type), source.uuid());
+        if (type == MessageType.SYSTEM)
+            Messenger.sendSystemMessage(List.of(this), message, ChatType.SYSTEM);
+        else
+            Messenger.sendUnsignedMessage(List.of(this), new MessageSender(Component.text("UNKNOWN SENDER",
+                    NamedTextColor.RED), null), message);
```
BomBardyGamer, line 674: Brackets

src/main/java/net/minestom/server/command/ArgumentsSignature.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public record ArgumentsSignature(long timestamp, long salt, Map<String, byte[]> signatures, boolean signedPreview)
+implements Writeable {
```
BomBardyGamer, line 19: Is this correct wrapping?

src/main/java/net/minestom/server/utils/crypto/KeyUtils.java
```diff
@@ -0,0 +1,41 @@
...
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Base64;
+
+public final class KeyUtils {
+    private static final Base64.Encoder MIME_ENCODER = Base64.getMimeEncoder(76, "\n".getBytes(StandardCharsets.UTF_8));
+    private static final String RSA_HEADER = "-----BEGIN RSA PUBLIC KEY-----\n";
+    private static final String RSA_FOOTER = "\n-----END RSA PUBLIC KEY-----\n";
+
+    public enum SignatureAlgorithm {SHA256withRSA,SHA1withRSA}
+    public enum KeyAlgorithm {RSA}
```
BomBardyGamer, line 17: Okay... no.

BomBardyGamer (commented):
This just contains nitpicks and styling issues. Other than that, looks alright to me, though I can't say it's good, since I didn't fully look at it.

Good PR though :)

src/main/java/net/minestom/server/adventure/audience/PacketGroupingAudience.java
Kebab11noel: Yeah it might look better, and iirc there is another one of this somewhere, i'll change that too

src/main/java/net/minestom/server/command/ArgumentsSignature.java
Kebab11noel: Yes, Mode already mentioned this, but i forgot to change it.

src/main/java/net/minestom/server/command/ArgumentsSignature.java
Kebab11noel: Well `implements` was on the vertical guide line, should i keep it as a long line or wrap it in the middle of the args?

src/main/java/net/minestom/server/utils/crypto/KeyUtils.java
Kebab11noel: I don't see the problem here... :D

src/main/java/net/minestom/server/crypto/SignatureValidator.java
Kebab11noel: What's wrong with them? IntelliJ said they are fine.

src/main/java/net/minestom/server/command/ArgumentsSignature.java
BomBardyGamer: I only ask because there's another instance in here where `implements` is wrapped and aligned with an 8 space indent, which means this is inconsistent. This is why I usually avoid wrapping on things like `implements`, though unfortunately, there isn't really much else you can wrap on here, except the parameters.

src/main/java/net/minestom/server/crypto/SignatureValidator.java
BomBardyGamer: The problem with star imports is that they import everything from that package. Therefore, not only is it difficult to see exactly what's being used, but it's also annoying if you want to use something else that's got the same name as one of these things. There are other reasons that I can't really recall that are better than this, but it's also [against Google's style guide](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports) too.

BomBardyGamer: Also, there is another instance in here when IntelliJ corrected a star import to a separated import, which again makes this inconsistent.

src/main/java/net/minestom/server/Configuration.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public static class Builder {
+        private boolean requireValidPlayerPublicKey = false;
+        private Component missingPlayerPublicKeyMessage = Component.text("Missing public key!", NamedTextColor.RED);
+        private Component invalidPlayerPublicKeyMessage = Component.text("Invalid public key!", NamedTextColor.RED);
+        private ChatTypeBuilder playerChatType = ChatTypeBuilder.builder(ChatType.CHAT.key()).chat(ChatDecoration.contentWithSender("chat.type.text"));
+        private ChatTypeBuilder systemChatType = ChatTypeBuilder.builder(ChatType.SYSTEM.key()).chat();
+
+        public Builder setRequireValidPlayerPublicKey(boolean requireValidPlayerPublicKey) {
```
BomBardyGamer, line 24: Don't most of the other builders in Minestom omit the `set` prefix?

Kebab11noel: I'll change that, and this config change will be made in master first

src/main/java/net/minestom/server/utils/FinalObject.java
```diff
@@ -0,0 +1,48 @@
+package net.minestom.server.utils;
+
+import java.util.function.Supplier;
+
+/**
+ * Used to mimic the final keyword
```
BomBardyGamer, line 6: Well that's not really true, it's more of a lazy initialized object than a final object, i.e. it can only be set once, and can be set lazily.

src/main/java/net/minestom/server/utils/crypto/KeyUtils.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final String RSA_FOOTER = "\n-----END RSA PUBLIC KEY-----\n";
+
+    public enum SignatureAlgorithm {
+        SHA256withRSA,
+        SHA1withRSA
+    }
+
+    public enum KeyAlgorithm {
+        RSA
+    }
```
BomBardyGamer, line 23: Thinking about it, these enums don't really even need to exist. These are only used in the string constant context, i.e. by calling `name()` all the time, so they could just be string constants instead of enums.

Kebab11noel: With the enum you can be sure that the method will exists, but with a string you might accidentally pass a wrong value.

BomBardyGamer:
I think this PR should wait until Adventure 4.12.0 is released, as I believe that quite a few things will be implemented in that, and so not required here. We don't want to push an API that will later be rewritten. I think this should be blocked by KyoriPowered/adventure#777

src/main/java/net/minestom/server/utils/crypto/KeyUtils.java
BomBardyGamer: That is true, but is there even even that anyone other than us internally could pass an incorrect value? Because if there is not (and I'm pretty sure there isn't), we don't need the enum to guard against that.

src/main/java/net/minestom/server/utils/ObjectCache.java
```diff
@@ -0,0 +1,28 @@
+package net.minestom.server.utils;
+
+import java.util.function.Supplier;
+
+/**
+ * Used to cache objects
+ *
+ * @param <T> type of the object
+ */
+public final class ObjectCache<T> {
```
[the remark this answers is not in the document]
BomBardyGamer: Yeah that's fair enough. I recommend something like `MemoizingSupplier`.

src/main/java/net/minestom/server/utils/FinalObject.java
BomBardyGamer: Can I recommend a new name like `LazyInitializedObject` or `LazyLoadedValue` for this class?

Kebab11noel: We try to avoid this object, so it's name probably won't matter, but if it ends up in the final pr i'll choose a better name.

MenschenToaster:
> I think this PR should wait until Adventure 4.12.0 is released, as I believe that quite a few things will be implemented in that, and so not required here. We don't want to push an API that will later be rewritten. I think this should be blocked by [KyoriPowered/adventure#777](https://github.com/KyoriPowered/adventure/pull/777)

We now have adventure 4.12.0. When will we have this implemented?

PR #1178, opened 2022-06-21
Configuration

Kebab11noel:
Make things configurable. Feel free to comment values I missed.


src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -202,89 +216,44 @@ public static boolean isStopping() {
...
-        Check.stateCondition(serverProcess.isAlive(), "The compression threshold cannot be changed after the server has been started.");
-        MinecraftServer.compressionThreshold = compressionThreshold;
+        return configuration().compressionThreshold();
     }
 
     /**
      * Gets if the built in Minestom terminal is enabled.
      *
      * @return true if the terminal is enabled
+     * @deprecated use {@link }
```
BomBardyGamer, line 257: lol

src/main/java/net/minestom/server/network/packet/client/login/LoginStartPacket.java
```diff
@@ -59,7 +57,7 @@ public void process(@NotNull PlayerConnection connection) {
             byte[] nonce = new byte[4];
             ThreadLocalRandom.current().nextBytes(nonce);
             socketConnection.setNonce(nonce);
-            socketConnection.sendPacket(new EncryptionRequestPacket("", publicKey, nonce));
+            socketConnection.sendPacket(new EncryptionRequestPacket(MinecraftServer.configuration().serverId(), publicKey, nonce));
```
BomBardyGamer, line 60: Pretty sure that server ID is meant to be empty.

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -85,6 +85,11 @@ public static MinecraftServer init(Configuration configuration) {
         return new MinecraftServer();
     }
 
+    @ApiStatus.Internal
+    public static ServerProcess updateProcess() {
```
BomBardyGamer, line 89: It's an internal method. Surely breaking changes don't matter at all.

src/main/java/net/minestom/server/network/packet/client/login/LoginStartPacket.java
Kebab11noel: Well it's sent to the session server as an SHA-1 hash in the form of serverID + serverPubKey + clientPrivKey

src/main/java/net/minestom/server/MinecraftServer.java
Kebab11noel: It's definitelly not because I was too lazy to write the annotation and javadoc for everything and just copy-pasted the template and forgot to fill in the detail...

src/main/java/net/minestom/server/MinecraftServer.java
Kebab11noel: Tests probably doesn't care about some messages and values, so you can see this as a convenience method.

src/main/java/net/minestom/server/network/packet/client/login/LoginStartPacket.java
BomBardyGamer: Yeah I know, the only reason I pointed this out is because vanilla uses a blank server ID, and it's standard to use a blank server ID

Kebab11noel: I believe it used to be configurable in vanilla at some point, anyways i'm not sure what to do with this, since it has no real effect and having one more field in the config doesn't hurt either.

PR #1181, opened 2022-06-23
Additional methods to Player class to work better with events

maximbreitman:


TheMode:
These methods are already available in `EquipmentHandler`: https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/inventory/EquipmentHandler.java#L52

maximbreitman:
Oops. Didn't notice.

PR #1188, opened 2022-06-30
process chunk history based on actual history size

### LINKED ISSUE #1187: Stuck on Loading Terrain...
The spawn instance was created from the vanilla Minecraft client. 

```java
public class Main {
    @SuppressWarnings("UnstableApiUsage")
    public static void main(String[] args) {
        MinecraftServer server = MinecraftServer.init();

        EventNode<Event> node = MinecraftServer.getGlobalEventHandler();
        node.addListener(PlayerLoginEvent.class, event -> {
            IChunkLoader loader = new AnvilLoader("/path/to/my/vanilla/world");
            InstanceContainer instance = MinecraftServer.getInstanceManager().createInstanceContainer(loader);
            instance.setChunkSupplier(DynamicChunk::new);
            Pos pos = new Pos(164F, 68F, -504F); // should be arbitrary

            event.getPlayer().setRespawnPoint(pos);
        });
        node.addListener(PlayerSpawnEvent.class, event -> {
            Pos pos = new Pos(0, 100, 0);
            event.getPlayer().teleport(pos).thenRun(() -> {
                event.getPlayer().setGameMode(GameMode.CREATIVE);
                event.getPlayer().setFlying(true);
                Pos start = pos.sub(0, 1, 0);

                for(int i = 0; i < 100; i++) {
                    for(int j = 0; j < 100; j++) {
                        event.getSpawnInstance().setBlock(start.add(i, 0, j), Block.GOLD_BLOCK);
                    }
                }
            });
        });
    }
}
```

The first time you join, you get teleported, but no gold blocks are visible.
The second time, the game hangs on Loading Terrain
I couldn't get into this world with larger code starting from bcab1b1, but not 100% sure if that's the root of the issue.

tahmid-23:
seems to fix #1187
the limit checker was zeroed initially, so it refused to send chunks with zero chunk key
in the integration test, this was mistaken for the chunk having already been added to the history (which it wasn't, so I now send it when the player spawns)

### DIFF
--- a/src/main/java/net/minestom/server/utils/chunk/ChunkUpdateLimitChecker.java
+++ b/src/main/java/net/minestom/server/utils/chunk/ChunkUpdateLimitChecker.java
@@ -5,31 +5,38 @@ import org.jetbrains.annotations.ApiStatus;
 
 @ApiStatus.Internal
 public final class ChunkUpdateLimitChecker {
-    private final int historySize;
+    private final int maxHistorySize;
     private final long[] chunkHistory;
+    private int historySize;
 
     public ChunkUpdateLimitChecker(int historySize) {
-        this.historySize = historySize;
-        this.chunkHistory = new long[historySize];
+        this.maxHistorySize = historySize;
+        this.chunkHistory = new long[historySize + 1];
+        this.historySize = 0;
     }
 
     /**
      * Adds the chunk to the history
      *
-     * @param chunk chunk to add
+     * @param index chunk index to add
      * @return {@code true} if it's a new chunk in the history
      */
-    public boolean addToHistory(Chunk chunk) {
-        final long index = ChunkUtils.getChunkIndex(chunk);
+    public boolean addToHistory(long index) {
         boolean result = true;
-        final int lastIndex = historySize - 1;
+        final int lastIndex = historySize;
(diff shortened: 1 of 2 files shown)


TheMode:
Could you test it?

tahmid-23:
it fixed my issues in the sample code in the issue and in my larger codebase

TheMode:
> it fixed the sample code in the issue and in my larger codebase

I mean that you should include a test in your PR

tahmid-23:
I don't really know what a new test should do
originally, the debounce test gave a false positive, now it's a true positive, but the test itself already covered the previously problematic 0,0 chunk

src/main/java/net/minestom/server/utils/chunk/ChunkUpdateLimitChecker.java
```diff
@@ -5,31 +5,38 @@
...
                 result = false;
+                break;
             }
-            chunkHistory[i] = chunkHistory[i + 1];
         }
         chunkHistory[lastIndex] = index;
+        if (historySize < maxHistorySize) {
+            historySize++;
+        }
+        else {
```
BomBardyGamer, line 37:
```suggestion
        } else {
```

tahmid-23:
fixed by another pr

PR #1196, opened 2022-07-01
Fix floating point precision issue with RayUtils method

### LINKED ISSUE #1193: entities falling through block
on a platform one block thick, giving an Entity with EntityType ZOMBIE a velocity (10, 0, 10) (also an issue with smaller velocities) makes it fall through the platform

Steanky:
Should fix #1193. 
There appears to be a floating point precision bug in RayUtils#BoundingBoxIntersectionCheck, where very specific parameters (bounding box heights) cause the method to report no collision where there should be one. Applying the EPSILON operator to the bbOffMin/bbOffMax calculations fixes the problem. 

Includes a relevant unit test that tries ~9000 different heights, some of which cause it to fail without this fix.

### DIFF
--- a/src/main/java/net/minestom/server/collision/RayUtils.java
+++ b/src/main/java/net/minestom/server/collision/RayUtils.java
@@ -114,8 +114,8 @@ final class RayUtils {
         Point rayCentre = rayStart.add(bbCentre);
 
         // Translate bounding box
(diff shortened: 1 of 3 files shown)


TheMode:
Could you perhaps add a new physic test too?

Steanky:
Done! There are certainly more values that would cause the non-epsilon'd code to fail, but these are the ones I measured during testing.

PR #1208, opened 2022-07-05
Item and experience orbs pickup fix

RinesThaix:
No matter how you expand the bounding box on Y direction, it will always start at y=0 and end at y=height. Because of that, items and experience orbs can't be picked up if they fall lower than player position. For example, if an item goes into a stair lower half-block and there is no way for player to lower his own position to that height, he will never be able to pick it up.


ElAndree (approved):
good fix

PR #1222, opened 2022-07-09
Serverside entity's hasNoGravity check doesn't rely on it's metadata anymore

RinesThaix:
Accessing entity's metadata is expensive, especially in it's `velocityTick()`: metadata is stored in synchronized map, therefore when amount of entities reach certain point, you could see that blocked monitors consume a lot of CPU for no objective reason:
image


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -107,6 +107,7 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     // Velocity
     protected Vec velocity = Vec.ZERO; // Movement in block per second
     protected boolean hasPhysics = true;
+    protected boolean hasNoGravity = false;
```
Kebab11noel, line 110: Maybe make the field private so it will always stay in sync with the meta.

Kebab11noel (approved):

PR #1237, opened 2022-07-14
Fix attribute reading

CryptForge:
Currently any attribute that isn't fully uppercase cannot be read. This PR fixes that.


TheMode:
Can you add a test case? We have `ItemAttributeTest`

PR #1245, opened 2022-07-19
Add default tinylog writer

HSGamer:


TheMode:
Could you explain the change?

HSGamer:
> Could you explain the change?

Related to #1238
We include the default writers of tinylog so that young developers don't have to modify the services to include these writers.

PR #1258, opened 2022-07-20
Improve PlayerChangeHeldSlotEvent

### LINKED ISSUE #1257: PlayerChangeHeldSlotEvent#getSlot() returns the old slot, not the new one.
As the title says, when calling event.getSlot() you get the old slot, not the new slot.

Code used in test:
```java
public class PlayerChangeHeldSlotListener implements EventListener<PlayerChangeHeldSlotEvent> {
    @Override
    public @NotNull Class<PlayerChangeHeldSlotEvent> eventType() {
        return PlayerChangeHeldSlotEvent.class;
    }

    @Override
    public @NotNull Result run(@NotNull PlayerChangeHeldSlotEvent event) {
        byte slot = event.getSlot();
        ItemStack item = event.getPlayer().getItemInMainHand();
        if (item.isAir()) {
            event.getPlayer().sendMessage("AADOSAJDOSAIJDASO");
            return Result.SUCCESS;
        }
        event.getPlayer().sendMessage("This will be ran even if you switch to air from an item that isnt air");
        return Result.SUCCESS;
    }
}
```

Shouldn't be a difficult fix

GreatWyrm:
Changes the PlayerChangeHeldSlotEvent to better clarify what the getSlot methods will do, and add a couple of convenience methods.

Resolves #1257.

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerChangeHeldSlotEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerChangeHeldSlotEvent.java
@@ -3,6 +3,7 @@ package net.minestom.server.event.player;
 import net.minestom.server.entity.Player;
 import net.minestom.server.event.trait.CancellableEvent;
 import net.minestom.server.event.trait.PlayerInstanceEvent;
+import net.minestom.server.item.ItemStack;
 import net.minestom.server.utils.MathUtils;
 import net.minestom.server.utils.validate.Check;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/event/player/PlayerChangeHeldSlotEvent.java
```diff
@@ -13,33 +14,77 @@
...
+
+    /**
+     * Gets the slot which the player will hold.
+     * @deprecated Use {{@link #getNewSlot()}} instead
      * @return the held slot
      */
+    @Deprecated
     public byte getSlot() {
-        return slot;
+        return newSlot;
```
Marlonzockt, line 43: Should you really introduce a breaking change in a method that you are deprecating? Seems like a bad idea

src/test/java/net/minestom/server/entity/PlayerHeldIntegrationTest.java
```diff
@@ -45,7 +45,7 @@ public void playerHeldEvent(Env env) {
         var listener = env.listen(PlayerChangeHeldSlotEvent.class);
         listener.followup(event -> {
             assertEquals(player, event.getPlayer());
-            assertEquals(1, event.getSlot());
+            assertEquals(1, event.getNewSlot());
```
Marlonzockt, line 48: You probably also want to add an assert for the getOldSlot method

GreatWyrm: Added that along with a slightly more complicated test

src/main/java/net/minestom/server/listener/PlayerHeldListener.java
```diff
@@ -14,26 +14,27 @@ public static void heldListener(ClientHeldItemChangePacket packet, Player player
             return;
         }
 
-        final byte slot = (byte) packet.slot();
+        final byte newSlot = (byte) packet.slot();
```
TheMeinerLP, line 17: I think we need to add some security checks here to prevent exploits or/and cast exceptions

TheMeinerLP (changes requested):
Can you update the PR ?

src/main/java/net/minestom/server/listener/PlayerHeldListener.java
GreatWyrm: We already that on line 13, we ignore packets if the slot is invalid

GreatWyrm:
@TheMeinerLP Updated PR

mworzala (approved):
Kinda late 💀 but anyway, thanks for this :D

PR #1264, opened 2022-07-21
Create AbsoluteBlockBatch#getAffectedChunks()

MrGazdag:
This PR creates a new method in AbsoluteBlockBatch, that returns the chunk indexes that will be affected by this `AbsoluteBlockBatch`. This is useful in combination with `ChunkUtils.optionalLoadAll(Instance, long[], ChunkCallback)`, to ensure that all chunks that will be affected, are loaded at the time of calling `AbsoluteBlockBatch#apply(Instance)`.

Inspired by #1263

### DIFF
--- a/src/main/java/net/minestom/server/instance/batch/AbsoluteBlockBatch.java
+++ b/src/main/java/net/minestom/server/instance/batch/AbsoluteBlockBatch.java
@@ -49,6 +49,17 @@ public class AbsoluteBlockBatch implements Batch<Runnable> {
         this.options = options;
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


mworzala:
I don't like this API very much (the packed long isn't very intuitive to most people), but generally I think the more relevant request here is to be able to load affected chunks. We currently don't have a way to load chunks and ensure they stay loaded while operating then (allow) unload after so I dont think it makes sense to improve this quite yet. #2654 should improve this.

That said, I would prefer to do this as a BatchOption so will likely include it in an updated version of #2448. Closing for now in any case.

PR #1272, opened 2022-07-22
Add commit info from JitPack into Minestom's jar

off-by-0point5:
Adds a `Git` class holding information about the build from JitPack.
So any extension or server implementation would have access to these.

I took group and artifact fields too, so one can tell what fork of Minestom is used.

### DIFF
--- a/src/main/java/net/minestom/server/Git.java
+++ b/src/main/java/net/minestom/server/Git.java
@@ -6,5 +6,4 @@ public class Git {
 
     public static final String GROUP = "&GROUP";
     public static final String ARTIFACT = "&ARTIFACT";
-    public static final String VERSION = "&VERSION";
 }
--- a/src/main/java/net/minestom/server/Git.java
+++ b/src/main/java/net/minestom/server/Git.java
@@ -1,0 +1,10 @@
+package net.minestom.server;
+
+public class Git {
+    public static final String COMMIT = &COMMIT;
+    public static final String BRANCH = &BRANCH;
+
+    public static final String GROUP = &GROUP;
+    public static final String ARTIFACT = &ARTIFACT;
+    public static final String VERSION = &VERSION;
+}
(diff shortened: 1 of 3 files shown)


Geolykt:
Are you sure that it is viable to have the fields set as `static final`? Javac tends to inline those fields so any extensions will only know which version they compiled against at best, but not which version is currently running

off-by-0point5:
> Javac tends to inline those fields so any extensions will only know which version they compiled against at best, but not which version is currently running

Didn't know that.
Do you know if it is better to make the fields just `public static` or make them `private static final` and use getters?

Geolykt:
I believe that is up to you. I'd personally say `private static final` as these fields wouldn't be able to be modified - so JIT will have an easier time on that front

off-by-0point5:
I did that. But I'm still not sure if this will be merged eventually...

Geolykt:
The getters should be static too

src/main/java/net/minestom/server/Git.java
```diff
@@ -0,0 +1,17 @@
+package net.minestom.server;
+
+public class Git {
```
TheMode, line 3: Class should be `final`

src/main/java/net/minestom/server/Git.java
```diff
@@ -0,0 +1,17 @@
...
+    private static final String COMMIT = "&COMMIT";
+    private static final String BRANCH = "&BRANCH";
+
+    private static final String GROUP = "&GROUP";
+    private static final String ARTIFACT = "&ARTIFACT";
+
+
+
+    public String getCommit() { return COMMIT; }
```
TheMode, line 12: Drop the `get` prefix in all the getter. So just `commit` in this case

TheMode (changes requested):

mworzala (approved):

PR #1273, opened 2022-07-23
Advanced teleportation API

LooFifteen:
This PR adds support for the [Synchronize Player Position](https://wiki.vg/Protocol#Synchronize_Player_Position) packet's `Flag` field. This field allows for relative teleportation, meaning the client's velocity, pitch and yaw can be preserved when teleported.

All new methods are labelled as experimental until tested properly.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -291,12 +291,30 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
      * @throws IllegalStateException if you try to teleport an entity before settings its instance
      */
     public @NotNull CompletableFuture<Void> teleport(@NotNull Pos position, long @Nullable [] chunks) {
+        return teleport(position, chunks, new RelativeTeleportFlag[0]);
+    }
+
+    /**
+     * Teleports the entity only if the chunk at {@code position} is loaded or if
+     * {@link Instance#hasEnabledAutoChunkLoad()} returns true.
+     *
+     * @param position the teleport position
+     * @param chunks   the chunk indexes to load before teleporting the entity,
+     *                 indexes are from {@link ChunkUtils#getChunkIndex(int, int)},
+     *                 can be null or empty to only load the chunk at {@code position}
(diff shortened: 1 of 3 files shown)


TheMode:
Couldn't we automatically add those flags depending on the teleport position? Adding complexity to `teleport` doesn't seem like the right move.

SLLCoding:
These flags allow for the server to depict whether the player's camera angle or velocity should be changed when they are teleported. I have followed in the footsteps of Paper with this PR.

joshuasing (approved):

joshuasing:
This would be helpful for many things, it would be helpful if it could be merged

NxDs:
I liked the idea but the way this is implemented doesn't quite work, I'm referring to the line
https://github.com/Minestom/Minestom/blob/39dab6aeab1182d7ef9c083ec685362fe949ba35/src/main/java/net/minestom/server/entity/Player.java#L1526
That sends the packet with relative location regardless, which means that unless you set all the flags to relative the teleportation will fail client side.
I'm not sure how this PR would look like to take into account this other than checking the flag and make the coordinates relative
One possible solution would be a check like this inside synchronizePosition()
```java
 Pos offsetPos = position;
for (RelativeTeleportFlag flag : flags) {
    switch (flag) {
        case X -> offsetPos = offsetPos.sub(previousPosition.x(), 0, 0);
        case Y -> offsetPos = offsetPos.sub(0, previousPosition.y(), 0);
        case Z -> offsetPos = offsetPos.sub(0, 0, previousPosition.z());
        case YAW -> offsetPos = offsetPos.withYaw(position.yaw() - previousPosition.yaw());
        case PITCH -> offsetPos = offsetPos.withPitch(position.pitch() - previousPosition.pitch());
    }
}
```
This does work (I tested) but it's not very elegant... 
My suggestion would be to instead add a new method, teleportRelative(Pos) that would handle everything internally, this would simplify things, on top of this I'd suggest to simply drop the support for specific flags and keep it simple so you either teleport in all absolute coords or in all relative coords, I can't think of any use case where one would benefit from being able to use a position with mixed abs/rel coordinates.

LooFifteen:
Closed for #1739

PR #1274, opened 2022-07-23
Add meta for banners and shields

LooFifteen:
This PR adds a meta class, `BannerMeta`, to be used on banners and shields. Both items use the same meta.


GoldenStack:
It looks like you manually edited the DyeColor enum instead of changing DyeColorGenerator (as it says `  * AUTOGENERATED by DyeColorGenerator` at the top)

SLLCoding:
I didn't even see that, my bad.

SLLCoding:
Requires Minestom/MinestomDataGenerator#12 to continue.

src/main/java/net/minestom/server/item/banner/BannerPattern.java
```diff
@@ -0,0 +1,36 @@
+package net.minestom.server.item.banner;
+
+import net.minestom.server.color.DyeColor;
+import org.jetbrains.annotations.NotNull;
+import org.jglrxavpok.hephaistos.nbt.NBT;
+import org.jglrxavpok.hephaistos.nbt.NBTCompound;
+
+import java.util.Map;
+
+public record BannerPattern(@NotNull DyeColor color, @NotNull PatternType type) {
```
TheMode, line 10: How about moving it in `BannerMeta` as an inner class? `BannerMeta.Pattern`

src/main/java/net/minestom/server/item/banner/PatternType.java
```diff
@@ -0,0 +1,75 @@
+package net.minestom.server.item.banner;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum PatternType {
```
TheMode, line 9: Can also be moved to the banner meta class. Although I am also not sure if we want to make this an enum?

src/main/java/net/minestom/server/item/banner/PatternType.java
```diff
@@ -0,0 +1,75 @@
...
+    GRADIENT("gra"),
+    GRADIENT_UP("gru"),
+    BRICKS("bri"),
+    SKULL("sku"),
+    FLOWER("flo"),
+    MOJANG("moj"),
+    GLOBE("glb"),
+    PIGLIN("pig");
+
+    private static final @NotNull Map<String, PatternType> BY_IDENTIFIER = new HashMap<>();
```
TheMode, line 53: I'd suggest avoiding mutable maps used as an immutable view. You can stream all the enum entries and convert it to an immutable map: https://www.baeldung.com/java-stream-immutable-collection. Or more simply, instantiate your `HashMap` in the static initializer and set the field with `Map.copyOf`

src/main/java/net/minestom/server/item/banner/PatternType.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    private final @NotNull String identifier;
+
+    PatternType(@NotNull String identifier) {
+        this.identifier = identifier;
+    }
+
+    public @NotNull String getIdentifier() {
```
TheMode, line 67: `identifier` without the get prefix

TheMode (changes requested):
Some design concerns. You should additionally test it

src/main/java/net/minestom/server/item/banner/PatternType.java
SLLCoding: This won't be an enum, was just waiting for the data PR to be merged.

PR #1303, opened 2022-08-01
Instance registration events

RinesThaix:


src/main/java/net/minestom/server/event/instance/InstanceRegisteredEvent.java
```diff
@@ -0,0 +1,22 @@
+package net.minestom.server.event.instance;
+
+import net.minestom.server.event.trait.InstanceEvent;
+import net.minestom.server.instance.Instance;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Called when a new instance is registered.
```
KrystilizeNevaDies, line 8: It may not always be a new instance

TheMeinerLP:
Duplicated #608

PR #1305, opened 2022-08-01
Add convenience playSound method in PacketGroupingAudience class

GreatWyrm:
This PR adds a convenience method to the Instance class so it is quicker to play sounds in that instance.


mworzala (approved):

mworzala (changes requested):
My approval was a lie it turns out, @TheMode made a good point that this should be in the adventure audience along with the other playSound methods [here](https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/adventure/audience/PacketGroupingAudience.java#L100).

GreatWyrm:
Alright, moved the method to the PacketGroupingAudience class.

mworzala (approved):

PR #1318, opened 2022-08-08
Allowing setting pufferfish size to any value

iam4722202468:
This is useful for entity shadows. Needed for WSEE.


GoldenStack:
Shouldn't the bounding box be updated to the correct value if the provided `size` also happens to be a valid number for a state?

iam4722202468:
I don't think that is necessary, setState exists and can be used to set a valid state. This method is to set invalid states, in order to use the shadow of the pufferfish.

PR #1319, opened 2022-08-09
Add permissions wildcard support to no-verifier hasPermission

dev-hydrogen:
Adds wildcard support to hasPermission(Permission permission) and relevant tests

If a player has permission `foo.*`, `foo.bar` and `foo.bar.baz` will return true, but simply `foo` will not. Technically you could give `foo*` as a permission, which would make this return true for ”foo”.
If a player has the permission `*`, this method will always return true.

Some fancy word here for saying that this is a working/better version of #719


emortaldev:
> Some fancy word here for saying that this is a working/better version of https://github.com/Minestom/Minestom/pull/719

successor?

KrystilizeNevaDies:
I'd like to see support for proper pattern matching.

e.g.
`foo.b*r` could match queries for `foo.bar`, or `foo.boar`
`foo.bar.*s` could match `foo.bar.baz.bos`
`f*o.bar` could match `fooooooo.bar`

dev-hydrogen:
> I'd like to see support for proper pattern matching.
> 
> e.g. `foo.b*r` could match queries for `foo.bar`, or `foo.boar` `foo.bar.*s` could match `foo.bar.baz.bos` `f*o.bar` could match `fooooooo.bar`

I have an idea on how to do this, but it would mean i need to know the separator, which minestom currently doesnt take any stance on. ("odd" permissions like "namespace:foo/bar" are allowed), what should be the list of supported separators?

KrystilizeNevaDies:
Uhh, I don't think separators should even be considered personally.

They should be there as a standard, but nothing in code should identify them imo.

KrystilizeNevaDies (approved):
Looks good to me 👍🏻

KrystilizeNevaDies:
I would like to get this merged, please update to latest master commit.

PR #1322, opened 2022-08-10
 Resolve misc todos

KrystilizeNevaDies:


src/main/java/net/minestom/server/entity/ExperienceOrb.java
```diff
@@ -96,9 +88,14 @@ public void setExperienceCount(short experienceCount) {
...
     private Player getClosestPlayer(Entity entity, float maxDistance) {
+        return getClosestPlayer(entity, player -> true, maxDistance);
+    }
+
+    private Player getClosestPlayer(Entity entity, Predicate<Player> predicate, float maxDistance) {
         Player closest = entity.getInstance()
                 .getPlayers()
                 .stream()
+                .filter(predicate)
                 .min(Comparator.comparingDouble(a -> a.getDistance(entity)))
```
Marlonzockt, line 99: Might as well use getSquaredDistance here and below, right?

src/main/java/net/minestom/server/listener/SpectateListener.java
```diff
@@ -16,7 +16,12 @@ public static void listener(ClientSpectatePacket packet, Player player) {
         if (target == null || target == player)
             return;
 
-        // TODO check if 'target' is in a different instance
+        // Check if the target is in a different instance
+        if (target.getInstance() != player.getInstance()) {
+            //noinspection ConstantConditions
+            player.setInstance(target.getInstance()).thenRun(() -> player.spectate(target));
```
Marlonzockt, line 22: Doesn't this let the client switch instances if it knows the UUID of a player in that instance and is in spectator mode? Or are they shown in the middle click menu?

KrystilizeNevaDies: That's a good point, they would be able to do that. I'm not sure if it should be removed though... E.g. the vanilla server allows this.

Maybe adding an event for users to decide is a good solution.

Marlonzockt: Yes it does, however the way instances are handled in for example the ArgumentEntity makes me believe that instances are not like vanilla dimensions but rather like vanilla servers with only one dimension

KrystilizeNevaDies: Yep I think I'll have that behaviour by default with the ability to change it via an event.

mworzala:
Not sure the content but certainly many years out of date now :(

Issue #1327, opened 2022-08-12
Command bug in syntax parsing


jhooc77:
When there're same type arguments having different name, inflict bug to the later ones.
Exactly If text can be parsed to that types even if they are not of the same type, the bug occurs.
Tooltip and suggestion display properly, only processing the syntax have a bug.
The names of arguments are also important in terms of readability, so I hope they will be corrected.

In `/test` Command
Case 1:
```Java
setDefaultExecutor((sender, context) -> {
    sender.sendMessage("not work");
});

addSyntax((sender, context) -> {
    sender.sendMessage("entity");
}, ArgumentType.Entity("entity"));

addSyntax((sender, context) -> {
    sender.sendMessage("entity + wow");
}, ArgumentType.Entity("entity + wow"), ArgumentType.Literal("wow"));

addSyntax((sender, context) -> {
    sender.sendMessage("double");
}, ArgumentType.Double("double"));

addSyntax((sender, context) -> {
    sender.sendMessage("double + wow");
}, ArgumentType.Double("double + wow"), ArgumentType.Literal("wow"));
```
Result:
`/test 123` -> print `double`
`/test 123 wow` -> print `not work`
`/test username` -> print `entity`
`/test username wow` -> print `not work`

Case 2:
```Java
setDefaultExecutor((sender, context) -> {
    sender.sendMessage("not work");
});

addSyntax((sender, context) -> {
    sender.sendMessage("entity + wow");
}, ArgumentType.Entity("entity + wow"), ArgumentType.Literal("wow"));

addSyntax((sender, context) -> {
    sender.sendMessage("entity");
}, ArgumentType.Entity("entity"));

addSyntax((sender, context) -> {
    sender.sendMessage("double + wow");
}, ArgumentType.Double("double + wow"), ArgumentType.Literal("wow"));

addSyntax((sender, context) -> {
    sender.sendMessage("double");
}, ArgumentType.Double("double"));
```
Result:
`/test 123` -> print `not work`
`/test 123 wow` -> print `double + wow`
`/test username` -> print `not work`
`/test username wow` -> print `entity + wow`

Case 3: same type and same name arguemtns work properly
```Java
setDefaultExecutor((sender, context) -> {
    sender.sendMessage("not work");
});

addSyntax((sender, context) -> {
    sender.sendMessage("entity");
}, ArgumentType.Entity("entity_same"));

addSyntax((sender, context) -> {
    sender.sendMessage("entity + wow");
[...]

Issue #1326, opened 2022-08-12
ConditionalSyntax only affect last argument


jhooc77:
```Java
addConditionalSyntax((sender, commandString) -> {
    return sender instanceof ConsoleSender;
}, (sender, context) -> {
    double x = context.get(xArgument);
    double y = context.get(yArgument);
    double z = context.get(zArgument);
    Pos position = new Pos(x, y, z);
    Component message = Component.text("You have been teleported to ").append(Component.text(position.toString()).color(NamedTextColor.GRAY).decorate(TextDecoration.ITALIC)).append(Component.text(" by Console"));
    for (Entity target : context.get(targetArgument).find(sender)) {
        target.teleport(position).thenAccept(__ -> {
            if (target instanceof Player p) {
                p.sendMessage(message);
            }
        });
    }
}, targetArgument, xArgument, yArgument, zArgument);
```
![image](https://user-images.githubusercontent.com/41156145/184322815-a4d21766-d1fd-4bab-ac17-3e119dd7be73.png)

PR #1368, opened 2022-08-28
Fix command bug

jhooc77:
- [ ] #1327 
- [x] #1326


TheMode:
You may want to target the `command-parser-rework` branch instead

PR #1373, opened 2022-08-30
Corrected the incorrect warning message

Anzide:
For example,when i entered my server with a oak sign block without handler,it tips me that Block minecraft:sign doesn't have corresponding handler.However,there is no "sign" block,there is only "oak sign" block and "sign" block entity.So I think we should change the message.

### DIFF
--- a/src/main/java/net/minestom/server/instance/block/BlockManager.java
+++ b/src/main/java/net/minestom/server/instance/block/BlockManager.java
@@ -44,7 +44,7 @@ public final class BlockManager {
         if (handler == null) {
             if (dummyWarning.add(namespace)) {
                 LOGGER.warn("""
-                        Block {} does not have any corresponding handler, default to dummy.
+                        Block Entity {} does not have any corresponding handler, default to dummy.
                         You may want to register a handler for this namespace to prevent any data loss.""", namespace);
             }
             handler = BlockHandler.Dummy.get(namespace);


GoldenStack:
I'm pretty sure that Minestom doesn't have the concept of block entities. Additionally, this would apply for _every_ block, which would make it inaccurate even more of the time.

Anzide:
> 
There of course is Block Entity.You could get it by Block#registry()#getBlockEntity() or like.
And it will not warn when the block doesn't have coresponding block entity.

GoldenStack:
Looks like I didn't look into it enough - I didn't see that `BlockManager#getHandlerOrDummy` was internal and that its only use is when the block must be a block entity. Thanks.

Also, now that I'm looking at this specific code, it looks like the print message here:
https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/instance/AnvilLoader.java#L283
could be changed to refer to them as block entities instead of tile entities, as this is the only code in the repository to use "tile" instead of "block" and Hephaistos additionally sets up the idea that tile entities is the old name here:
https://github.com/jglrxavpok/Hephaistos/blob/master/common/src/main/kotlin/org/jglrxavpok/hephaistos/mca/readers/ChunkReader.kt#L104
but I don't know if that's something that should be done here (or at all)

TheMode:
A non-block entity (in the vanilla sense) can also have a handler. The system is used to store block NBT on disk, as block entity

Anzide:
> 

Maybe I should change "Block" to "Block Handler"?

PR #1374, opened 2022-08-30
Warn message update

Anzide:
Made the warn message more detailed so people can fastly locate the problem.


Anzide:
Seems I created the branch at a wrong commit...I wanted to sperate the two commits into 2 PR.Anything I can do to remedy it?Or just merge one of them.

mworzala:
Regarding the change to the block handler warn message:
A block missing a handler is not always a vanilla block entity. If you attach a handler to a dirt block, save the world, then load it again without that handler registered, you will see this warning.

Regarding the change to the command log message:
It seems unnecessary. The logger is in the command class (created with a reference), so the log message either already includes the class name (indicating it is referring to a command), or can be configured that way.

PR #1380, opened 2022-09-02
Make the start log use BrandName instead of hardcoded "Minestom"

TomJuri:


TomJuri:
Weird these tests fail even without modified code...

GreatWyrm:
The test this PR failed on is particularly flaky for some unknown reason, I've seen it fail on my PRs, and when the test was run again, it passed.

TomJuri:
Can I re run the tests?

TomJuri:
There we go

PR #1387, opened 2022-09-06
Fix block memory leak with properties function

KodingDev:
### Description

Some blocks don't have any property types, yet the `BlockImpl#properties()` function doesn't check if the list is empty. This can cause thousands of 0 length `String[]`'s to be allocated in memory alongside some unnecessary overhead. We can fix this by returning an empty immutable map.

> Unsure if caching / precomputing the properties map would be another possible performance improvement? Given Minestom already depends on caffiene.

### DIFF
--- a/src/main/java/net/minestom/server/instance/block/BlockImpl.java
+++ b/src/main/java/net/minestom/server/instance/block/BlockImpl.java
@@ -164,8 +164,8 @@ record BlockImpl(@NotNull Registry.BlockEntry registry,
     public @Unmodifiable @NotNull Map<String, String> properties() {
         final PropertyType[] propertyTypes = PROPERTIES_TYPE.get(id());
         assert propertyTypes != null;
-        if (propertyTypes.length == 0) return Collections.emptyMap();
         final int length = propertyTypes.length;
+        if (length == 0) return Map.of();
         String[] keys = new String[length];
         String[] values = new String[length];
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/instance/block/BlockImpl.java
```diff
@@ -164,6 +164,7 @@ static Collection<Block> values() {
     public @Unmodifiable @NotNull Map<String, String> properties() {
         final PropertyType[] propertyTypes = PROPERTIES_TYPE.get(id());
         assert propertyTypes != null;
+        if (propertyTypes.length == 0) return Collections.emptyMap();
         final int length = propertyTypes.length;
```
TheMode, line 168: Should be `Map.of`. `Collections` util methods are generally unnecessary.

You can also move it below to get access to the `length` variable.

mworzala: The implementation, for your viewing pleasure
```java
static <K, V> Map<K, V> of() {
    return (Map<K,V>) ImmutableCollections.EMPTY_MAP;
}
```

KodingDev: Ah should've checked, for some reason assumed it'd use varargs even though that class has overloads. Will push those two fixes.

PR #1402, opened 2022-09-13
Fix incorrect functionality of ClientSpectatePacket

KodingDev:
See code for an explanation of why this is needed.

### DIFF
--- a/src/main/java/net/minestom/server/listener/SpectateListener.java
+++ b/src/main/java/net/minestom/server/listener/SpectateListener.java
@@ -3,6 +3,8 @@ package net.minestom.server.listener;
 import net.minestom.server.entity.Entity;
 import net.minestom.server.entity.GameMode;
 import net.minestom.server.entity.Player;
+import net.minestom.server.event.EventDispatcher;
+import net.minestom.server.event.player.PlayerSpectateEvent;
 import net.minestom.server.network.packet.client.play.ClientSpectatePacket;
 
 import java.util.UUID;
@@ -37,7 +39,7 @@ public class SpectateListener {
         // Despite the name of this packet being spectate, it is sent when the player
         // uses their hotbar to switch between entities, which actually performs a teleport
         // instead of a spectate.
-        player.teleport(target.getPosition());
+        EventDispatcher.call(new PlayerSpectateEvent(player, target));
     }
 
 }
--- a/src/main/java/net/minestom/server/listener/SpectateListener.java
+++ b/src/main/java/net/minestom/server/listener/SpectateListener.java
@@ -5,6 +5,7 @@ import net.minestom.server.entity.GameMode;
 import net.minestom.server.entity.Player;
 import net.minestom.server.event.EventDispatcher;
 import net.minestom.server.event.player.PlayerSpectateEvent;
+import net.minestom.server.instance.Instance;
 import net.minestom.server.network.packet.client.play.ClientSpectatePacket;
 
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


Moulberry:
Firstly, I wouldn't implement this as a default feature, since not everyone using Minestom would want this behaviour.
I would recommend adding a "SpectatorTeleportEvent," then people can opt-in to this behavior by implementing it themselves.

Furthermore, when implementing default features into a library, you need to be exceedingly careful of not enabling exploits.
Your implementation allows a spectator to teleport to any player on the server, even if the spectating player cannot see them.
A hacker could use this to teleport to an admin (who might be in a special admin world)

Moulberry (changes requested):
Firstly, I wouldn't implement this as a default feature, since not everyone using Minestom would want this behaviour.
I would recommend adding a "SpectatorTeleportEvent," then people can opt-in to this behavior by implementing it themselves.

Furthermore, when implementing default features into a library, you need to be exceedingly careful of not enabling exploits.
Your implementation allows a spectator to teleport to any player on the server, even if the spectating player cannot see them.
A hacker could use this to teleport to an admin (who might be in a special admin world)

src/main/java/net/minestom/server/listener/SpectateListener.java
```diff
@@ -13,11 +13,22 @@ public static void listener(ClientSpectatePacket packet, Player player) {
...
             return;
+        }
 
-        // TODO check if 'target' is in a different instance
-        player.spectate(target);
+        if (target.getInstance() == null || player.getInstance() == null) {
+            return;
+        }
+
+        if (target.getInstance().getUniqueId() != player.getInstance().getUniqueId()) {
```
Moulberry, line 24: Won't this always return?

KodingDev: If they're in the same instance they'll bypass this check, i see no reason for it to always return.

KodingDev:
> Firstly, I wouldn't implement this as a default feature, since not everyone using Minestom would want this behaviour. I would recommend adding a "SpectatorTeleportEvent," then people can opt-in to this behavior by implementing it themselves.
> 
> Furthermore, when implementing default features into a library, you need to be exceedingly careful of not enabling exploits. Your implementation allows a spectator to teleport to any player on the server, even if the spectating player cannot see them. A hacker could use this to teleport to an admin (who might be in a special admin world)

Commenting on this, this is just fixing a bug over anything else. The same issues mentioned apply to the previous implementation. The issue is that the spectate function is called, but there is no default logic allowing a player to exit from spectating, so they're stuck in that state until they relog. The main purpose of this PR is to correct this to the Vanilla implementation which is just to teleport the spectator to the player.

Adjusting the implementation of this brings along the question of if you want to allow the other default listeners to be disabled too, which seems out of scope for a bugfix PR like this.

Moulberry (approved):
lgtm

src/main/java/net/minestom/server/listener/SpectateListener.java
```diff
@@ -1,23 +1,45 @@
...
         final Entity target = Entity.getEntity(targetUuid);
 
         // Check if the target is valid
-        if (target == null || target == player)
+        if (target == null || target == player) {
+            return;
+        }
+
+        // Ignore if they're not attached to any instances
+        if (target.getInstance() == null || player.getInstance() == null) {
```
TheMode, line 29: Store the instances as variable and reuse them instead.

src/main/java/net/minestom/server/network/packet/client/play/ClientSpectatePacket.java
```diff
@@ -7,6 +7,15 @@
...
 
+/**
+ * The ClientSpectatePacket is sent when the client interacts with their hotbar to switch between entities.
+ * Contrary to its name, it is actually used to teleport the player to the entity they are switching to,
+ * rather than spectating them.
+ * <p>
+ * Minestom will, by default, handle this behavior for you.
+ *
+ * @param target The UUID of the entity the player is switching to.
+ */
```
TheMode, line 18: The doc seems unnecessary here, it is also not something we want to maintain

TheMode (changes requested):

PR #1406, opened 2022-09-14
Allow listening for packets sent in the LOGIN and STATUS states

FluxCapacitor2:
Minestom packet listeners currently have no effect for subclasses of `ClientPreplayPacket` (likely because a `Player` object does not exist at the time these packets are received). This PR adds the ability to listen to these packets, and uses it for the default listeners similarly to other PLAY packets. Also, the `process` method of each packet has been moved to its own listener class (thank you @Eoghanmc22 for the suggestion!)

My usecase for this feature is listening to every `LoginPluginResponsePacket` to receive information from a proxy before the player logs in to the server. I have not used the testing framework yet, but if tests are needed I will definitely implement those.

See corresponding discussion: #1394


FluxCapacitor2:
Now that we have a configuration phase during the login process with a [public API](https://github.com/Minestom/Minestom/blob/0c9527118abc3cbd0cd0eae10e5120e28311a383/src/main/java/net/minestom/server/listener/manager/PacketListenerManager.java#L163) and #2074 has been merged, Minestom addresses all use cases for this PR.

PR #1416, opened 2022-09-27
Add EntityTeleportEvent and PlayerTeleportEvent

Elikill58:
This PR add 2 events that can be cancelled: `EntityTeleportEvent` & `PlayerTeleportEvent`

It also remove some unused imports.


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -292,6 +293,12 @@ public boolean isOnGround() {
      */
     public @NotNull CompletableFuture<Void> teleport(@NotNull Pos position, long @Nullable [] chunks) {
         Check.stateCondition(instance == null, "You need to use Entity#setInstance before teleporting an entity!");
+        
+        EntityTeleportEvent teleportEvent = this instanceof Player p ? new PlayerTeleportEvent(p, position) : new EntityTeleportEvent(this, position);
+        EventDispatcher.call(teleportEvent);
```
GreatWyrm, line 298: You may want to replace this with `EventDispatcher.callCancellable()` instead, may make the code a bit cleaner.

Elikill58: `EventDispatcher.callCancellable()` can't return a value. For me, it's not the best way for this method, and I made the same as another method below

Elikill58:
I fixed conflict

Elikill58:
Azy flemme

Issue #1413, opened 2022-09-21
Entity Velocity Packet spam


Arektor:
Any entity affected by gravity will cause all its viewers to receive an entity velocity packet **per tick**, even when perfectly still. An amount of packets that can quickly rise to absurd levels the more entities and players you add.
These packets however seem to be completely unnecessary, and to be safely skippable without any side effect.

Skipping these unnecessary packets can be achieved simply by commenting out line 596 of Entity, in function velocityTick:
`if (!isPlayer) sendPacketToViewers(getVelocityPacket());`
 
  
 
Please note however that my testing so far might not have been as extensive as it could've been, and consists of the following:
- An EntityCreature of type Zombie with NoGravity set to false - gravity seems to still be perceived just fine client-side with the fix, whether it is after knockback or simply falling from spawn or broken block.
- An EntityCreature of type Zombie with NoGravity set to true - the issue doesn't affect NoGravity entities in the first place, and the fix doesn't seem to affect any of the perceived behavior client-side.

TheMode:
I believe that such packets are sent every X ticks in vanilla, do we really want to completely remove it?

By the way you may be interested in our [testing framework](https://github.com/Minestom/Minestom/tree/master/src/test/java/net/minestom/server/entity) (all files ending with `IntegrationTest`)

Arektor:
The idea that it would be vanilla behavior completely skipped over my head, but after a few tests both on a paper and on a vanilla server, it isn't the case.

Paper (1.19.2):
https://user-images.githubusercontent.com/6461173/191709940-c9b81d1e-c3b7-46d8-ab85-bcfd3758c7a3.mp4

Vanilla (1.18.1):
https://user-images.githubusercontent.com/6461173/191710317-804cb1c7-51ac-46b1-91c1-3ce0397e7bc0.mp4
(note for this one, the entity velocity packet seems to be net.minecraft.class_2743, which does not appear on the logs because the entity does not move. There's also a slight but unlikely possibility for it to be net.minecraft.class_2684$class_2685, which appears about 2-3 times.)

 

I also tested how often the velocity packet would be sent when the entity is falling from gravity. On both paper and vanilla server, it would be sent every 150ms - so, every 3 ticks.

Regarding the testing framework... I actually have no idea how to use one😅Will try to look into it though

TheMode:
> Regarding the testing framework... I actually have no idea how to use one😅Will try to look into it though

[...]

PR #1426, opened 2022-10-02
Velocity packet fix

Arektor:
Aiming to fix the issue #1413 I pointed out about a week ago, here's what this PR does:

- When an entity has no velocity, and the latest sent velocity update corresponds to this same state, we no longer send another packet to notify the client of an inexistent change.
If however the entity has velocity, even if that velocity has not changed since the last update, we still send an update: this is because the client will predict a reduction of this velocity over time, so the fact that the velocity did not change is, in itself, a change.
- Adds an update interval for velocity. 1 by default, this lets minestom users define how often the velocity of their entity should be update, should they need to change it. As some entities like paintings cannot have velocity - at least on the client side, this also lets them disable the updates as a whole for that entity.
(It might be also worth noting that as of 1.19.2, the velocity packet does absolutely nothing on entities that are not projectiles, dynamic tiles, thrown items, or experience orbs: another reason to let people disable those updates if they so desire.)
- Tied to the point above, this also prevents the velocity update packet to be sent when Entity#setVelocity is invoked if the interval is set to 0 or lower. As the updates are disabled, no kind of velocity update whatsoever should be sent to the client for that entity, otherwise this could cause it to appear at a different location than it actually is if this entity falls into one of the categories mentioned above.

This does not alter any behavior whatsoever on the server's side. It only aims at reducing the amount of unnecessary packets sent to the client.

### DIFF
--- a/src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
+++ b/src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
@@ -9,6 +9,9 @@ import net.minestom.server.network.packet.server.play.EntityVelocityPacket;
 import net.minestom.server.utils.chunk.ChunkUtils;
 import org.junit.jupiter.api.Test;
 
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BooleanSupplier;
+
 import static org.junit.jupiter.api.Assertions.*;
 
 @EnvTest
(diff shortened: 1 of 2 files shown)


Arektor:
At the very least the update interval should be changeable (or if not changeable, have a default value) on a per-entity-type basis (since some types are unaffected by the packets, while others might potentially require less frequent updates than others). Making it a per-entity thing would mostly only be useful for niche cases, and of the order of a micro-optimization.
Regardless, as TheMode pointed out, doing it the way I first did it would mostly cause confusion, so following his request I made it a constant for now.

src/test/java/net/minestom/server/entity/EntityVelocityIntegrationTest.java
```diff
@@ -160,6 +163,30 @@ public void testHasVelocity(Env env) {
...
     }
 
+    @Test
+    public void countVelocityPackets(Env env) {
+        final int VELOCITY_UPDATE_INTERVAL = 1;
+
+        var instance = env.createFlatInstance();
+        var viewerConnection = env.createConnection();
+        var entity = new Entity(EntityType.ZOMBIE);
+        entity.setInstance(instance, new Pos(0,40,0));
```
TheMode, line 173: Should block with `#join()`

Arektor: My bad! I just realized what that was for 😅

PR #1428, opened 2022-10-04
Make separate module for testing framework

TheMode:


Articdive:
Looks good to me, There might be some more options in Gradle that could integrate more nicely with the testing but I believe this should do this twick as long as it's working on your end I don't see anything wrong. (I have not tested it on my machine)

Articdive (approved):

PR #1448, opened 2022-10-17
ComponentHoldingPacket Enhancements

Etrayed:
Some packets implementing `ComponentHoldingPacket` do not consider every component they're holding.

- Made `EntityMetaDataPacket#components()` also include `SetSlot`-type entries
- `EntityEquipmentPacket` now also considers the lore of equipment items
- Added a missing null-check in `ChatPreviewPacket`

Commit 73b4fa47917c3bce78f9d3baa1cd0f3cc1169644 is a bit controversial as I'm not really sure what the aim of these methods really is. If they're only there for Adventure translation then this commit is pretty counter-productive. But I myself for example use these methods to implement my own translation-system into Minestom. This still wouldn't justify this commit as null values are skipped by translation-systems, but I tought maybe other developers can take advantage of these changes.


Etrayed:
@TheMode is there anything wrong with this PR which I should fix?

TheMode:
My bad, seem good to me. Though we will likely to think of a better solution for items, I highly doubt the current approach to be efficient

PR #1451, opened 2022-10-18
finish refreshRecipesPacket

KyllianGamer:
Replaced the "TODO; refresh and update players recipes" with a method call to new refreshPlayerRecipes method.

### DIFF
--- a/src/main/java/net/minestom/server/recipe/RecipeManager.java
+++ b/src/main/java/net/minestom/server/recipe/RecipeManager.java
@@ -1,6 +1,10 @@
 package net.minestom.server.recipe;
 
+import net.minestom.server.MinecraftServer;
+import net.minestom.server.entity.Player;
 import net.minestom.server.network.packet.server.play.DeclareRecipesPacket;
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/recipe/RecipeManager.java
```diff
@@ -130,7 +134,25 @@ private void refreshRecipesPacket() {
         }
 
         declareRecipesPacket = new DeclareRecipesPacket(recipesCache);
-        // TODO; refresh and update players recipes
+        MinecraftServer.getConnectionManager().getOnlinePlayers().forEach(this::refreshPlayerRecipes);
+    }
+
+    public void refreshPlayerRecipes(Player player) {
```
TheMode, line 140: Shouldn't need to be public

TheMode (commented):
Nitpick. Also tests would be appreciated

mworzala:
Recipes in the protocol have been largely reworked this is no longer relevant

PR #1453, opened 2022-10-18
Add waterlogged if Stair get placed underwater

JonasDevDE:
Support for waterlogged stairs has been added when placed underwater.

### DIFF
--- a/src/main/java/net/minestom/server/instance/block/rule/vanilla/StairsPlacementRule.java
+++ b/src/main/java/net/minestom/server/instance/block/rule/vanilla/StairsPlacementRule.java
@@ -31,14 +31,13 @@ public class StairsPlacementRule extends BlockPlacementRule {
         Facing facing = this.getFacing(player);
         Shape shape = this.getShape(instance, blockPosition, facing);
         BlockFace half = BlockFace.BOTTOM; // waiting for new block faces to be implemented
-        String waterlogged = "false"; // waiting for water to be implemented
+        Boolean waterlogged = instance.getBlock(blockPosition.relative(blockFace)).equals(Block.WATER);
 
         return block.withProperties(Map.of(
-                "facing", facing.toString(),
-                "half", half.toString(),
-                "shape", shape.toString(),
-                "waterlogged", waterlogged));
-
+                "facing", facing.name().toLowerCase(),
+                "half", half.name().toLowerCase(),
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/instance/block/rule/vanilla/StairsPlacementRule.java
```diff
@@ -31,14 +31,13 @@ public Block blockPlace(@NotNull Instance instance,
...
         return block.withProperties(Map.of(
-                "facing", facing.toString(),
-                "half", half.toString(),
-                "shape", shape.toString(),
-                "waterlogged", waterlogged));
-
+                "facing", facing.name().toLowerCase(),
+                "half", half.name().toLowerCase(),
+                "shape", shape.name().toLowerCase(),
+                "waterlogged", String.valueOf(waterlogged)));
```
GreatWyrm, line 40: Since you're using a Boolean object already, I'd recommend:
```suggestion
                "waterlogged", waterlogged.toString());
```

src/main/java/net/minestom/server/instance/block/rule/vanilla/StairsPlacementRule.java
```diff
@@ -77,9 +76,7 @@ private enum Facing {
 
         @NotNull
         public Pair<@Nullable Shape, @Nullable Facing> getFront(@NotNull Instance instance, @NotNull Point blockPosition) {
-            // TODO FIX
-            return null;
-            //return this.getProperties(instance, blockPosition.clone().add(this.front));
+            return this.getProperties(instance, blockPosition.add(this.front)); // @wb do we have to clone the blockPosition before modify?
```
GreatWyrm, line 79: No, Points are immutable, so no cloning is necessary

src/main/java/net/minestom/server/instance/block/rule/vanilla/StairsPlacementRule.java
```diff
@@ -171,4 +166,5 @@ private Facing getFacing(@NotNull Player player) {
             return Facing.WEST;
         }
     }
+
```
GreatWyrm, line 169: Empty line cropped up? I'd recommend removing this for cleanliness.

GreatWyrm (changes requested):

PR #1454, opened 2022-10-19
Fix respawning players falling into the void

GreatWyrm:
As documented in [wiki.vg's Respawn Packet](https://wiki.vg/Protocol#Respawn), the client automatically unloads chunks when sending a respawn packet to the server.

However, Minestom was not properly updating their chunks when respawning them. The client unloaded their chunks, but Minestom did not send chunk packets back to the client, so the client started falling endlessly into the void.

This PR rectifies this issue by properly handling the chunk unload, and then reloading the chunks for the player before teleporting them.

### DIFF
--- a/src/test/java/net/minestom/server/entity/player/PlayerRespawnChunkTest.java
+++ b/src/test/java/net/minestom/server/entity/player/PlayerRespawnChunkTest.java
@@ -3,11 +3,20 @@ package net.minestom.server.entity.player;
 import net.minestom.server.MinecraftServer;
 import net.minestom.server.coordinate.Pos;
 import net.minestom.server.entity.Player;
+import net.minestom.server.event.player.PlayerChunkLoadEvent;
+import net.minestom.server.event.player.PlayerChunkUnloadEvent;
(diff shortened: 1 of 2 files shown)


bendgk:
Can confirm that this was an issue and that this PR fixes it, LGTM!

src/test/java/net/minestom/server/entity/player/PlayerRespawnChunkTest.java
```diff
@@ -0,0 +1,103 @@
...
+        loadChunkTracker.assertCount(chunkLoads);
+    }
+
+    @Test
+    public void testPlayerTryRespawn(Env env) {
+        var instance = env.createFlatInstance();
+        var connection = env.createConnection();
+        Player player = connection.connect(instance, new Pos(0, 40, 0)).join();
+
+        player.eventNode().addListener(PlayerChunkLoadEvent.class, playerChunkLoadEvent -> {
```
TheMode, line 60: There is no reason to load any chunk when the player respawn at the same position. Packets being sent is an implementation detail

TheMode (changes requested):

PR #1246, opened 2022-07-19
ComponentLogger & Colored Terminal

HSGamer:
https://user-images.githubusercontent.com/25300232/179666348-fb64d108-42a3-48e6-9410-29dd9b495e5d.mp4

PR #1460, opened 2022-10-22
ComponentLogger & Colored Terminal

HSGamer:
Continue of #1246 

Only change necessary stuff.

https://user-images.githubusercontent.com/25300232/179666348-fb64d108-42a3-48e6-9410-29dd9b495e5d.mp4

### DIFF
--- a/src/main/java/net/minestom/server/adventure/provider/TerminalColorConverter.java
+++ b/src/main/java/net/minestom/server/adventure/provider/TerminalColorConverter.java
@@ -1,0 +1,108 @@
+package net.minestom.server.adventure.provider;
+
+import net.kyori.adventure.text.format.NamedTextColor;
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+import net.minestom.server.utils.PropertyUtils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A string converter to convert a string to an ansi-colored one.
+ *
+ * @see <a href="https://github.com/Minecrell/TerminalConsoleAppender/blob/master/src/main/java/net/minecrell/terminalconsole/MinecraftFormattingConverter.java">TerminalConsoleAppender</a>
+ * @see <a href="https://github.com/PaperMC/Paper/blob/41647af74caed955c1fd5b38d458ee59298ae5d4/patches/server/0591-Add-support-for-hex-color-codes-in-console.patch">Paper</a>
+ */
+public final class TerminalColorConverter {
+    public static final char COLOR_CHAR = '§';
+    public static final LegacyComponentSerializer SERIALIZER = LegacyComponentSerializer.builder()
+            .character(TerminalColorConverter.COLOR_CHAR)
+            .flattener(MinestomFlattenerProvider.INSTANCE)
+            .hexColors()
+            .build();
+    private static final boolean SUPPORT_HEX_COLOR = PropertyUtils.getBoolean("minestom.terminal.support-hex-color", true);
+    private static final boolean SUPPORT_COLOR = PropertyUtils.getBoolean("minestom.terminal.support-color", true);
+
+    private static final String RGB_ANSI = "\u001B[38;2;%d;%d;%dm";
+    private static final String ANSI_RESET = "\u001B[m";
+    private static final String LOOKUP = "0123456789abcdefklmnor";
+    private static final String[] ANSI_CODES = new String[]{
+            getAnsiColor(NamedTextColor.BLACK, "\u001B[0;30m"), // Black §0
+            getAnsiColor(NamedTextColor.DARK_BLUE, "\u001B[0;34m"), // Dark Blue §1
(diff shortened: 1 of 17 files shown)


src/main/java/net/minestom/server/adventure/provider/TerminalColorConverter.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A string converter to convert a string to an ansi-colored one.
+ *
+ * @see <a href="https://github.com/Minecrell/TerminalConsoleAppender/blob/master/src/main/java/net/minecrell/terminalconsole/MinecraftFormattingConverter.java">TerminalConsoleAppender</a>
+ * @see <a href="https://github.com/PaperMC/Paper/blob/41647af74caed955c1fd5b38d458ee59298ae5d4/patches/server/0591-Add-support-for-hex-color-codes-in-console.patch">Paper</a>
+ */
+public final class TerminalColorConverter {
```
TheMode, line 16: Should be tested.
Also do we really need to expose the class? Couldn't it become package-private in `terminal`?

HSGamer: > Should be tested. Also do we really need to expose the class? Couldn't it become package-private in `terminal`?

Probably not. This may be useful if someone want to make their own terminal.

TheMode: I'd be for hiding it at least for now. And the terminal package seems more fitting anyway.

HSGamer: > I'd be for hiding it at least for now. And the terminal package seems more fitting anyway.

Not just terminal use this. `MinestomComponentLoggerProvider` need this too, at least the `SERIALIZER` of the converter. I just don't want to duplicate `COLOR_CHAR` just for the package-private one.
Also, if a developer needs this for their own terminal, they don't have to make an PR to change it back.

HSGamer: Or I'm just bad at creating classes that split the purpose for both `MinestomComponentLoggerProvider` and the `terminal`.

HSGamer: but I'll move the converter to `terminal` anyways

TheMode: I am indeed not sure that the serializer has to be here

HSGamer: I take this back. `MinestomFlattenerProvider` is used in the converter. Should I public that instance @TheMode ?

TheMode: Cant you move the serializer in `MinestomComponentLoggerProvider`? `COLOR_CHAR` being duplicated doesn't sound like a huge issue. I'd prefer limiting exposed stuff to the strict minimum until necessary

HSGamer: > Cant you move the serializer in `MinestomComponentLoggerProvider`? `COLOR_CHAR` being duplicated doesn't sound like a huge issue. I'd prefer limiting exposed stuff to the strict minimum until necessary

I just got a better idea to use `SECTION_CHAR` from `LegacyComponentSerializer` instead of creating a constant for it.
Okay, I've moved the converter to the terminal package. Not sure about package-private thing, but I'd leave them as public for developers who want to create their own terminal.

src/main/java/net/minestom/server/terminal/TerminalColorConverter.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A string converter to convert a string to an ansi-colored one.
+ *
+ * @see <a href="https://github.com/Minecrell/TerminalConsoleAppender/blob/master/src/main/java/net/minecrell/terminalconsole/MinecraftFormattingConverter.java">TerminalConsoleAppender</a>
+ * @see <a href="https://github.com/PaperMC/Paper/blob/41647af74caed955c1fd5b38d458ee59298ae5d4/patches/server/0591-Add-support-for-hex-color-codes-in-console.patch">Paper</a>
+ */
+public final class TerminalColorConverter {
```
TheMode, line 16: Should be package-private

src/main/java/net/minestom/server/adventure/provider/TerminalColorConverter.java
TheMode: There is nothing we do there that third party cannot do, may as well hide it to allow ourselves some breaking change down the line

src/main/java/net/minestom/server/terminal/TerminalColorConverter.java
HSGamer: fine

PR #1474, opened 2022-10-24
Add a way to override hasPermission method

LoisDuplain:
Permission manager plugins needs to override hasPermission method.

Can be used like any other events:

```java
	public void handlePlayerCheckPermissionEvent(PlayerCheckPermissionEvent event) {
		boolean hasPermission = event.hasPermission();

		if (hasPermission)
			return;
		
		final Player player = event.getPlayer();

		// Your behavior

		event.setHasPermission(/* true or false */);
	}
```

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerCheckPermissionEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerCheckPermissionEvent.java
@@ -1,0 +1,42 @@
+package net.minestom.server.event.player;
+
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.trait.PlayerEvent;
+import net.minestom.server.permission.Permission;
+import org.jetbrains.annotations.NotNull;
+
+public class PlayerCheckPermissionEvent implements PlayerEvent {
+
+    private final Player player;
+    private final Permission permission;
+    private final String permissionName;
+    private boolean hasPermission;
+
+    public PlayerCheckPermissionEvent(Player player, Permission permission, String permissionName, boolean hasPermission) {
+        this.player = player;
+        this.permission = permission;
+        this.permissionName = permissionName;
+        this.hasPermission = hasPermission;
+    }
+
+    @Override
+    public @NotNull Player getPlayer() {
+        return this.player;
+    }
+
+    public Permission getPermission() {
+        return permission;
+    }
+
+    public String getPermissionName() {
+        return permissionName;
+    }
+
+    public boolean hasPermission() {
+        return hasPermission;
+    }
+
+    public void setHasPermission(boolean hasPermission) {
+        this.hasPermission = hasPermission;
+    }
+}
--- a/src/main/java/net/minestom/server/permission/PermissionHandler.java
+++ b/src/main/java/net/minestom/server/permission/PermissionHandler.java
@@ -1,5 +1,8 @@
 package net.minestom.server.permission;
 
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.EventDispatcher;
+import net.minestom.server.event.player.PlayerCheckPermissionEvent;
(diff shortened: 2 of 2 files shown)


src/main/java/net/minestom/server/event/player/PlayerCheckPermissionEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class PlayerCheckPermissionEvent implements PlayerEvent {
+
+    private final Player player;
+    private final Permission permission;
+    private final String permissionName;
+    private boolean hasPermission;
+
+    public PlayerCheckPermissionEvent(Player player, Permission permission, String permissionName, boolean hasPermission) {
```
dev-hydrogen, line 15: Why use a Player here? Just use the PermissionHandler itself instead. Not everything that has permissions is a player entity.

src/main/java/net/minestom/server/permission/PermissionHandler.java
```diff
@@ -63,12 +66,22 @@ default void removePermission(@NotNull String permissionName) {
...
         for (Permission permissionLoop : getAllPermissions()) {
             if (permissionLoop.equals(permission)) {
-                return true;
+                result = true;
+                break;
             }
         }
-        return false;
+
+        if (this instanceof Player player) {
```
dev-hydrogen, line 78: Again using Player leaves out ConsoleSenders aswell as ServerSenders, possibly more

src/main/java/net/minestom/server/permission/PermissionHandler.java
```diff
@@ -100,13 +113,21 @@ default Permission getPermission(@NotNull String permissionName) {
...
+        boolean result = false;
 
+        final Permission permission = getPermission(permissionName);
         if (permission != null) {
             // Verify using the permission verifier
-            return permissionVerifier == null || permissionVerifier.isValid(permission.getNBTData());
+            result = permissionVerifier == null || permissionVerifier.isValid(permission.getNBTData());
+        }
+
+        if (this instanceof Player player) {
```
dev-hydrogen, line 124: I dont think the *exact same event* should override both the hasPermission method without a verifier and the one with a verifier. Possibly split it up to 2 events, or a nullable PermissionVerifier in the original event?
Again, just return the PermissionHandler here too.

src/main/java/net/minestom/server/event/player/PlayerCheckPermissionEvent.java
```diff
@@ -0,0 +1,42 @@
+package net.minestom.server.event.player;
+
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.trait.PlayerEvent;
+import net.minestom.server.permission.Permission;
+import org.jetbrains.annotations.NotNull;
+
+public class PlayerCheckPermissionEvent implements PlayerEvent {
```
dev-hydrogen, line 8: This should not be a PlayerEvent.

src/main/java/net/minestom/server/event/player/PlayerCheckPermissionEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final Player player;
+    private final Permission permission;
+    private final String permissionName;
```
dev-hydrogen, line 12: Why do you need both the permission as well as its name? You can just get the name from the Permission object easily.

dev-hydrogen (changes requested):

HSGamer:
I did something similar but for `PermissionHandler`
#1138 
Man, it's old now.

dev-hydrogen:
@LoisDuplain you should probably address the issues i commented on if you still want this merged

mworzala:
Permissions are removed as of #2445

PR #1488, opened 2022-10-29
WIP Instance Abstration

KrystilizeNevaDies:
WIP Instance Abstraction.

The goal of this pr is to remove `Instance`'s reliance on chunks/sections.
`Instance` implementations in the future should be able to be as simple as `Map<Point, Block>`.


mworzala:
Closing this, can come back if there is ever more incentive to work on it. Needs some more discussion before large amount of work anyway.

PR #1493, opened 2022-10-29
1.19.3 support

TheMode:


MenschenToaster:
Any updates on this?

SteinGaming:
There is an issue when using 0488915fdaff4b51c4375da27be2dcf79215435d (from JitPack), which causes materials from items not being sent correctly. 

While Minestom says the item is Material.COMPASS, the client has a Material.SLIME_BALL (maybe shows that the id moved 2 to the back (from 833, the compass id, to 831, the slime ball id), didn't test much more than this)

The issue isn't present in the 1.19 branch, and I confirmed that the issue isn't due to the proxy or it's plugins.
I also confirmed that the Material is set correctly using the Remote JVM debugger, which showed that the item itself and in the players inventory was set to COMPASS.

TheMode:
We need to update our data generator https://github.com/Minestom/MinestomDataGenerator/tree/1.19.3 which hasn't been done yet due to some api changes

SquaredStudios-MC:
Any updates?

MenschenToaster:
> Any updates?

The Branch is stable for the stuff I use minestom for. 
I'm worried, that it will take them as long as this for 1.19.4 tho :/

CodeDoctorDE:
I think this can be closed by #1851?

mworzala:
Done in #1851

PR #1497, opened 2022-11-01
Fix entities vanishing upon respawn

### LINKED ISSUE #1480: Player is invisible after Respawn
Hello, if a player respawn, the play cant see the other players!

GreatWyrm:
Due what I suspect is the client unloading entities when respawning, current Minestom will see all other entities vanish upon respawning.

This PR resolves this by resending the entities when respawning, and also makes a minor fix to some surrounding code to use the Player's chunk render distance instead of the server default render distance.

Closes #1480 .

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -455,7 +455,7 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
         Pos respawnPosition = respawnEvent.getRespawnPosition();
 
         // The client unloads chunks when respawning, so resend all chunks next to spawn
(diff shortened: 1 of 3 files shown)


ElAndree:
What happens if settings.getViewDistance() > MinecraftServer.getChunkViewDistance() ?
Does the client send this setting or where does it come from? 
Maybe there should be a check that settings.getViewDistance() is maximum MinecraftServer.getChunkViewDistance()

GreatWyrm:
Good point, I should implement a couple of checks to make sure it can't go to an invalid bounds.

This setting comes from the JoinGamePacket and the ClientSettingsPacket.

PR #1502, opened 2022-11-05
add BungeeGuard authentication

tahmid-23:


Steanky:
My name is Steank and I approve this message.

nekorii04:
My name is nekorii04 and I approve this message.

src/main/java/net/minestom/server/extras/bungee/BungeeCordProxy.java
```diff
@@ -1,12 +1,27 @@
...
+ * but managing your firewall is still recommended.
  * <p>
  * Please consider using {@link net.minestom.server.extras.velocity.VelocityProxy} instead.
  */
 public final class BungeeCordProxy {
 
+    /**
+     * Indicates that a BungeeGuard authentication was invalid due missing, multiple, or invalid tokens.
+     */
+    public static final Component INVALID_TOKEN = Component.text("Invalid connection, please connect through the proxy", NamedTextColor.RED);
```
dev-hydrogen, line 21: Possibly add something like "Please notify an administrator about this"

src/main/java/net/minestom/server/network/packet/client/handshake/HandshakePacket.java
```diff
@@ -78,7 +85,8 @@ public void process(@NotNull PlayerConnection connection) {
                 );
 
                 List<GameProfile.Property> properties = new ArrayList<>();
-                if (split.length == 4) {
+                boolean foundBungeeGuardToken = false;
+                if (hasProperties) {
```
dev-hydrogen, line 89: Nitpick but why not put foundBungeeGuardToken inside the if statement?

src/main/java/net/minestom/server/network/packet/client/handshake/HandshakePacket.java
```diff
@@ -92,8 +100,24 @@ public void process(@NotNull PlayerConnection connection) {
...
+
+                            foundBungeeGuardToken = true;
+                        }
+
                         properties.add(new GameProfile.Property(nameString, valueString, signatureString));
                     }
+
+                    if (BungeeCordProxy.isBungeeGuardEnabled() && !foundBungeeGuardToken) {
+                        socketConnection.sendPacket(new LoginDisconnectPacket(BungeeCordProxy.INVALID_TOKEN));
+                        socketConnection.disconnect();
```
dev-hydrogen, line 118: Should probably extract the disconnection into its own method as its re-used code 3 times

dev-hydrogen (changes requested):

src/main/java/net/minestom/server/network/packet/client/handshake/HandshakePacket.java
tahmid-23: what's the proper way to log something like this?

dev-hydrogen: I'm definitely not the right person to say what the *proper* way is, however most other classes seem to use     
`public final static Logger LOGGER = LoggerFactory.getLogger(YourClass.class);`
And some just use `MinecraftServer.LOGGER`
It's a bit inconsistent, but id personally probably go with the former rather than the latter.

src/main/java/net/minestom/server/network/packet/client/handshake/HandshakePacket.java
```diff
@@ -27,14 +29,20 @@
 public record HandshakePacket(int protocolVersion, @NotNull String serverAddress,
                               int serverPort, int nextState) implements ClientPreplayPacket {
 
+    public final static Logger LOGGER = LoggerFactory.getLogger(HandshakePacket.class);
```
TheMode, line 32: Doesnt need to be public

src/main/java/net/minestom/server/network/packet/client/handshake/HandshakePacket.java
```diff
@@ -27,14 +29,20 @@
...
     /**
      * Text sent if a player tries to connect with an invalid version of the client
      */
     private static final Component INVALID_VERSION_TEXT = Component.text("Invalid Version, please use " + MinecraftServer.VERSION_NAME, NamedTextColor.RED);
-    private static final Component INVALID_BUNGEE_FORWARDING = Component.text("If you wish to use IP forwarding, please enable it in your BungeeCord config as well!", NamedTextColor.RED);
+
+    /**
+     * Indicates that a BungeeGuard authentication was invalid due missing, multiple, or invalid tokens.
+     */
+    public static final Component INVALID_BUNGEE_FORWARDING = Component.text("Invalid connection, please connect through the BungeeCord proxy. If you believe this is an error, contact a server administrator.", NamedTextColor.RED);
```
TheMode, line 42: Shall be `private` as well

src/main/java/net/minestom/server/extras/bungee/BungeeCordProxy.java
```diff
@@ -24,4 +34,45 @@ public static void enable() {
...
+    public static boolean isValidBungeeGuardToken(@NotNull String token) {
+        return isBungeeGuardEnabled() && bungeeGuardTokens.contains(token);
+    }
+
+    /**
+     * Gets the maximum length of the handshake packet.
+     *
+     * @return The maximum length of the handshake packet
+     */
+    public static int getMaxHandshakeLength() {
```
TheMode, line 73: Looks like an implementation detail of the handshake packet, and no need to be exposed

TheMode (changes requested):

PR #1510, opened 2022-11-08
Added player itself as a viewer for hand swing methods

Obyvante:


TheMode:
Have you tested it? It should cause the method to be called indefinitely

Obyvante:
> Have you tested it? It should cause the method to be called indefinitely

I haven't tested with this updated code base but have tested with a test command with the content of:
```java
player.sendPacketToViewersAndSelf(new EntityAnimationPacket(getEntityId(), hand));
```

emortaldev:
why is this needed?

TheMode:
> > Have you tested it? It should cause the method to be called indefinitely
> 
> I haven't tested with this updated code base but have tested with a test command with the content of:
> 
> ```java
> player.sendPacketToViewersAndSelf(new EntityAnimationPacket(player.getEntityId(), hand));
> ```
> 
> and it was working fine for me. (both hands)

Test the current PR.

Obyvante:
@TheMode As you said it becomes an infinite loop(swinging). Since it is not that important and you can achieve the same thing with one line of code with packets, we can ignore this update.

Thank you for your time!

PR #1514, opened 2022-11-10
Add getter for buffer

Elikill58:
Everything is in title.

It's a quick pull request to made easier for others to get default buffer.

### DIFF
--- a/src/main/java/net/minestom/server/network/NetworkBuffer.java
+++ b/src/main/java/net/minestom/server/network/NetworkBuffer.java
@@ -264,6 +264,9 @@ public final class NetworkBuffer {
         }
     }
 
+    public ByteBuffer getBuffer() {
+        return this.nioBuffer;
+    }
 
     public sealed interface Type<T> permits NetworkBufferTypes.TypeImpl {
     }


TheMode:
Could you explain your use-case?

Elikill58:
> Could you explain your use-case?

Be able to use all direct buffer method, or copy it to another platform as netty. For example to convert ByteBuffer (from nio) to ByteBuf (from netty) it's `Unpooled.wrappedBuffer(nioBuffer)` which is very simple. But now, it's very complicated

TheMode:
You can use `copyTo`

Elikill58:
> You can use `copyTo`

Too complicated, and it can be useful for multiple others things because it's just the main things that is now accessible in an easy way

TheMode:
Exposing the raw internal buffer would be much more dangerous. There is no guarantee that the buffer will stay the same for the entire `NetworkBuffer` lifetime (i.e. resizing), you may break some of the class features by messing with the buffer (position/limit), and you do not even have a guarantee that the class will keep using `ByteBuffer` instead of a raw address from `Unsafe`, or panama's future `MemorySegment`.

A better fix would be the addition of new `copyTo` methods. Although I do not understand what make the current one complicated.

Elikill58:
Maybe we can just return a copied buffer.

It's complicated as we should do multiple things. It's longer than just a guetter and call a method with one argument.
Also, allow to get it and edit it is at our own risk, but it also can help if someone want to do packet editions and be sure data are fine for the server as ViaVersion can do, or maybe moded things.
But yes, I will edit my PR to a copy method

TheMode:
> Maybe we can just return a copied buffer.

You can make an utils using `copyTo` or even `toByteArray`. The more I think about it the less I believe that additional methods are required. They would just make future changes riskier.

PR #1518, opened 2022-11-13
new block tracker api

KrystilizeNevaDies:
New block tracker api designed to help optimize things such as section palettes and world updates


mworzala:
Long out of date and not being worked on, partially solved by new block batch api, and partially should be external library. Can reopen if there is more work on the subject.

PR #1523, opened 2022-11-17
new area api

KrystilizeNevaDies:


mworzala:
Same comments as https://github.com/Minestom/Minestom/pull/1518

PR #1542, opened 2022-11-27
Fix coordinate_scale type (int -> double)

### LINKED ISSUE #653: Minestom doesn't let players connect with Gate proxy
When the proxy sends you to a Minestom server it just loses connection
image
however Minestom says a player logs on and then leaves

Doesn't work with both Velocity and Bungee forwarding

ZakShearman:
Closes #653 

See https://wiki.vg/Protocol#Login_.28play.29 for the protocol.

coordinate_scale was previously treated as an int, not a double like specified.


ZakShearman:
Meant to mark as draft, don't merge yet

ZakShearman:
finished and tested now :)

PR #1549, opened 2022-11-29
Add drop reason and drop amount to item drop event

### LINKED ISSUE #1548: [Feature request] Add reason and dropType to ItemDropEvent
This feature proposes the addition of two bits of information to `ItemDropEvent`:
- a reason on why the item was dropped
  Examples:
  - `INVENTORY`: This item was dropped from an inventory.
  - `CLOSED_INVENTORY`: This item was dropped since an inventory was closed.
  - `HELD_ITEM`: This item was dropped using Q without using the player inventory.
  - `OTHER`: Not listed here.
- a drop type indicating how much of the item was dropped:
  - `STACK`: The entire stack of items were dropped
  - `SINGLE`: Only a single item from the stack was dropped.

other potential ideas:
- original dropped stack count: if `SINGLE` is used, should be the amount of items before dropping

GreatWyrm:
Closes #1548. 

This adds 2 new fields to item drop event to better describe what originated the item drop event and what it will do.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -1028,16 +1028,18 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
     }
 
     /**
-     * Calls an {@link ItemDropEvent} with a specified item.
+     * Calls an {@link ItemDropEvent} with a specified item, drop reason, and drop amount.
      * <p>
      * Returns false if {@code item} is air.
      *
      * @param item the item to drop
+     * @param reason The reason why this ItemStack was dropped
+     * @param amount The status of the amount of items, SINGLE for a single item, and STACK for the entire ItemStack
      * @return true if player can drop the item (event not cancelled), false otherwise
      */
-    public boolean dropItem(@NotNull ItemStack item) {
+    public boolean dropItem(@NotNull ItemStack item, ItemDropEvent.DropReason reason, ItemDropEvent.DropAmount amount) {
         if (item.isAir()) return false;
-        ItemDropEvent itemDropEvent = new ItemDropEvent(this, item);
+        ItemDropEvent itemDropEvent = new ItemDropEvent(this, item, reason, amount);
         EventDispatcher.call(itemDropEvent);
         return !itemDropEvent.isCancelled();
     }
@@ -1458,7 +1460,7 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
         }
         if (!cursorItem.isAir()) {
             // Add item to inventory if he hasn't been able to drop it
-            if (!dropItem(cursorItem)) {
+            if (!dropItem(cursorItem, ItemDropEvent.DropReason.CLOSED_INVENTORY, ItemDropEvent.DropAmount.STACK)) {
                 getInventory().addItemStack(cursorItem);
             }
         }
--- a/src/main/java/net/minestom/server/event/item/ItemDropEvent.java
+++ b/src/main/java/net/minestom/server/event/item/ItemDropEvent.java
@@ -11,12 +11,16 @@ public class ItemDropEvent implements PlayerInstanceEvent, ItemEvent, Cancellabl
 
     private final Player player;
     private final ItemStack itemStack;
+    private final DropReason dropReason;
+    private final DropAmount dropAmount;
 
     private boolean cancelled;
 
-    public ItemDropEvent(@NotNull Player player, @NotNull ItemStack itemStack) {
+    public ItemDropEvent(@NotNull Player player, @NotNull ItemStack itemStack, @NotNull DropReason reason, @NotNull DropAmount amount) {
         this.player = player;
         this.itemStack = itemStack;
+        this.dropReason = reason;
+        this.dropAmount = amount;
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


TheMeinerLP:
Its possible to update the PR ?

GreatWyrm:
@TheMeinerLP Updated PR to latest on master

TheMeinerLP:
> Recommendation for extensibility: `DropReason` should be a class like [`DamageType`](https://javadoc.minestom.net/net/minestom/server/entity/damage/DamageType.html), and move the predefines one to `public static` fields.

It can't be like that because there is none generated data for it.

src/main/java/net/minestom/server/item/drop/DropReason.java
```diff
@@ -0,0 +1,17 @@
+package net.minestom.server.item.drop;
+
+import net.minestom.server.inventory.AbstractInventory;
+import org.jetbrains.annotations.NotNull;
+
+public class DropReason {
```
TheMeinerLP, line 6: Nitpick: Make it final for hardening api or make it into sealed interface

TheMeinerLP: Add also javadocs.

src/main/java/net/minestom/server/item/drop/DropReason.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public static DropReason fromInventory(@NotNull AbstractInventory inventory) {
+        return new DropReasonInventory(inventory);
+    }
+    public static DropReason fromInventoryClose() {
+        return new DropReasonClose();
+    }
+    public static DropReason fromHotbar(int slot) {
+        return new DropReasonHotbar(slot);
+    }
```
TheMeinerLP, line 16: Missing javadocs

src/main/java/net/minestom/server/item/drop/DropReasonHotbar.java
```diff
@@ -0,0 +1,17 @@
+package net.minestom.server.item.drop;
+
+/**
+ * The item was dropped from the hotbar (default key: q)
+ */
+public class DropReasonHotbar extends DropReason {
```
TheMeinerLP, line 6: Nitpick: Add final

src/main/java/net/minestom/server/item/drop/DropReasonClose.java
```diff
@@ -0,0 +1,6 @@
+package net.minestom.server.item.drop;
+
+/**
+ * The item was dropped because the inventory was closed
+ */
+public class DropReasonClose extends DropReason {}
```
TheMeinerLP, line 6: Nitpick: add final

src/main/java/net/minestom/server/item/drop/DropReasonInventory.java
```diff
@@ -0,0 +1,20 @@
+package net.minestom.server.item.drop;
+
+import net.minestom.server.inventory.AbstractInventory;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The item was dropped from an inventory
+ */
+public class DropReasonInventory extends DropReason {
```
TheMeinerLP, line 9: Nitpick: add final

TheMeinerLP (changes requested):

src/main/java/net/minestom/server/item/drop/DropReason.java
GreatWyrm: The entire point of having the DropReason class is that server implementations can make their own version if they need it (say if they create another reason to drop an item, like DropReasonHitByMob).
Making that an interface and sealed would defeat the point of it.

We won't have generated data for this.

I've addressed your other nitpicks.

TheMeinerLP: The point is that we should rather do this with events and an interface structure but that should be hardened by the server itself. 

I have the following in mind:
```java
// This class is harded
public sealed interface DropReason permits CustomDropReason {
}
// This class can be extend from User
public non-sealed interface CustomDropReason extends DropReason {
}
```

src/main/java/net/minestom/server/item/drop/DropReasonClose.java
```diff
@@ -0,0 +1,6 @@
+package net.minestom.server.item.drop;
+
+/**
+ * The item was dropped because the inventory was closed
+ */
+public final class DropReasonClose extends DropReason {}
```
KermanIsPretty, line 6: Could this be a record?

src/main/java/net/minestom/server/item/drop/DropReasonHotbar.java
```diff
@@ -0,0 +1,17 @@
+package net.minestom.server.item.drop;
+
+/**
+ * The item was dropped from the hotbar (default key: q)
+ */
+public final class DropReasonHotbar extends DropReason {
```
KermanIsPretty, line 6: Could this be a record?

src/main/java/net/minestom/server/item/drop/DropReasonInventory.java
```diff
@@ -0,0 +1,20 @@
+package net.minestom.server.item.drop;
+
+import net.minestom.server.inventory.AbstractInventory;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The item was dropped from an inventory
+ */
+public final class DropReasonInventory extends DropReason {
```
KermanIsPretty, line 9: Could this be a record?

src/main/java/net/minestom/server/item/drop/DropReason.java
```diff
@@ -0,0 +1,30 @@
+package net.minestom.server.item.drop;
+
+import net.minestom.server.inventory.AbstractInventory;
+import org.jetbrains.annotations.NotNull;
+
+public class DropReason {
```
KermanIsPretty, line 6: Consider making DropReason a interface instead of a class.
I would also just include the default implementations inside of DropReason. Like DropReason.InventoryClose, DropReason.Hotbar. As they are just simple data holders.

If you do seal it I wouldn't mind having it under non-sealed DropReason.Custom

I was just thinking this to be more aligned with #2531 Click

PR #1550, opened 2022-11-30
Prevent unsafe velocity packet

Obyvante:
Changed "short" params to "int" to have a safe constructor. Limiting to a short minimum and maximum will help us to prevent unsafe values after surpassing the short minimum and maximum.

Tested with the default game environment and everything works well. (passed all tests from EntityVelocityIntegrationTest except knockback tests which shouldn't be that predictable, IMO.)
image

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -1528,10 +1528,7 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
     }
 
     protected @NotNull Vec getVelocityForPacket() {
-        return this.velocity.mul(8000f / MinecraftServer.TICK_PER_SECOND)
-                .withX(x -> MathUtils.clamp(x, Short.MIN_VALUE, Short.MAX_VALUE))
-                .withY(y -> MathUtils.clamp(y, Short.MIN_VALUE, Short.MAX_VALUE))
-                .withZ(z -> MathUtils.clamp(z, Short.MIN_VALUE, Short.MAX_VALUE));
+        return this.velocity.mul(8000f / MinecraftServer.TICK_PER_SECOND);
     }
 
     protected @NotNull EntityVelocityPacket getVelocityPacket() {
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -51,6 +51,7 @@ import net.minestom.server.timer.Schedulable;
 import net.minestom.server.timer.Scheduler;
 import net.minestom.server.timer.TaskSchedule;
 import net.minestom.server.utils.ArrayUtils;
+import net.minestom.server.utils.MathUtils;
 import net.minestom.server.utils.PacketUtils;
 import net.minestom.server.utils.async.AsyncUtils;
 import net.minestom.server.utils.block.BlockIterator;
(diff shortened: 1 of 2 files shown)


TheMode:
Could you explain why you modified the constructor to use `int`s?

Obyvante:
> Could you explain why you modified the constructor to use `int`s?

Here we multiply with 8000 / server tps and the developer has to predicate it is not surpassing the short maximum and minimum(which is minimum: -81.9175 and maximum +81.9175). To avoid negative/unpredictable values, I use int to handle any value and then clamp it to make it safer for the client.

I am sure no one wants to clamp their math result for velocity before setting the entity's velocity every time. If you set to for example 83 instead of 81.91, the client will see nothing moving since it becomes a negative maximum.

Please let me know if I miss something and thank you for your time!

https://github.com/Minestom/Minestom/blob/e5d0a43ba42c7023a93e8e33356909e614783ac7/src/main/java/net/minestom/server/entity/Entity.java#L1529-L1531

TheMode:
The change should then be in the code responsible for sending the packet, not in the packet itself.

Obyvante:
> The change should then be in the code responsible for sending the packet, not in the packet itself. Adding some magic bound in the velocity packet would be even more confusing (especially since Java already has the perfect type - a short). As people could expect it to effectively take an `int`

Hmm. So should I move clamping to "getVelocityPacket" instead of to the packet itself?

TheMode:
Yes

TheMode:
Or you could move it to the packet constructor taking the `Point`

src/main/java/net/minestom/server/network/packet/server/play/EntityVelocityPacket.java
```diff
@@ -11,6 +11,7 @@
 
 public record EntityVelocityPacket(int entityId, short velocityX, short velocityY,
                                    short velocityZ) implements ServerPacket {
+
```
TheMode, line 14: Unnecessary line change

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1527,7 +1528,10 @@ private void scheduleRemove(TaskSchedule schedule) {
     }
 
     protected @NotNull Vec getVelocityForPacket() {
-        return this.velocity.mul(8000f / MinecraftServer.TICK_PER_SECOND);
+        return this.velocity.mul(8000f / MinecraftServer.TICK_PER_SECOND)
+                .withX(x -> MathUtils.clamp(x, Short.MIN_VALUE, Short.MAX_VALUE))
+                .withY(y -> MathUtils.clamp(y, Short.MIN_VALUE, Short.MAX_VALUE))
+                .withZ(z -> MathUtils.clamp(z, Short.MIN_VALUE, Short.MAX_VALUE));
```
TheMode, line 1534: You are creating 3 new `Vec` objects here

TheMode (changes requested):
Performance concerns.
Then as said you could instead move this to the constructor requiring a `Point`. As it could indeed still create wrong packets with negative values

src/main/java/net/minestom/server/network/packet/server/play/EntityVelocityPacket.java
```diff
@@ -1,9 +1,11 @@
 package net.minestom.server.network.packet.server.play;
 
+import net.minestom.server.MinecraftServer;
```
TheMode, line 3: Unused as well

PR #1564, opened 2022-12-10
Allow player heads to be loaded without explicitly defining owner

Etrayed:
The vanilla client can only render player head's with textures if the `SKULL_OWNER` property is set.

Problem:

```java
new PlayerHeadMeta.Builder().playerSkin(...).build(); // does not render client-side

new PlayerHeadMeta.Builder().skullOwner(UUID.randomUUID()).playerSkin(...).build(); // renders client-side
```

But often you just want to set the texture data without explicitly defining the owner UUID of the head, so I tried to remove this boilerplate snippet by just setting a random skull owner if needed.

### DIFF
--- a/src/main/java/net/minestom/server/item/metadata/PlayerHeadMeta.java
+++ b/src/main/java/net/minestom/server/item/metadata/PlayerHeadMeta.java
@@ -66,6 +66,10 @@ public record PlayerHeadMeta(TagReadable readable) implements ItemMetaView<Playe
         }
 
         public Builder playerSkin(@Nullable PlayerSkin playerSkin) {
+            if (getTag(SKULL_OWNER) == null) {
+                setTag(SKULL_OWNER, UUID.randomUUID());
+            }
+
             setTag(SKIN, playerSkin);
             return this;
         }


Etrayed:
I dont know why the checkstyle has failed but I dont think it's my fault lol

TheMode:
I do not think that we want to arbitrarily set some tags. Although I agree that having to set the skull owner feels unnecessary

Etrayed:
> I do not think that we want to arbitrarily set some tags. Although I agree that having to set the skull owner feels unnecessary

I don't really understand the point you're trying to make, as the "arbitrary" tag is only being set if another tag (the `SKULL_OWNER` one) is about to be set. Also both "tags" are part of the same NBT-struct which is going to be sent to the client, so no unnecessary allocations there! It would just be a niche way to remove some boilerplate.

Etrayed:
Found one bug with changing skull owners repeatedly, will try to fix that first

Etrayed:
Nvm the bug was not related to this PR

mworzala:
This is no longer relevant, head rendering is based entirely on the PROFILE component which just contains a texture.

PR #1591, opened 2022-12-18
Minor Property Tweaks

### LINKED ISSUE #1171: World border max radius is hardcoded
Apparently the max radius of the world border is sent in a packet, but wiki.vg describes it incorrectly as "Portal teleport boundary" and with a note of "Resulting coordinates from a portal teleport are limited to ±value. Usually 29999984." (https://wiki.vg/Protocol#Initialize_World_Border)

Minestom hard-codes this as 29999984, but it should have a way to set (you can set it on a vanilla server with the max-world-size option in server.properties) it.
(https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/instance/WorldBorder.java#L42=)

Setting the max radius of the world border can be abused to create non-square world borders, hence why it would be useful to set.

_Originally posted by @dev-hydrogen in https://github.com/Minestom/Minestom/discussions/1170#discussioncomment-2977168_

GreatWyrm:
This PR:

- Adds the ability to customize the maximum border with a System Property (`minestom.world-border-size`)
- Changes the properties call in ExtensionManager, as there is a native Java method to supply a default value when retrieving a System Property
- Changes the `getString` method in PropertyUtils to use the native Java method.

Closes #1171.

### DIFF
--- a/src/main/java/net/minestom/server/extensions/ExtensionManager.java
+++ b/src/main/java/net/minestom/server/extensions/ExtensionManager.java
@@ -37,7 +37,7 @@ public class ExtensionManager {
     private final Map<String, Extension> extensions = new LinkedHashMap<>();
     private final Map<String, Extension> immutableExtensions = Collections.unmodifiableMap(extensions);
 
-    private final File extensionFolder = new File(PropertyUtils.getString("minestom.extension.folder", "extensions"));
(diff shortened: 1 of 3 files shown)


TheMode:
I do not think there is a need to remove the `getString` method. You could however change its impl to use the jdk version

mworzala (approved):

Issue #1592, opened 2022-12-18
Get all registered listeners


Distractic:
# Context

In my project, I need to retrieve all registered listeners from GlobalEventHandler.
We can add or remove but it's not possible to get the listeners.

Is it possible to implement this function ?

mworzala:
What is the use case, this does not seem like a good feature to me.

Distractic:
> What is the use case, this does not seem like a good feature to me.

I need to check if exactly a list of listeners is registered. With the exacte number of listeners, same type etc.

mworzala:
(kind of a necropost, I know)
We generally suggest not adding/removing listeners a lot and instead filtering what you need with an EventFilter on the nodes or listeners.

PR #1593, opened 2022-12-18
feat: Add method to retrieve all listeners from node

Distractic:
Implementation for #1592

### DIFF
--- a/src/main/java/net/minestom/server/event/EventNode.java
+++ b/src/main/java/net/minestom/server/event/EventNode.java
@@ -337,8 +337,7 @@ public sealed interface EventNode<T extends Event> permits EventNodeImpl {
      * Get all listeners registered to this node.
      * @return A collection of listeners.
      */
-    @Contract(value = "-> this")
-    @NotNull Collection<EventListener<T>> getListeners();
+    @NotNull Collection<EventListener<T>> listeners();
 
     /**
      * Maps a specific object to a node.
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/event/EventNode.java
```diff
@@ -332,6 +333,13 @@ default void removeChildren(@NotNull String name) {
     @Contract(value = "_ -> this")
     @NotNull EventNode<T> removeListener(@NotNull EventListener<? extends T> listener);
 
+    /**
+     * Get all listeners registered to this node.
+     * @return A collection of listeners.
+     */
+    @Contract(value = "-> this")
+    @NotNull Collection<EventListener<T>> getListeners();
```
TheMode, line 341: The `Contract` is wrong. Also `listeners()` would be a better name

TheMode (changes requested):
I am not sure to be convinced, especially given the slowness of the getter

src/main/java/net/minestom/server/event/EventNode.java
Distractic: I removed the contract and change the method's name.

Distractic:
> I am not sure to be convinced, especially given the slowness of the getter

Sure, this method should be called exceptionally and it's my use case
After, if you have an idea to get listeners with a faster implementation, I'm interested

NitrinCloud:
> > I am not sure to be convinced, especially given the slowness of the getter
> 
> Sure, this method should be called exceptionally and it's my use case After, if you have an idea to get listeners with a faster implementation, I'm interested

You could use 2 for loops instead of streams

mworzala:
Closing with the issue, can be reopened later if more discussion happens

Issue #1594, opened 2022-12-18
Sound packets seeds not randomized


Am-Gone:
In AdventurePacketConvertor, methods used to create a sound packet always use the seed 0 so if you play sounds like `entity.zombie.hurt` which are currently 2 different sounds (or more, depending on what you play), the client will always play the same one

hapily04:
I fixed this in https://github.com/hollow-cube/minestom-ce/pull/74

TheMeinerLP:
@hapily04 Can you pull back the changes ?

GodModed:
It seems like there is a bit of confusion. This bug has already been resolved. This issue can be closed.

PR #1595, opened 2022-12-18
Randomize sound seed in AdventurePacketConvertor

Am-Gone:
Fix for #1594

### DIFF
--- a/src/main/java/net/minestom/server/adventure/AdventurePacketConvertor.java
+++ b/src/main/java/net/minestom/server/adventure/AdventurePacketConvertor.java
@@ -19,12 +19,14 @@ import net.minestom.server.utils.TickUtils;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Collection;
+import java.util.Random;
 
 /**
  * Utility methods to convert adventure enums to their packet values.
  */
 public class AdventurePacketConvertor {
     private static final Object2IntMap<NamedTextColor> NAMED_TEXT_COLOR_ID_MAP = new Object2IntArrayMap<>(16);
+    private static final Random SOUND_SEED_RANDOM = new Random();
 
     static {
         NAMED_TEXT_COLOR_ID_MAP.put(NamedTextColor.BLACK, 0);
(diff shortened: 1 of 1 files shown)


NitrinCloud:
Great initial idea, but when we just randomize the sound we can't decide which sound to use. You should consider using the seed provided by the given sound and in my opinion, if it isn't given, you should randomize it.

For example like so...
```java 
sound.seed().orElse(SOUND_SEED_RANDOM.nextLong())
```

Am-Gone:
In my opinion this is a pretty bad idea as this is a seed, not the actual played value. Also, if you really want to use a specific seed, you can still create the packet yourself.

NitrinCloud:
From what I know, every sound in minecraft has a spefic seed and the seed which is send to the client will then be translated to the nearest seed from the sound.

It would be unecessary to create your own paket since the seed would be just optional. Since you can specify the seed for the sound, everyone would exspect (as default behaviour) the seed to be the same as specified.

Am-Gone:
That is not what I found... For me, the seed only helps determining which sound file must be used. Also, the original vanilla code always use the level random source.

NitrinCloud:
Take a look at [this](https://github.com/Owen1212055/mc-sound-seeds) repository. If the client would randomize the sound anyway, you didn't even have to send a random seed

Am-Gone:
Well yes this is 

> In my opinion this is a pretty bad idea as this is a seed, not the actual played value. Also, if you really want to use a specific seed, you can still create the packet yourself.

I still think that it should stay like this. What you've sent is a compilation of seeds that were manually found that plays a specific sounds and is a very specific use-case. Also, this is not version consistent and you can't pull this kind of data off the registry, which means no data generator.

NitrinCloud:
Even tho you are right, there is no real downside in using the provided seed if specified. Like I said, most users will assume that this is the case

hapily04:
Completed [here](https://github.com/hollow-cube/minestom-ce/pull/74) and was merged into this repo with the Minestom-ce -> Minestom massive merge

PR #1610, opened 2022-12-22
Add a Permission Condition + Example Code

FrederikHeinrich:
image
after run /permissionsremove
image


GoldenStack:
You can just have /permissions <check|add|remove> instead of three different commands

src/main/java/net/minestom/server/permission/Permission.java
```diff
@@ -59,6 +59,20 @@ public NBTCompound getNBTData() {
...
+     * @param sender
+     * @param commandString
+     * @return whether the sender has the authorization to execute the command.
+     */
+    public boolean matchCondition(@NotNull CommandSender sender, @Nullable String commandString) {
+        if (!(sender.hasPermission(getPermissionName()))) {
+            return false;
+        }
+        return true;
+    }
```
GoldenStack, line 74: You can just `return sender.hasPermission(getPermissionName());` instead of this unnecessary code

GoldenStack: Also is this method really necessary?

FrederikHeinrich: ohh i just deleted the noPermissions Message ;) so this code is real stupid xD

> You can just have /permissions <check|add|remove> instead of three different commands

I have to check how Sub Commands works.

FrederikHeinrich:
can we mearge it? or do you have any questions?

mworzala:
Permissions are removed as of #2445

PR #1640, opened 2022-12-30
Update to JoinGamePacket and RespawnPacket

Protonull:
This PR does two things of substance: adds support for death locations + fixes a deserialisation bug

### DIFF
--- a/src/main/java/net/minestom/server/network/NetworkBufferTypes.java
+++ b/src/main/java/net/minestom/server/network/NetworkBufferTypes.java
@@ -7,6 +7,7 @@ import net.minestom.server.coordinate.Vec;
 import net.minestom.server.entity.Entity;
 import net.minestom.server.item.ItemStack;
 import net.minestom.server.item.Material;
+import net.minestom.server.network.packet.server.play.data.DeathLocation;
 import net.minestom.server.utils.Direction;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.UnknownNullability;
@@ -504,6 +505,24 @@ final class NetworkBufferTypes {
             buffer -> {
                 final int ordinal = buffer.read(VAR_INT);
                 return Entity.Pose.values()[ordinal];
+            });
+    static final TypeImpl<DeathLocation> DEATH_LOCATION = new TypeImpl<>(DeathLocation.class,
+            (buffer, value) -> {
+                if (value == null) {
+                    buffer.write(BOOLEAN, false);
+                }
+                else {
+                    buffer.write(BOOLEAN, true);
+                    buffer.write(STRING, value.dimension());
+                    buffer.write(BLOCK_POSITION, value.position());
+                }
+                return -1;
+            },
+            buffer -> {
+                if (buffer.read(BOOLEAN)) {
+                    return new DeathLocation(buffer.read(STRING), buffer.read(BLOCK_POSITION));
+                }
+                return null;
             });
 
     record TypeImpl<T>(@NotNull Class<T> type,
--- a/src/main/java/net/minestom/server/network/NetworkBufferTypes.java
+++ b/src/main/java/net/minestom/server/network/NetworkBufferTypes.java
@@ -510,8 +510,7 @@ final class NetworkBufferTypes {
             (buffer, value) -> {
                 if (value == null) {
                     buffer.write(BOOLEAN, false);
-                }
-                else {
+                } else {
                     buffer.write(BOOLEAN, true);
                     buffer.write(STRING, value.dimension());
                     buffer.write(BLOCK_POSITION, value.position());
--- a/src/main/java/net/minestom/server/network/packet/server/play/JoinGamePacket.java
+++ b/src/main/java/net/minestom/server/network/packet/server/play/JoinGamePacket.java
@@ -1,5 +1,6 @@
 package net.minestom.server.network.packet.server.play;
 
+import net.minestom.server.coordinate.Point;
 import net.minestom.server.entity.GameMode;
 import net.minestom.server.network.NetworkBuffer;
 import net.minestom.server.network.packet.server.ServerPacket;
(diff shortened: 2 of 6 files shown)


src/main/java/net/minestom/server/network/packet/server/play/JoinGamePacket.java
```diff
@@ -54,12 +87,23 @@ public void write(@NotNull NetworkBuffer writer) {
...
+            writer.write(BLOCK_POSITION, deathLocation().position());
+        }
     }
 
     @Override
     public int getId() {
         return ServerPacketIdentifier.JOIN_GAME;
     }
 
+    public record DeathLocation(@NotNull String dimension, @NotNull Point position) {
```
TheMode, line 105: I believe that the same is present in the respawn packet

TheMode:
Also there is no need to change the record formatting

Protonull:
> Also there is no need to change the record formatting

I did give reasoning in the commit message

src/main/java/net/minestom/server/network/packet/server/play/JoinGamePacket.java
Protonull: Just looked at the protocol spec and you're correct. I'm thinking then that `DeathLocation` should be relocated into the `net.minestom.server.network.packet.server.play.data` package. Are these two instances enough to warrant a `NetworkBufferType`, or would a static parser be enough?

TheMode: A new network type seems fine, we can always remove it down the line

TheMode:
Still do not think that it should be done, at least not here

src/main/java/net/minestom/server/network/packet/server/play/JoinGamePacket.java
Protonull: Working on it now... I don't really know what all the `return -1;` signify in regards to the type writers though.

Protonull:
> Still do not think that it should be done, at least not here

Fair enough, I'll revert the commit. I do hope there's an effort to reorganise each of the packets this way though since it really was a hassle trying to determine which of the fields was causing the problem.

Anyway, while I'm at it, I'd like guidance on how to solve another problem. The protocol spec states that the previous gamemode value can be `-1` to signify no previous gamemode. The problem is that `GameMode.fromId(-1)` will throw an exception because it's only setup to return a valid `GameMode` enum value, a non-null value. Currently `JoinGamePacket ` will correctly serialise a null previous `GameMode` but will throw attempting to deserialise one. How would you prefer this be fixed? I'm thinking to either update the `GameMode.fromId` method, add another static method to it, such as `GameMode.fromNullableId()`, or doing some witchery during the network read.

TheMode:
Wouldnt making the gamemode nullable work?

Protonull:
> Wouldnt making the gamemode nullable work?

That can work. I didn't want to edit something like that, particularly if it means making a returned value nullable, without consulting you first.

src/main/java/net/minestom/server/network/packet/server/play/JoinGamePacket.java
TheMode: This is used to signed the buffer how much to advance the writer index, `-1` indicate nothing/ignore as using the `write` method inside the lambda already advance the index

Protonull:
Hmmm, `RespawnPacket` has a lot of instantiations, so there's a few empty `DeathLocation` variables lying around now. They're all commented with `// TODO: Add some way to determine last death location` so they're extremely easy to find with a search tool. Should I use this PR as an opportunity to make some kind of DeadLocationGetter or some sort, or leave that to another PR?

src/main/java/net/minestom/server/network/NetworkBufferTypes.java
```diff
@@ -505,6 +506,24 @@ public int available() {
                 final int ordinal = buffer.read(VAR_INT);
                 return Entity.Pose.values()[ordinal];
             });
+    static final TypeImpl<DeathLocation> DEATH_LOCATION = new TypeImpl<>(DeathLocation.class,
+            (buffer, value) -> {
+                if (value == null) {
+                    buffer.write(BOOLEAN, false);
+                }
+                else {
```
TheMode, line 514: Same line as the bracket

src/main/java/net/minestom/server/network/packet/server/play/data/DeathLocation.java
```diff
@@ -0,0 +1,11 @@
+package net.minestom.server.network.packet.server.play.data;
+
+import net.minestom.server.coordinate.Point;
+import org.jetbrains.annotations.NotNull;
+
+public record DeathLocation(
+        @NotNull String dimension,
+        @NotNull Point position
+) {
```
TheMode, line 9: Wrong formatting + empty line below

src/main/java/net/minestom/server/entity/GameMode.java
```diff
@@ -29,8 +27,9 @@ public boolean canTakeDamage() {
         return canTakeDamage;
     }
 
-    public static @NotNull GameMode fromId(int id) {
+    public static GameMode fromId(int id) {
```
TheMode, line 30: We may want to replace this method with a gamemode network type

TheMode (changes requested):

TheMode:
> Hmmm, `RespawnPacket` has a lot of instantiations, so there's a few empty `DeathLocation` variables lying around now. They're all commented with `// TODO: Add some way to determine last death location` so they're extremely easy to find with a search tool. Should I use this PR as an opportunity to make some kind of DeadLocationGetter or some sort, or leave that to another PR?

We can leave it to another PR. The most important part is having the packets

src/main/java/net/minestom/server/network/NetworkBufferTypes.java
Protonull: You'd prefer `if (value == null) buffer.write(BOOLEAN, false);` ?

TheMode: The `else {` is in a different line, definitely confusing

Protonull: Ah, you're one of *those* :P I'll fix

src/main/java/net/minestom/server/network/packet/server/play/data/DeathLocation.java
Protonull: A little confused as to what you mean. I assuming you'd prefer the formatting of the fields to be inline rather than line separated, but I'm confused as to what you mean by an empty line below. You'd prefer an empty record body to be inline too?

TheMode: I was referring to the empty line just below L10. Not that everything should be inlined

Protonull:
Have done everything as you've requested, but I've been looking through the spec and I'm having second thoughts about creating a network buffer type for `GameMode`. Game modes are represented in the spec as [unsigned] bytes, but also as varints, and floats. At that point, it's probably better just to change the static getter method, or create one just for this scenario.

Protonull:
Dropped all the commits that were reverted. The nullable gamemode is now parsed via a private getter on the `JoinGamePacket` record itself.

PR #1645, opened 2023-01-01
Add getPath function to the anvilloader

DevBlokJochem:
You can now get the path from the anvilloader. This mean that when you set an anvilloader to an instance, you can get the path from the "world" everytime you want.

### DIFF
--- a/src/main/java/net/minestom/server/instance/AnvilLoader.java
+++ b/src/main/java/net/minestom/server/instance/AnvilLoader.java
@@ -361,7 +361,7 @@ public class AnvilLoader implements IChunkLoader {
         return blockStateId2ObjectCacheTLS.get().computeIfAbsent(block.stateId(), _unused -> new BlockState(block.name(), block.properties()));
     }
 
-    private Path getPath() {
+    public Path getPath() {
         return this.path;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/instance/AnvilLoader.java
```diff
@@ -361,6 +361,10 @@ private BlockState getBlockState(final Block block) {
         return blockStateId2ObjectCacheTLS.get().computeIfAbsent(block.stateId(), _unused -> new BlockState(block.name(), block.properties()));
     }
 
+    public Path getPath() {
```
mworzala, line 364: @NotNull

mworzala (commented):
My general concern here is that we are exposing a detail which may change in the future (requiring us to deprecate/maintain compatibility). For example, if AnvilLoader is changed to support taking some function of `String -> InputStream` (region name -> content) eg to support non-filesystem loading. In this case there would no longer be a path to return directly.

src/main/java/net/minestom/server/instance/AnvilLoader.java
DevBlokJochem: But normal world files would still be supported. 
What if I change this to a nullable function? If you use the normal world files, it return a path. But if you use a non-filesystem world, it returns null.

mworzala: Nullable + a javadoc stating that it is implementation dependent (even though it is not right now) would be an improvement IMO. But can also wait to see what others say.

DevBlokJochem: The anvil loader extends the IChunkLoader. Maybe it is better to create another class that extends the IChunkLoader too. That class can use the non-filesystem. (if you want to make a new chunkloader)

DevBlokJochem:
^^ The latest merge wasn't planned. I just wanted to fork it again :)

PR #1650, opened 2023-01-02
Hand in PlayerItemAnimationEvent

HEROOSTECH:
Hi, i am creating disguise extension and i am at translating item animation. It's impossible to get hand from the event and i don't wanna use packet (Especially since this event can be cancelled), so i did small changes to minestom that adds hand to the event.

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerItemAnimationEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerItemAnimationEvent.java
@@ -37,7 +37,7 @@ public class PlayerItemAnimationEvent implements PlayerInstanceEvent, Cancellabl
      *
      * @return the hand
      */
-    public Player.Hand getHand() {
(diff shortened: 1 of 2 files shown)


mworzala (approved):

src/main/java/net/minestom/server/event/player/PlayerItemAnimationEvent.java
```diff
@@ -31,6 +32,15 @@ public PlayerItemAnimationEvent(@NotNull Player player, @NotNull ItemAnimationTy
         return itemAnimationType;
     }
 
+    /**
+     * Gets the hand that was used.
+     *
+     * @return the hand
+     */
+    public Player.Hand getHand() {
```
mworzala, line 40: Should be annotated with @NotNull

mworzala (changes requested):

mworzala (approved):

PR #1658, opened 2023-01-05
Prevent block place above 320

### LINKED ISSUE #867: Exception when placing blocks above the height limit
Placing a block above the dimension height limit causes an exception:
```
java.lang.ArrayIndexOutOfBoundsException: Index 24 out of bounds for length 24
	at java.base/java.util.ImmutableCollections$ListN.get(ImmutableCollections.java:680)
	at net.minestom.server.instance.DynamicChunk.getSection(DynamicChunk.java:109)
	at net.minestom.server.instance.Chunk.getSectionAt(Chunk.java:93)
	at net.minestom.server.instance.DynamicChunk.setBlock(DynamicChunk.java:71)
	at net.minestom.server.instance.InstanceContainer.UNSAFE_setBlock(InstanceContainer.java:131)
	at net.minestom.server.instance.InstanceContainer.placeBlock(InstanceContainer.java:166)
	at net.minestom.server.listener.BlockPlacementListener.listener(BlockPlacementListener.java:133)
	at net.minestom.server.listener.manager.PacketListenerManager.processClientPacket(PacketListenerManager.java:88)
```

This should be easy to fix by checking if the placed block is within the world limits inside of `BlockPlacementListener`.

Nuckerr:
This checks if the position of the block is greater or equal to 320 and then prevents it's placement if it does

Closes #867

### DIFF
--- a/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
+++ b/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
@@ -83,11 +83,14 @@ public class BlockPlacementListener {
             canPlaceBlock = usedItem.meta().canPlaceOn(interactedBlock);
         }
 
+
         // Get the newly placed block position
         final int offsetX = blockFace == BlockFace.WEST ? -1 : blockFace == BlockFace.EAST ? 1 : 0;
         final int offsetY = blockFace == BlockFace.BOTTOM ? -1 : blockFace == BlockFace.TOP ? 1 : 0;
         final int offsetZ = blockFace == BlockFace.NORTH ? -1 : blockFace == BlockFace.SOUTH ? 1 : 0;
         final Point placementPosition = blockPosition.add(offsetX, offsetY, offsetZ);
+
+        if(placementPosition.y() >= instance.getDimensionType().getMaxY()) return;
 
         if (!canPlaceBlock) {
             // Send a block change with the real block in the instance to keep the client in sync,
@@ -127,8 +130,6 @@ public class BlockPlacementListener {
             refresh(player, chunk);
             return;
         }
-
-        if(placementPosition.y() >= instance.getDimensionType().getMaxY()) return;
 
         // BlockPlacementRule check
         Block resultBlock = playerBlockPlaceEvent.getBlock();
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/listener/BlockPlacementListener.java
```diff
@@ -131,6 +131,7 @@ public static void listener(ClientPlayerBlockPlacementPacket packet, Player play
         // BlockPlacementRule check
         Block resultBlock = playerBlockPlaceEvent.getBlock();
         final BlockPlacementRule blockPlacementRule = BLOCK_MANAGER.getBlockPlacementRule(resultBlock);
+        if(placementPosition.y() >= 320) return;
```
TogAr2, line 134: You should probably not hardcode the number, use the dimension to get the maximum y value. Also you should probably move this to just after the initialization of `placementPosition` for a fast exit

src/main/java/net/minestom/server/listener/BlockPlacementListener.java
```diff
@@ -28,6 +28,7 @@
 
 public class BlockPlacementListener {
     private static final BlockManager BLOCK_MANAGER = MinecraftServer.getBlockManager();
+    private static final double BLOCK_HEIGHT = 320.0;
```
TogAr2, line 31: This isn't constant for every instance, use `instance.getDimensionType().getMaxY()`

src/main/java/net/minestom/server/listener/BlockPlacementListener.java
```diff
@@ -128,6 +129,8 @@ public static void listener(ClientPlayerBlockPlacementPacket packet, Player play
             return;
         }
 
+        if(placementPosition.y() >= BLOCK_HEIGHT) return;
```
TogAr2, line 132: I meant moving it to line 92, in which case the events would not be called, but in my opinion they shouldn't be in this case

TheMode:
Could you add a test?

Nuckerr:
Which package/file would you reccomend?

TheMode:
Both the instance & entity package would work

Nuckerr:
The unit test fails because set block also triggers this error. Where should I add this check

Nuckerr:
bump

mworzala:
Merged in with minestom-ce in #1937

PR #1661, opened 2023-01-06
Support dynamic selection of a port using '0'

NitrinCloud:
> A port number of zero will let the system pick up an ephemeral port in a bind operation.  

Since minestom just saves the port before the binding process this possibility will never affect the port value, thus the actual port and the saved port are not the same.


NitrinCloud:
@TheMode Just checked the test. Do you want that the field 'port' is 0 when the system selects a port? If so, wouldn't it be a good idea to being able to access both from other classes.

PR #1670, opened 2023-01-08
Use MinestomDataGenerator updated for 1.19.3

leNicDev:
This PR updates the MinestomDataGenerator dependency to the one updated for 1.19.3.


gradle/libs.versions.toml
```diff
@@ -7,7 +7,8 @@ adventure = "4.11.0"
 kotlin = "1.6.20"
 hydrazine = "1.7.2"
 dependencyGetter = "v1.0.1"
-minestomData = "76da04ca1d"
+minestomData = "1.19.3-SNAPSHOT"
```
TheMode, line 10: Do not use snapshot builds. Also no need to comment out the previous version

leNicDev: Fixed.

PR #1682, opened 2023-01-11
Add ParticleOptions API

devPesto:
ParticleOptions enforces that valid data is written to particle packets so that the client is not kicked for recieving malformed packets


src/main/java/net/minestom/server/particle/ParticleCreator.java
```diff
@@ -23,14 +30,26 @@ public static ParticlePacket createParticlePacket(Particle particleType, boolean
...
     }
 
-    public static ParticlePacket createParticlePacket(Particle particleType,
+
+    public static ParticlePacket createParticlePacket(Particle particleType, boolean distance,
+                                                      double x, double y, double z,
+                                                      float offsetX, float offsetY, float offsetZ,
+                                                      float particleData, int count, ParticleOptions options) {
+
+        Check.argCondition(particleType != options.type(), "ParticleOptions mismatch!");
```
GreatWyrm, line 43: I'd recommend this error message be a bit more clear, something lie `"Particle type " + particleType.namespace() + " does not support ParticleOptions " + options != null ? options.getClass().getName() : "no options";`

Also, you should probably have some `@Nullable` and `@NotNull` annotations on the method parameters.

PestoNotPasta: I improved the error message and enforced `@NotNull` arguments but I don't think `@Nullable` would make sense here as people could user the other method for particles that don't require options.

src/main/java/net/minestom/server/particle/ParticleOptions.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    public Particle type() {
+        return type;
+    }
+    
+    /**
+     * Options for {@link Particle#DUST}
+     */
+    public static final class Dust extends ParticleOptions {
```
KrystilizeNevaDies, line 33: All these options should be records, duplicating a bit of code here because of records is fine.

src/main/java/net/minestom/server/particle/ParticleOptions.java
```diff
@@ -0,0 +1,148 @@
...
+import net.minestom.server.entity.Entity;
+import net.minestom.server.item.ItemStack;
+import net.minestom.server.utils.MathUtils;
+import net.minestom.server.utils.binary.BinaryWriter;
+import net.minestom.server.utils.validate.Check;
+import org.jetbrains.annotations.NotNull;
+
+import static net.minestom.server.particle.ParticleOptions.*;
+
+public sealed abstract class ParticleOptions permits BlockStateOptions, Dust, DustColorTransition, Item, SculkCharge, Shriek, Vibration {
```
KrystilizeNevaDies, line 14: This should be an interface, and the records should have an explicit type function (do not put the type function in the record constructor.)

KrystilizeNevaDies:
LGTM

src/main/java/net/minestom/server/particle/ParticleOptionImpl.java
```diff
@@ -0,0 +1,392 @@
...
+import net.minestom.server.entity.Entity;
+import net.minestom.server.instance.block.Block;
+import net.minestom.server.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+import static net.minestom.server.particle.ParticleOption.*;
+
+public interface ParticleOptionImpl {
+
+    static ParticleOption AMBIENT_ENTITY_EFFECT() {
```
Moulberry, line 15: I'd prefer using camelCase instead of SCREAMING_SNAKE_CASE for these method names

src/main/java/net/minestom/server/particle/ParticleCreator.java
```diff
@@ -23,14 +31,27 @@ public static ParticlePacket createParticlePacket(Particle particleType, boolean
...
     }
 
-    public static ParticlePacket createParticlePacket(Particle particleType,
+
+    public static ParticlePacket createParticlePacket(@NotNull Particle particleType, boolean distance,
+                                                      double x, double y, double z,
+                                                      float offsetX, float offsetY, float offsetZ,
+                                                      float particleData, int count, ParticleOption options) {
+
+        Check.argCondition(particleType != options.type(), "Particle type ''{0}'' does not support Particle options ''{1}''",
```
Moulberry, line 44: This Check.argCondition is pointless. Remove the `Particle particleType` parameter and just get the type via `options.type()`

src/main/java/net/minestom/server/particle/ParticleCreator.java
```diff
@@ -23,14 +31,27 @@ public static ParticlePacket createParticlePacket(Particle particleType, boolean
...
+                particleType.namespace(), options.getClass().getSimpleName());
+        return new ParticlePacket(particleType.id(), distance, x, y, z, offsetX, offsetY, offsetZ, particleData, count, options);
+    }
+
+
+    public static ParticlePacket createParticlePacket(Particle particleType, boolean distance,
```
Moulberry, line 50: This method is bad because you can still cause client crashes by giving Particle.DUST, for example.
Instead, the previous method should be used which ensures that you give options when they are needed.

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
```diff
@@ -3,21 +3,39 @@
...
+                             byte[] data) implements ServerPacket {
+
     public ParticlePacket(@NotNull NetworkBuffer reader) {
         this(reader.read(VAR_INT), reader.read(BOOLEAN),
                 reader.read(DOUBLE), reader.read(DOUBLE), reader.read(DOUBLE),
                 reader.read(FLOAT), reader.read(FLOAT), reader.read(FLOAT),
                 reader.read(FLOAT), reader.read(INT), reader.read(RAW_BYTES));
     }
 
+    public ParticlePacket(int particleId, boolean longDistance, double x, double y, double z,
```
Moulberry, line 27: I would recommend against having ParticleOption inside ParticlePacket.
ParticleOption should be part of the ParticleCreator API, not the network protocol.

I'd prefer it if the original packet definition is used (with byte[] data) and the byte[] data is calculated in ParticleCreator#createParticlePacket

src/main/java/net/minestom/server/particle/ParticleOptionImpl.java
```diff
@@ -0,0 +1,392 @@
...
+import net.minestom.server.color.Color;
+import net.minestom.server.coordinate.Point;
+import net.minestom.server.entity.Entity;
+import net.minestom.server.instance.block.Block;
+import net.minestom.server.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+
+import static net.minestom.server.particle.ParticleOption.*;
+
+public interface ParticleOptionImpl {
```
Moulberry, line 13: Why is this called ParticleOptionImpl?
Generally, classes ending in Impl shouldn't be exposed publicly at all.
Perhaps `ParticleOptions` would be more fitting?

Moulberry (changes requested):

src/main/java/net/minestom/server/particle/ParticleOption.java
```diff
@@ -0,0 +1,760 @@
...
+
+    Particle type();
+
+    default void write(@NotNull BinaryWriter writer) {
+
+    }
+
+    record AMBIENT_ENTITY_EFFECT() implements ParticleOption {
+        @Override
```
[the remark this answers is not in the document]
devPesto: Done

src/main/java/net/minestom/server/particle/ParticleOption.java
```diff
@@ -0,0 +1,760 @@
...
+        public Particle type() {
+            return Particle.SCULK_SOUL;
+        }
+    }
+
+    record SCULK_CHARGE(float angle) implements ParticleOption {
+
+        public SCULK_CHARGE {
+            Check.argCondition(MathUtils.isBetween(angle, 0, 2 * Math.PI), "Angle is not within bounds");
```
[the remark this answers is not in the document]
devPesto: Forgot to remove this

src/main/java/net/minestom/server/particle/ParticleCreator.java
devPesto: Done

src/main/java/net/minestom/server/particle/ParticleCreator.java
devPesto: Done

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
devPesto: Done

src/main/java/net/minestom/server/particle/ParticleOptionImpl.java
devPesto: I think so too. Renamed

mworzala:
Superseded by https://github.com/Minestom/Minestom/pull/2004

PR #1696, opened 2023-01-18
Apply GameProfile to players when using Mojang Authenication

### LINKED ISSUE #1695: Player#getSkin() is always null when using MojangAuth.init()
When using `MojangAuth.init();`, calling Player#getSkin() will always return null.

I believe the bug is most likely caused by the Game Profile not being set (using the `socketConnection.UNSAFE_setProfile()` method) in the EncryptionResponsePacket's `process` method.

Fixing the bug would go along the lines of adding this code to the end of the client.sendAsync().whenComplete() (after line 106) method

```
List<GameProfile.Property> propertyList = new ArrayList<>();
// Add texture properties
socketConnection.UNSAFE_setProfile(new GameProfile(profileUUID, profileName, propertyList));
```

GreatWyrm:
Resolves #1695. 

Now applies the GameProfile to players when using Mojang Authentication, and cleans up a few todos.

~~Currently a draft PR as I have not tested this yet.~~

Testing complete, this now properly sets the Game Profile for players.

### DIFF
--- a/src/main/java/net/minestom/server/network/packet/client/login/EncryptionResponsePacket.java
+++ b/src/main/java/net/minestom/server/network/packet/client/login/EncryptionResponsePacket.java
@@ -1,7 +1,9 @@
 package net.minestom.server.network.packet.client.login;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import net.kyori.adventure.text.Component;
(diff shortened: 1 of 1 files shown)


GreatWyrm:
This change was merged in with the Minestom-CE pullback, specifically in commit https://github.com/Minestom/Minestom/commit/edf7b870c798a4e88e7afbcd188226d4bfa584af.

PR #1706, opened 2023-01-22
Fix some block events

emortaldev:
i was a silly and forgor to add a getter (TWICE!) :smile:


emortaldev:
now this is really silly!

PR #1713, opened 2023-01-24
Fix for #1712

### LINKED ISSUE #1712: Entity block y-position is misreported in some circumstances
The method `CollisionUtils#handlePhysics(Entity, Vec, PhysicsResult)` sometimes returns `PhysicsResult`s whose new positions are slightly inside the block below, when the entity is moving downwards into a full block.

These new positions are off by less than `Vec#EPSILON`, and they don't result in any obvious issues at first. However, there are a few problems:

* Calling `Point#blockY` on the entity's position will give the position of the block that the entity should be above
* This behavior is inconsistent and depends on the exact starting y-position

I made a unit test that reproduces this behavior (on c995f9c3a94a8206eceeddb83a76e2b88594d2ba):

```java
@Test
public void entityBlockPositionTest(Env env) {
    var instance = env.createFlatInstance();

    var entity = new Entity(EntityType.ZOMBIE);
    entity.setInstance(instance, new Pos(0, 43.01, 0));

    var deltaPos = new Vec(0.0, -10, 0.0);
    var physicsResult = CollisionUtils.handlePhysics(entity, deltaPos, null);

    var newPos = physicsResult.newPosition();
    assertEquals(43, newPos.blockY());
}
```

Curiously, the test passes if the entity's initial y-value is at a greater distance from 43 (for example, 43.1), *or* is equal to 43. 

I think that the entity's y-value should always be accurate when standing on a full block, and in general for block heights that allow an accurate double representation, like half blocks. However, if this is not feasible, the deviation should at least not depend on the entity's starting y-value, as it appears to do now.

Steanky:
Should resolve #1712. Also includes a few extra unit tests to validate the fix.


iam4722202468:
@TheMode LGTM

PR #1736, opened 2023-02-05
Use platform-agonistic theme handling for README

LeoDog896:
GitHub already supports the picture element - view it for yourself!


mworzala:
It seems to work fine as is, please include the problematic platforms + reopen if this is still an issue.

PR #1739, opened 2023-02-05
Add support for relative teleportation

NxDs:
A different and simplified approach to #1273

This PR aims to provide a way to teleport an entity using relative coordinates, the true benefit is actually seen when used on players as setting the flags as relative (see https://wiki.vg/Protocol#Synchronize_Player_Position) has the additional effect of making the client retain its state such as its velocity as opposed to the typical teleportation which resets it.

This is a compromise of the original PR which in my opinion was a bit overcomplicated allowing full control of which coordinates should be considered relative/absolute, I do believe that in the typical use case you'd either decide to specify a position fully absolute or fully relative and not something mixed.

The internal method Entity#synchronizePosition(boolean includeSelf, boolean relativePosition) now has an additional parameter to switch to this new behavior, I'm not a fan but that's the way includeSelf was already handled (only Player actually makes use of it) so I followed along.


DeidaraMC:
Resolved with #2029

PR #306, opened 2021-05-25
Create FencePlacementRule.java

andrew-bilotti:
Works better than wallplacementrule for fences!

PR #1742, opened 2023-02-09
Improve PlacementRules

NxDs:
This PR removes the explicit and manual registration of various placement rules for blocks in the utility class PlacementRules and instead registers all the blocks iteratively providing a more complete and accurate result.

It also separates the placement rule types in different methods so that they can be enabled separately  (i.e. only stairs, only walls, etc)
This PR does not provide the implementation of the various placement rules, just their registration.


NxDs:
Notes: 
This PR should be blocked by #893 as init() here will now also register stairs placement rules which at the current state will throw exceptions
Additionally #893 should be blocked by #443

NxDs:
Notes 2: Ideally this could be improved further with #306

luzzardik:
For info: #893 is based on #443

PR #1748, opened 2023-02-15
Update StairsPlacementRule.java

Jochengehtab:
Hi, i am coding currently a Project with Minestom and i run in the issue, that are these class are updatet 2 years ago. I hope i can change with this on line the Block Facing.


MrGazdag:
this does not account for the placement on the sides of the block, and for those you need to use the cursor position

NxDs:
There is already #893 that fixes stairs placement, although the PR is still waiting for the author to revert the double parameters to a Point

Jochengehtab:
Ok thank you

PR #1784, opened 2023-03-13
Wrong generated window id for newly created inventories.

Obyvante:


iam4722202468:
Could you explain what inputs would cause the previous line, `final byte id = (byte) Math.abs((byte) ID_COUNTER.incrementAndGet());` to return a negative integer?

Also, please write a test for inventories which currently fails, but doesn't fail with your fix. Something like generating 300 inventories, and ensuring none are 0 or negative.

Obyvante:
> Looking at your message in feedback on discord, it does indeed look like -128 will cause an issue with the line `inal byte id = (byte) Math.abs((byte) ID_COUNTER.incrementAndGet());`.
> 
> Please write a test for inventories which currently fails, but doesn't fail with your fix. Something like generating 300 inventories, and ensuring none are 0 or negative.

Thank you for your suggestion. There was a test with 256 times but I do not believe it is enough. Therefore, I have increased it to 1,000 to ensure everything works. I also confirm that it will work with 100,000 (I have tested it in the local environment).

src/test/java/net/minestom/server/inventory/InventoryTest.java
```diff
@@ -78,9 +78,7 @@ public void testAdd() {
 
     @Test
     public void testIds() {
-        for (int i = 0; i <= 256; ++i) {
-            final Inventory inventory = new Inventory(InventoryType.CHEST_1_ROW, "title");
-            assertTrue(inventory.getWindowId() != 0);
-        }
+        for (int i = 0; i <= 1000; ++i)
+            assertTrue(new Inventory(InventoryType.CHEST_1_ROW, "title").getWindowId() != 0);
```
iam4722202468, line 82: This should be `> 0`, not `!=` since negative numbers are also invalid

src/main/java/net/minestom/server/inventory/Inventory.java
```diff
@@ -58,11 +58,12 @@ public Inventory(@NotNull InventoryType inventoryType, @NotNull String title) {
     }
 
     private static byte generateId() {
-        final byte id = (byte) Math.abs((byte) ID_COUNTER.incrementAndGet());
-        if (id == 0) { // zero is player's inventory id
-            return generateId();
+        int id = ID_COUNTER.incrementAndGet();
+        if (id > 127) {
+            ID_COUNTER.set(1);
+            id = 1;
```
iam4722202468, line 64: Start `ID_COUNTER` at 0, then get the ID using `final byte id = (byte)ID_COUNTER.updateAndGet(i -> i == 128 ? 1 : i + 1);`

Setting the value to 1 then doing `id = 1` could potentially be unsafe, if multiple threads are generating an ID at the same time you could get multiple inventories with id = 1.

src/test/java/net/minestom/server/inventory/InventoryTest.java
```diff
@@ -78,9 +78,7 @@ public void testAdd() {
 
     @Test
     public void testIds() {
-        for (int i = 0; i <= 256; ++i) {
-            final Inventory inventory = new Inventory(InventoryType.CHEST_1_ROW, "title");
-            assertTrue(inventory.getWindowId() != 0);
-        }
+        for (int i = 0; i <= 1000; ++i)
+            assertTrue(new Inventory(InventoryType.CHEST_1_ROW, "title").getWindowId() > 0);
```
iam4722202468, line 82: Check if the value is within [1, 127] since this is the valid range

Obyvante: The window id method's return value is a byte so I don't believe we should change this current check.

iam4722202468: alright

src/test/java/net/minestom/server/inventory/InventoryTest.java
```diff
@@ -78,9 +78,9 @@ public void testAdd() {
 
     @Test
     public void testIds() {
-        for (int i = 0; i <= 256; ++i) {
-            final Inventory inventory = new Inventory(InventoryType.CHEST_1_ROW, "title");
-            assertTrue(inventory.getWindowId() != 0);
+        for (int i = 0; i <= 1000; ++i) {
+            final int windowId = new Inventory(InventoryType.CHEST_1_ROW, "title").getWindowId();
```
iam4722202468, line 82: why are you using an `int` to store windowId, when getWindowId returns a byte?

Obyvante: I tried to not touch base variables. The previous contributor used atomic integer. I do not believe there is any atomic byte. Please correct me if I am wrong.

Do you have any better idea to fix this?

iam4722202468: `final byte windowId = new Inventory(InventoryType.CHEST_1_ROW, "title").getWindowId();`

Obyvante: > `final byte windowId = new Inventory(InventoryType.CHEST_1_ROW, "title").getWindowId();`

Oh, you were talking about the test I guess. Window/inventory packets request window id as a byte. That's why I use byte everywhere for window ids.

iam4722202468: it's an integer here though

Obyvante: > in the line you wrote, it's an integer

Oh, I am so sorry, I didn't see it 😢

PR #1802, opened 2023-03-24
Fix method javadoc

Tomasan7:
Fixes a Point#sameBlock(Point) method javadoc. The author copied it over from sameChunk(), but forgot to actually rewrite it.


src/main/java/net/minestom/server/coordinate/Point.java
```diff
@@ -261,12 +261,12 @@ default boolean sameBlock(int blockX, int blockY, int blockZ) {
...
-     * @param point the point to compare two
-     * @return true if 'this' is in the same chunk as {@code point}
+     * @param point the point to compare to
+     * @return true if 'this' is in the same block as {@code point}
      */
     default boolean sameBlock(@NotNull Point point) {
         return sameBlock(point.blockX(), point.blockY(), point.blockZ());
     }
-}
+}
```
TheMode, line 272: Add line break

src/main/java/net/minestom/server/coordinate/Point.java
```diff
@@ -261,12 +261,12 @@ default boolean sameBlock(int blockX, int blockY, int blockZ) {
     }
 
     /**
-     * Gets if two points are in the same chunk.
+     * Gets if two points are in the same block.
```
iam4722202468, line 264: `Checks` instead of `Gets`

Tomasan7: I wanted to stay consistent with other methods, but yes, this is better, I guess we can change all of them.

PR #1810, opened 2023-03-31
Projectile collision improvements

TogAr2:
This pull request intends to fix some issues with the current projectile collision by using the physics engine other entities use.


Obyvante:
Some projectile tests failed. You may want to update your code depending on the test result or take a look at the test code if something is wrong.

TogAr2:
This pr isn’t done yet, entity collision hasn’t been implemented which is why 2/3 tests failed

PR #1833, opened 2023-04-16
Fixes cat color (variant) metadata, adds frog meta

### LINKED ISSUE #1832: Setting a cats colour in CatMeta does not change the cats colour

ZakShearman:
Resolves #1832

Frog and Cat entities have special metadata values for their variants (see <https://wiki.vg/Entity_metadata#Cat>). This uses those types to ensure the `type` parameter of the meta is correct.


emortaldev:
Was thinking cat setColor should be setVariant like most other entities

ZakShearman:
^ probably a good time to change it now as well. It's unlikely others are using this method and if they are, it won't be working for them until this is merged.

PR #1847, opened 2023-05-06
Gradle improvements

mworzala:
Simplify gradle setup and add publishing to maven central


mworzala:
This was an oopsie, it was supposed to be https://github.com/hollow-cube/minestom-ce/pull/8

Issue #1787, opened 2023-03-15
1.19.4 update


MelonHell:

ArmaniS-Dev:
this is a bug list. That doesn't belong here.

mworzala:
Done in #1937

PR #1873, opened 2023-06-07
1.19.4 update

MelonHell:
prev pr #1858
issue #1787


src/main/java/net/minestom/server/entity/metadata/display/AbstractDisplayMeta.java
```diff
@@ -0,0 +1,141 @@
...
+        return super.metadata.getIndex(OFFSET + 2, Vec.ZERO);
+    }
+
+    public void setTranslation(@NotNull Point value) {
+        super.metadata.setIndex(OFFSET + 2, Metadata.Vector3(value));
+    }
+
+    public @NotNull Vec getScale() {
+        return super.metadata.getIndex(OFFSET + 3, Vec.ONE);
+    }
```
tahmid-23, line 46: since Metadata.Vector3 specifies Point, casting would fail if the entry is a Pos
(I'm not really sure if this is an issue)

src/main/java/net/minestom/server/registry/Registry.java
```diff
@@ -410,6 +416,19 @@ public EnchantmentEntry(String namespace, Properties main, Properties custom) {
         }
     }
 
+    public record DamageTypeEntry(NamespaceID namespace, double exhaustion,
```
tahmid-23, line 419: exhaustion is a float

tahmid-23, line 419: effects and death_message_type?

src/main/java/net/minestom/server/network/packet/server/play/UpdateEnabledFeaturesPacket.java
```diff
@@ -0,0 +1,29 @@
...
+import net.minestom.server.network.packet.server.ServerPacketIdentifier;
+import net.minestom.server.utils.NamespaceID;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+import java.util.List;
+
+import static net.minestom.server.network.NetworkBuffer.*;
+
+public record UpdateEnabledFeaturesPacket(Collection<NamespaceID> featureFlags) implements ServerPacket {
```
tahmid-23, line 14: Maybe use a Set here?

src/main/java/net/minestom/server/network/packet/server/play/DamageEventPacket.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * See <a href="https://wiki.vg/Protocol#Damage_Event">https://wiki.vg/Protocol#Damage_Event</a> for more info.
+ *
+ * @param targetEntityId ID of the entity being damaged
+ * @param damageTypeId   0 if there is no damage type, otherwise the damage type ID + 1
+ * @param sourceEntityId 0 if there is no source entity, otherwise it is entityId + 1
+ * @param sourceDirectId 0 if there is no direct source. For direct attacks (e.g. melee), this is the same as sourceEntityId. For indirect attacks (e.g. projectiles), this is the projectile entity id + 1
+ * @param sourcePos      null if there is no source position, otherwise the position of the source
+ */
+public record DamageEventPacket(int targetEntityId, int damageTypeId, int sourceEntityId, int sourceDirectId, @Nullable Point sourcePos) implements ServerPacket {
```
tahmid-23, line 22: shouldn't you send the packet somewhere?
Ex: LivingEntity's use of EntityAnimationPacket

tahmid-23 (commented):
- DeclareRecipesPacket: show notification boolean
- Make ChunkBiomeDataPacket
- Make BundlePacket
- PlayerPositionAndLookPacket and EntityTeleportPacket might be mutually exclusive now? not sure
- RespawnPacket: copyMeta is now a byte with flags "keep attributes" and "keep meta"
- ServerDataPacket: MOTD is no longer optional (default is "")
- ClientClickWindowPacket: if ClickType == PICKUP_ALL and button != 0 you basically pick up items from right to left instead of left to right
- EntityTeleportPacket is no longer sufficient to teleport players, does this change anything? (according to my testing, no)

src/main/java/net/minestom/server/network/packet/server/play/DamageEventPacket.java
MelonHell: i need help with this

src/main/java/net/minestom/server/entity/metadata/display/TextDisplayMeta.java
```diff
@@ -0,0 +1,118 @@
...
+        super.metadata.setIndex(OFFSET + 2, Metadata.VarInt(value.asARGB()));
+    }
+
+    public int getTextOpacity() {
+        return super.metadata.getIndex(OFFSET + 3, -1);
+    }
+
+    public void setTextOpacity(int value) {
+        super.metadata.setIndex(OFFSET + 3, Metadata.VarByte(value));
+    }
```
tahmid-23, line 53:
```suggestion
    public void setTextOpacity(byte value) {
        super.metadata.setIndex(OFFSET + 3, Metadata.Byte(value));
    }
```

src/main/java/net/minestom/server/entity/metadata/display/AbstractDisplayMeta.java
```diff
@@ -0,0 +1,141 @@
...
+
+    protected AbstractDisplayMeta(@NotNull Entity entity, @NotNull Metadata metadata) {
+        super(entity, metadata);
+    }
+
+    public long getInterpolationStartTicks() {
+        return super.metadata.getIndex(OFFSET, 0);
+    }
+
+    public void setInterpolationStartTicks(long value) {
```
tahmid-23, line 24:
```suggestion
    public void setInterpolationStartTicks(int value) {
```

tahmid-23 (changes requested):
I made some mistakes in the web editor, this should compile

src/main/java/net/minestom/server/registry/Registry.java
tahmid-23: can it cause issues when you convert it to NBT? currently the NBT conversion also uses double when it is actually a float

src/main/java/net/minestom/server/network/packet/server/play/DamageEventPacket.java
tahmid-23: in LivingEntity, when shouldAnimate is checked, you should send the new HitAnimationPacket instead of EntityAnimationPacket because TAKE_DAMAGE in that packet is now obsolete
you have to be able to figure out who the "attacker" of the damage is, either with `instanceof` with `EntityDamage` and `EntityProjectileDamage` or adding a new method like `getAttacker` to `Damage`, in order to figure out the correct yaw difference
there is a little bit of random variation in the yaw difference as well, you can see an example in `net.minestom.demo.PlayerInit`

for DamageEventPacket, you can send that packet right before the HitAnimationPacket
you have access to the entity id and damage type id easily
once again, you can do `instanceof` with `EntityDamage` and `EntityProjectileDamage` or you can add `getAttacker` and `getSource` to Damage which are nullable

I think that's the only place

src/main/java/net/minestom/server/registry/Registry.java
```diff
@@ -529,6 +563,10 @@ static Properties fromMap(Map<String, Object> map) {
 
         double getDouble(String name);
 
+        double getFloat(String name, float defaultValue);
```
KrystilizeNevaDies, line 566: should be returning a `float`

mworzala:
Closing in favor of #1937, which will include 1.20.2 changes, which include some more of the patches from this pr (see https://github.com/hollow-cube/minestom-ce/commit/b316a6dbfb7d0b9e7fb24b32b1e295deb338014c and https://github.com/hollow-cube/minestom-ce/commit/6a74dbbc0b0731d1622df6ac085478be35ccec23)

PR #1884, opened 2023-06-19
Fix for #1883

### LINKED ISSUE #1883: equals/hashCode contract violation between Adventure's KeyImpl and Minestom's NamespaceID
NamespaceID and KeyImpl use a slightly different method to compute a hashcode, however, they are considered equal if their namespace (domain) and value (path) are equal. This violates the general contract of `Object#hashCode()`. 

Unit test that reproduces this issue: 

```java
@Test
public void hashCodeConsistentWithEquals() {
    var namespace = NamespaceID.from("minecraft:any");
    var key = Key.key("minecraft:any");

    assertEquals(namespace, key);
    assertEquals(namespace.hashCode(), key.hashCode());
}
```

Steanky:
Fixes #1883 by changing `NamespaceID#hashCode()` to produce the same output as `KeyImpl#hashCode()` given the same input.


tahmid-23 (approved):
great work stnk

PR #1892, opened 2023-07-04
Fix slow dimension changes

mworzala:
Fixes an issue where dimension changes cause the client to sit on the loading terrain screen for a very long time.

Cherry picked from https://github.com/hollow-cube/minestom-ce/commit/388dbf71b9d21c21e9fbd8d691c08ed1fe889665


iam4722202468 (approved):
LGTM

Issue #1295, opened 2022-07-30
Commands broken again


RinesThaix:
I've just updated to the latest master revision, and one of my commands is broken now.
Steps to reproduce (I suppose):
1. Create /mycommand
2. Add a subcommand with name "subcommand" so that it can be executed via "/mycommand subcommand" (I also have an alias for it if it affects anything). Add a default syntax to it without any arguments.
3. Add a raw syntax to the main command with `ArgumentType.Entity("target").singleEntity(true).onlyPlayers(true)` argument.
After that, you won't be able to invoke your subcommand, because any command execution with an argument will be redirected to the raw syntax you added in the last step.

PR #1893, opened 2023-07-06
Rewrite CommandParserImpl recursively (fix #1327)

### LINKED ISSUE #1327: Command bug in syntax parsing

spannerdev:
Fix for #1327 

Rewrites `CommandParserImpl.java` to work recursively. This allows it to backtrack, whereas previously it could not. (This caused the bug; it found an argument that was successful initially, but looking further ahead fails, and it does not recover)

This change needs way more testing before going in, and probably a serious look at what this changes and how it affects existing implementations. I do not know how performant it is, and it may be better to write this iteratively, but I don't think the impact should be too great (keeping in mind that the previous version was not a full/proper implementation)

TODO:
- [x] Fix test failures
- [x] Check compatibility
- [x] Check performance
- [x] Add test for bugs fixed

### DIFF
--- a/src/main/java/net/minestom/server/command/GraphImpl.java
+++ b/src/main/java/net/minestom/server/command/GraphImpl.java
@@ -132,8 +132,17 @@ record GraphImpl(NodeImpl root) implements Graph {
         }
     }
 
-    private record ConversionNode(Argument<?> argument, ExecutionImpl execution,
-                                  Map<Argument<?>, ConversionNode> nextMap) {
+    private static final class ConversionNode {
+        final Argument<?> argument;
+        ExecutionImpl execution;
+        final Map<Argument<?>, ConversionNode> nextMap;
+
+        public ConversionNode(Argument<?> argument, ExecutionImpl execution, Map<Argument<?>, ConversionNode> nextMap) {
+            this.argument = argument;
+            this.execution = execution;
+            this.nextMap = nextMap;
+        }
+
         ConversionNode(Argument<?> argument, ExecutionImpl execution) {
             this(argument, execution, new LinkedHashMap<>());
         }
@@ -152,10 +161,9 @@ record GraphImpl(NodeImpl root) implements Graph {
                 ConversionNode syntaxNode = root;
                 for (Argument<?> arg : syntax.getArguments()) {
                     boolean last = arg == syntax.getArguments()[syntax.getArguments().length - 1];
-                    syntaxNode = syntaxNode.nextMap.computeIfAbsent(arg, argument -> {
-                        var ex = last ? ExecutionImpl.fromSyntax(syntax) : null;
-                        return new ConversionNode(argument, ex);
-                    });
+                    var ex = last ? ExecutionImpl.fromSyntax(syntax) : null;
+                    syntaxNode = syntaxNode.nextMap.computeIfAbsent(arg, argument -> new ConversionNode(argument, ex));
+                    if (syntaxNode.execution == null) syntaxNode.execution = ex;
                 }
             }
             // Subcommands
@@ -173,6 +181,7 @@ record GraphImpl(NodeImpl root) implements Graph {
             }
             return new ConversionNode(Literal(""), null, next);
         }
+
     }
 
     static Argument<String> commandToArgument(Command command) {
--- a/src/main/java/net/minestom/server/command/GraphImpl.java
+++ b/src/main/java/net/minestom/server/command/GraphImpl.java
@@ -4,6 +4,8 @@ import net.minestom.server.command.builder.Command;
 import net.minestom.server.command.builder.CommandExecutor;
 import net.minestom.server.command.builder.CommandSyntax;
 import net.minestom.server.command.builder.arguments.Argument;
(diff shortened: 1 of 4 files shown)


spannerdev:
Fixed as much as I can. As for the remaining test failures, they all follow along the lines of, for example:
`expected (integer=integer, number=5), got (integer=integer, name=name, number=5)"` for `/name integer 5`
The only one that doesn't is `CommandSuggestionIntegrationTest`. I don't really understand what `assertSingle` does or what exactly the error is.

spannerdev:
I've done some benchmarking using the pre-existing benchmarks for commands. The results can be found
[here](https://spanner.codes/share/minestom_command_perf.ods), but in summary:
- For the already slow benchmarks (all relating to number parsing), it is only slower by around 50% (22%,53%,92%), which is around 5µs more.
- For everything else, it increases by something between 400 and 500 nanoseconds

Granted, these benchmarks aren't completely comprehensive, but:
Overall, these results are expected and for pretty much every use case, **completely negligible**. The command parser is doing more work than it was before hence the difference, but developers are able to do significantly more things with this version than the previous.

spannerdev:
## Tested command syntaxes
If anyone has any more commands (preferably with code) for me to test (or have tested yourself) then please do send them to me. What I've tested so far *should* be enough to test that it all works, but it doesn't do anything particularly complicated. (Actually, of note, I might've found a bug in group arguments to do with the commands declaration packet for both this version and the old version, but I can't be sure yet. Either way, it wasn't introduced by this PR)
### From [Basics](https://github.com/spannerdev/basics)
`/basics`
`/give <item> [quantity] [targets] [transaction type]`
`/gamemode <game mode> [targets]`
`/gmc` (includes all other short gamemode commands)
`/teleport <target> <destination|location>` (**previously did not work**)
`/teleport <destination|location>`
`/summon <entity type> [position] [class]`
### Tested #1295
`/1295 subcommand` (**previously did not work**) (although could have been half-fixed before)
`/1295 @p`

src/main/java/net/minestom/server/command/GraphImpl.java
```diff
@@ -76,6 +84,17 @@ static NodeImpl command(Command command) {
         static NodeImpl rootCommands(Collection<Command> commands) {
             return ConversionNode.rootConv(commands).toNode();
         }
+
+        private static final java.util.Comparator<Node> nodePriority = (node1, node2) -> {
+            int node1Value = argumentValue(node1.argument());
+            int node2Value = argumentValue(node2.argument());
+            return Integer.compare(node1Value, node2Value);
+        };
+        private static int argumentValue(Argument<?> argument) {
```
ZakShearman, line 93: should be a space between these methods :)

PR #1917, opened 2023-08-02
Load heightmaps from anvil files

connorrein:
Previously, heightmaps were hardcoded, which made it rain under surfaces:

With this change, heightmaps are read from anvil files and it no longer rains under surfaces:

In a future commit, the heightmaps will need to be updated on block placement.


reosfire:
This logic also is impelemented here: #2044. If my pr will be merged this pr probably should be closed

reosfire:
@mworzala this pr can be closed

PR #1928, opened 2023-08-27
Entity#getUniqueId

NelminDev:
Hello,
I am an old Spigot Dev and I wondered where Player#getUniqueId went then I saw it's Player#getUuid and I thoughtt I can change it :)

PS: I didn't remove getUuid()

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -754,13 +754,17 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
      * Gets the entity {@link UUID}.
      *
      * @return the entity unique id
+     * @Deprecated in favor of {@link getUniqueId}
      */
+    @Deprecated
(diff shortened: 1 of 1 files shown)


ZakShearman:
Both methods shouldn't exist - a decision should be made on what naming scheme should be used and the old one deprecated.

Tjorven-Liebe (approved):

mworzala:
It seems like an unnecessary breaking change to me.

NelminDev:
Why breaking?

mworzala:
Sorry, I should have clarified. I agree with Zak that there should not be both of them, and removing `getUuid` will be a breaking change.

FlorianMichael:
There is no reason to change the name of the method so randomly, it will only unnecessarily break API that the method was using, such changes should only be done with major updates where API is definitely changed

KrystilizeNevaDies:
I will close this pr for now since the consensus is that we don't want to change the api without a good reason, and being consistent with spigot isn't a good enough reason for us.

PR #1935, opened 2023-09-03
Prioritise errors by chain length, build subcommands first (fix #1934)

### LINKED ISSUE #1934: Command suggestionCallback regression
suggestionCallback has been broken in some situations by #1893. This has caused some of our commands not to function.

The case demonstrated here is if you have a command `/badge admin add <1> <2>` and also `/badge set <3>` this causes the suggestion of `1` to never be called upon (however the mapInput of 1 is still called). The suggestion of `2` will be called correctly and overall, the command still executes.

This is not an issue with argument 1 as if you swap 1 and 2, then 2 will be broken. It's purely because of the position of the argument and presence of `/badge set <3>`  

Reproducer code:
```java
public class VanillaTest {

    public static void main(String... args) {
        minecraftServer.start("0.0.0.0", 25565);

        Instance instance = MinecraftServer.getInstanceManager().createInstanceContainer();
        instance.setGenerator(unit -> unit.modifier().fillHeight(-20, 0, Block.GRASS_BLOCK));

        MinecraftServer.getGlobalEventHandler().addListener(PlayerLoginEvent.class, event -> {
            event.getPlayer().setRespawnPoint(new Pos(0, 0, 0));
        }).addListener(PlayerSpawnEvent.class, event -> event.getPlayer().setFlying(true));

        Command command = new Command("badge");
        Command addSubCommand = new Command("admin");

        ArgumentLiteral addArg = new ArgumentLiteral("add");

        Argument<String> playerArg = new ArgumentWord("player").map(s -> {
            Audiences.all().sendMessage(Component.text("A - player mapInput called"));
            return s;
        }).setSuggestionCallback((sender, context, suggestion) -> {
            Audiences.all().sendMessage(Component.text("B - player suggestion called"));
            suggestion.addEntry(new SuggestionEntry("testOne"));
        });

        Argument<String> badgeArg = new ArgumentWord("badge")
                .setSuggestionCallback((sender, context, suggestion) -> {
                    Audiences.all().sendMessage(Component.text("C - badge suggestion called"));
                    suggestion.addEntry(new SuggestionEntry("testTwo"));
                });

        addSubCommand.addSyntax((sender, context) -> sender.sendMessage("admin add command!"), addArg, playerArg, badgeArg);

        command.addSubcommand(addSubCommand);

        command.addSyntax(((sender, context) -> {
            sender.sendMessage(Component.text("Hello World!"));
        }), new ArgumentLiteral("set"), badgeArg);

        MinecraftServer.getCommandManager().register(command);
    }
}
```

spannerdev:
More discussion of fixes found at #1934.

Despite #1893 adding priority to nodes, the graph converter converts command arguments (as subcommands) to literals before they reach the node. This is fixed by converting the subcommands before the syntaxes.
On top of that, a variation of the issue can be created by having two literals instead of a literal and subcommand. The error returned will depend on the order of addSyntax. This is wrong, and is fixed by returning the error with the longest chain length.

I would like to add a test to check that commands are always processed before literals, but I don't know how to do this given the current test utils. It doesn't fit in `CommandSyntaxMultiTest`, perhaps `CommandParseTest`? If someone who has better experience with the testing could give some feedback that would be great.

If anyone has code that relies on the errors reported, please test this - it works fine internally but I can't predict how it will affect stuff like that.

ps: ignore the first commit. I think the squash merge did some funky stuff with git and now it's still there.

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandParserImpl.java
+++ b/src/main/java/net/minestom/server/command/CommandParserImpl.java
@@ -74,6 +74,10 @@ final class CommandParserImpl implements CommandParser {
             return nodeResults.stream().map(x -> x.node.argument()).collect(Collectors.toList());
         }
 
+        int size() {
+            return nodeResults.size();
+        }
+
         Chain() {}
 
         Chain(CommandExecutor defaultExecutor,
@@ -137,7 +141,7 @@ final class CommandParserImpl implements CommandParser {
             NodeResult nodeResult = new NodeResult(node, chain, (ArgumentResult<Object>) result, suggestionCallback);
             chain.append(nodeResult);
             if (suggestionCallback != null) chain.suggestionCallback = suggestionCallback;
-            if (chain.nodeResults.size() == 1) { // If this is the root node (usually "Literal<>")
+            if (chain.size() == 1) { // If this is the root node (usually "Literal<>")
                 reader.cursor(start);
             } else {
                 if (!(result instanceof ArgumentResult.Success<?>)) {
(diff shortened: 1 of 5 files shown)


mworzala:
You can always add a new test file if it doesn't fit somewhere already

spannerdev:
:o me? a new file? I could never...

ok anyway this is ready to be merged as far as I can tell. @ZakShearman does this fix your issue #1934 ?

PR #1952, opened 2023-10-07
Remove default damage behaviour

GoldenStack:
Removes void and fire damage handling that Minestom handles by default.
It's just not something that should be default behaviour.


GoldenStack:
Moving to CE

PR #1986, opened 2024-02-09
Add aliases to a literal argument

Codestech1:
Adds a way to add aliases to a literal argument


mworzala:
This should be done either by adding another Literal with your alias, or add a Word with the possible values.

Codestech1:
bruh

PR #1994, opened 2024-02-11
Fix Maven Central Link

AndusDEV:
Just a simple change from https://central.sonatype.com/artifact/net.minestom/minestom to https://central.sonatype.com/artifact/net.minestom/minestom-snapshots


MelonHell (approved):

mworzala:
oops. thanks for the fix

PR #1996, opened 2024-02-15
Fix: FakePlayer is not registered in connection player map

Fauxle:
FakePlayers, when initialized, have no mapping in `connectionPlayerMap` in `ConnectionManager` because it looks like `ConnectionManager#createPlayer` is not being called. This commit adds a fallback to connection manager where if a player is being transitioned from login to config and no connection <-> player map exists, set it as a fallback. The map needs to be set because if the FakePlayer is disconnected, it causes inaccurate player online counts and issues for duplicate FakePlayers with the same username.

To see the issue, comment out the fallback fix in ConnectionManager. Then run the commands:
- `/fakeplayer Test`
- `/kick Test`
- `/fakeplayer Test`

The 2nd FakePlayer remains online and is not disconnected. Also, after the first player is kicked, the online player count becomes inaccurate.

I also added a test for checking when a FakePlayer is initialized and disconnected.

### DIFF
--- a/demo/src/main/java/net/minestom/demo/Main.java
+++ b/demo/src/main/java/net/minestom/demo/Main.java
@@ -75,6 +75,8 @@ public class Main {
         commandManager.register(new SetEntityType());
         commandManager.register(new RelightCommand());
         commandManager.register(new KillCommand());
+        commandManager.register(new KickCommand());
+        commandManager.register(new FakePlayerCommand());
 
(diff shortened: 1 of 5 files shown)


MelonHell:
@mworzala planned to remove FakePlayer https://github.com/hollow-cube/minestom-ce/pull/135

Fauxle:
I see thanks, didn't know
The pr can be closed then if fake players are removed but I'll leave it open if plans change. FakePlayer is very similar to how the citizens plugin does it with their npcs so that's probably why it's used that way

mworzala:
See #2006

PR #1997, opened 2024-02-17
General improvement for code readability

TheMeinerLP:
This PR improves the code structure and readability for the recipe manager. The packages were extracted into an interface with a static method and only released on package. A To-Do has been removed and the switch statement has been tidied up.

### DIFF
--- a/src/main/java/net/minestom/server/recipe/PacketDeclaration.java
+++ b/src/main/java/net/minestom/server/recipe/PacketDeclaration.java
@@ -1,0 +1,104 @@
+package net.minestom.server.recipe;
+
+import net.minestom.server.network.packet.server.play.DeclareRecipesPacket;
+import org.jetbrains.annotations.NotNull;
+
+interface PacketDeclaration {
+
+    @NotNull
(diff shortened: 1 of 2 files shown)


mworzala:
Sorry I haven't responded to this I have been thinking. I don't love the weird random interface with a bunch of impl in it. That feels very messy to me. I believe it would be be better at least as a final class clearly intended as a utility, or perhaps move the conversion to packet into a function in the Recipe which is overridden by each impl (to return a declared recipe packet object).

TheMeinerLP:
The goal of the PR was to offer readability, maintainability, backward compatibility and hardening of the API. Compared to before, this is more readable and less breakable than building a new structure that breaks APIs or software interfaces in general. 

In that sense, we have differentiated views on software architecture and software design. 

What would be the counter-proposal and what would your ideal vision look like?

iam4722202468:
Why would you use an interface when it's never implemented? A utility class makes a lot more sense, and fits with the current codebase.

How are you hardening the API when this code is not accessible from the API? This PR makes no API changes, all methods are package private.

mworzala:
Closing in favor of #2010. I tried to commit to your branch but apparently permission has been denied.

PR #1998, opened 2024-02-17
Improved socket handling for the server

TheMeinerLP:
The PR now closes the socket correctly as soon as the Stop method is called by the server class. In addition, thrown errors are handled in the sense that the program can still function stably afterwards. And the errors are passed to a logger.

The changes are due to the fact that on Unix systems with a low ulimit a `java.io.FileNotFoundException: XXX (Too many open files)` occurred. This happened especially with unit tests.

### DIFF
--- a/src/main/java/net/minestom/server/network/socket/Server.java
+++ b/src/main/java/net/minestom/server/network/socket/Server.java
@@ -15,8 +15,12 @@ import java.nio.channels.SocketChannel;
 import java.nio.file.Files;
 import java.util.Arrays;
 import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class Server {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Server.class);
 
     public static final boolean NO_DELAY = true;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/network/socket/Server.java
```diff
@@ -112,8 +116,21 @@ public void stop() {
...
-        this.selector.wakeup();
-        this.workers.forEach(worker -> worker.selector.wakeup());
+        try {
+            this.selector.close();
+        } catch (IOException e) {
+            LOGGER.error("Server socket sector could not be closed", e);
+            System.exit(-1);
+        }
+        this.workers.forEach(worker -> {
+            try {
```
mworzala, line 126: should just add a close function on Worker

src/main/java/net/minestom/server/network/socket/Worker.java
```diff
@@ -19,11 +19,15 @@
...
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @ApiStatus.Internal
 public final class Worker extends MinestomThread {
     private static final AtomicInteger COUNTER = new AtomicInteger();
 
+    static final Logger LOGGER = LoggerFactory.getLogger(Server.class);
```
mworzala, line 29: Referencing another classes logger seems very strange, but will be fixed by adding a close method here

src/main/java/net/minestom/server/network/socket/Server.java
```diff
@@ -112,8 +116,21 @@ public void stop() {
         } catch (IOException e) {
             MinecraftServer.getExceptionManager().handleException(e);
         }
-        this.selector.wakeup();
```
mworzala, line 115: I'm unconvinced about removing these, it seems valid to process anything pending before shutting down

TheMeinerLP: Close calls this internal also

src/main/java/net/minestom/server/network/socket/Worker.java
TheMeinerLP: As this had already been done before in other forms, I thought it wouldn't be a problem. A little hint, there are classes that continue to do this

mworzala: No need to add another case :)

mworzala (approved):

PR #2002, opened 2024-02-18
RelativeBlockBatch update

iam4722202468:


mworzala (changes requested):
getBlock should not exist, it exposes internal details and makes it harder to change in the future.

The diff and join methods should be on `Batch`

PR #2004, opened 2024-02-23
Add particle data types

iam4722202468:


src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
```diff
@@ -37,12 +38,12 @@ public void setSinglePoint(boolean value) {
...
     }
 
-//    public ParticleWrapper getParticle() {
-//        return super.metadata.getIndex((byte) 10, new ParticleWrapper(Particle.EFFECT, null));
-//    }
-//
-//    public void setParticle(ParticleWrapper value) {
-//        super.metadata.setIndex((byte) 11, Metadata.Particle(value));
-//    }
+    public Particle getParticle() {
```
mworzala, line 41: annotations

src/main/java/net/minestom/server/network/NetworkBuffer.java
```diff
@@ -75,6 +76,7 @@ public final class NetworkBuffer {
     public static final Type<Point> VECTOR3 = NetworkBufferTypes.VECTOR3;
     public static final Type<Point> VECTOR3D = NetworkBufferTypes.VECTOR3D;
     public static final Type<float[]> QUATERNION = NetworkBufferTypes.QUATERNION;
+    public static final Type<Particle> PARTICLE = NetworkBufferTypes.PARTICLE_DATA;
```
mworzala, line 79: these should be the same name

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
```diff
@@ -4,20 +4,56 @@
...
 
+import java.util.Objects;
+
 import static net.minestom.server.network.NetworkBuffer.*;
 
-public record ParticlePacket(int particleId, boolean longDistance,
-                             double x, double y, double z,
-                             float offsetX, float offsetY, float offsetZ,
-                             float particleData, int particleCount, byte[] data) implements ServerPacket {
+public final class ParticlePacket implements ServerPacket {
```
mworzala, line 15: Why is this converted away from a record? it should go back

src/main/java/net/minestom/server/particle/Particle.java
```diff
@@ -24,4 +25,7 @@ public sealed interface Particle extends StaticProtocolObject, Particles permits
     static @Nullable Particle fromId(int id) {
         return ParticleImpl.getId(id);
     }
+
+    Particle withData(ParticleData data);
```
mworzala, line 29: annotations

src/main/java/net/minestom/server/particle/ParticleCreator.java
```diff
@@ -14,23 +10,15 @@ public class ParticleCreator {
     public static ParticlePacket createParticlePacket(Particle particleType, boolean distance,
```
mworzala, line 10: This is a bit worthless now, no?

src/main/java/net/minestom/server/particle/data/BlockMarkerParticleData.java
```diff
@@ -0,0 +1,16 @@
+package net.minestom.server.particle.data;
+
+import net.minestom.server.instance.block.Block;
+import net.minestom.server.network.NetworkBuffer;
+import org.jetbrains.annotations.NotNull;
+
+public record BlockMarkerParticleData(Block block) implements ParticleData {
```
mworzala, line 7: annotations on all of these

src/main/java/net/minestom/server/particle/data/DustColorTransitionParticleData.java
```diff
@@ -0,0 +1,52 @@
...
+        if (fromBlue < 0 || fromBlue > 1) {
+            throw new IllegalArgumentException("fromBlue must be between 0 and 1");
+        }
+        if (toRed < 0 || toRed > 1) {
+            throw new IllegalArgumentException("toRed must be between 0 and 1");
+        }
+        if (toGreen < 0 || toGreen > 1) {
+            throw new IllegalArgumentException("toGreen must be between 0 and 1");
+        }
+        if (toBlue < 0 || toBlue > 1) {
```
mworzala, line 23: use Check.argCondition for these (same for other classes)

src/main/java/net/minestom/server/particle/data/DustColorTransitionParticleData.java
```diff
@@ -0,0 +1,52 @@
+package net.minestom.server.particle.data;
+
+import net.minestom.server.network.NetworkBuffer;
+import org.jetbrains.annotations.NotNull;
+
+public record DustColorTransitionParticleData(float fromRed, float fromGreen, float fromBlue, float scale, float toRed, float toGreen, float toBlue) implements ParticleData {
```
mworzala, line 6: What about instead making these Adventure RGBLike or whatever the type is? Or maybe an overload?

src/main/java/net/minestom/server/particle/data/ParticleData.java
```diff
@@ -0,0 +1,26 @@
...
+
+import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.particle.Particle;
+import org.jetbrains.annotations.NotNull;
+
+public interface ParticleData {
+    void write(@NotNull NetworkBuffer writer);
+
+    static ParticleData read(int particleId, NetworkBuffer reader) {
+        if (particleId == Particle.BLOCK.id()) return new BlockParticleData(reader);
```
mworzala, line 11: Area effect cloud?

src/main/java/net/minestom/server/particle/data/VibrationBlockParticleData.java
```diff
@@ -0,0 +1,18 @@
+package net.minestom.server.particle.data;
+
+import net.minestom.server.coordinate.Point;
+import net.minestom.server.network.NetworkBuffer;
+import org.jetbrains.annotations.NotNull;
+
+public record VibrationBlockParticleData(Point source, int ticks) implements ParticleData {
```
mworzala, line 7: I'm not super convinced on these being two different records, they do correspond to the same packet. For other packets with unions like this they are generally in a single record

src/test/java/net/minestom/server/entity/AreaEffectCloudTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import net.minestom.server.entity.metadata.other.AreaEffectCloudMeta;
+import net.minestom.server.instance.block.Block;
+import net.minestom.server.particle.Particle;
+import net.minestom.server.particle.data.BlockParticleData;
+import net.minestom.server.particle.data.DustColorTransitionParticleData;
+import net.minestom.server.particle.data.DustParticleData;
+import org.junit.jupiter.api.Test;
+
+public class AreaEffectCloudTest {
```
mworzala, line 11: More tests needed for reading and writing other particles probably

mworzala (commented):
I left some comments for pretty minor things but I think there is one bigger issue I have: Its still not very safe to create particles.
1. If you simply send a `Particle.DUST` it will cause an issue because there is no data by default. It would be nice if the particle constants had a default data associated to them, although that doesn't make much sense for things like skulk. Alternatively, there could be an error when trying to write a particle which doesn't have the correct data for its type.
2. You can set any data on any particle. If you try to set `DustParticleData` on a marker particle it should be an error.
3. Kinda mentioned in #1, but when serializing a particle it should be an error if the data is not the correct type (including checking if missing for a particle it should be present).

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
iam4722202468: we need the particleId on line 46 and 56, i'm not sure using a record is possible here

mworzala: Use the pattern of having a copy constructor and a private static read function, see UnlockRecipesPacket for a similar example

src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
```diff
@@ -37,12 +38,12 @@ public void setSinglePoint(boolean value) {
...
     }
 
-//    public ParticleWrapper getParticle() {
-//        return super.metadata.getIndex((byte) 10, new ParticleWrapper(Particle.EFFECT, null));
-//    }
-//
-//    public void setParticle(ParticleWrapper value) {
-//        super.metadata.setIndex((byte) 11, Metadata.Particle(value));
-//    }
+    public @NotNull Particle getParticle() {
```
mworzala, line 41: should just be nullable, no?

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
```diff
@@ -4,24 +4,70 @@
...
+        Float maxSpeed = reader.read(FLOAT);
+        Integer particleCount = reader.read(INT);
+        ParticleData data = ParticleData.read(particleId, reader);
+
+        return new ParticlePacket(particleId, longDistance, x, y, z, offsetX, offsetY, offsetZ, maxSpeed, particleCount, data);
     }
 
     @Override
     public void write(@NotNull NetworkBuffer writer) {
+        if (data != null && !data.validate(particleId)) throw new IllegalStateException("Particle data is not valid for this particle type");
```
mworzala, line 68: use `Check.stateCondition` (or maybe arg its more or less an argument), and provide the data type and particle type in the error message

src/main/java/net/minestom/server/particle/data/BlockMarkerParticleData.java
```diff
@@ -0,0 +1,32 @@
...
+        this(read(reader));
+    }
+
+    BlockMarkerParticleData() {
+        this(Block.STONE);
+    }
+
+    private static Block read(NetworkBuffer reader) {
+        Block block = Block.fromStateId(reader.read(NetworkBuffer.VAR_INT).shortValue());
+        if (block == null) return Block.STONE;
```
mworzala, line 19: Not so sure about this, seems like it should be air or an error

src/main/java/net/minestom/server/particle/data/BlockParticleData.java
```diff
@@ -0,0 +1,26 @@
+package net.minestom.server.particle.data;
+
+import net.minestom.server.instance.block.Block;
+import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.particle.Particle;
+import org.jetbrains.annotations.NotNull;
+
+public record BlockParticleData(Block block) implements ParticleData {
+    BlockParticleData(NetworkBuffer reader) {
+        this(Block.fromStateId(reader.read(NetworkBuffer.VAR_INT).shortValue()));
```
mworzala, line 10: Inconsistent with block marker which checks if the block is valid

src/main/java/net/minestom/server/particle/data/DustParticleData.java
```diff
@@ -0,0 +1,49 @@
...
+    DustParticleData(NetworkBuffer buffer) {
+        this(read(buffer));
+    }
+
+    private DustParticleData(DustParticleData copy) {
+        this(copy.color, copy.scale);
+    }
+
+    private static DustParticleData read(NetworkBuffer buffer) {
+        RGBLike color = new Color(
```
mworzala, line 24: this doesn't need a separate read method it can be done in the super call

src/main/java/net/minestom/server/particle/data/DustColorTransitionParticleData.java
```diff
@@ -0,0 +1,57 @@
...
+    DustColorTransitionParticleData(NetworkBuffer buffer) {
+        this(read(buffer));
+    }
+
+    private DustColorTransitionParticleData(DustColorTransitionParticleData copy) {
+        this(copy.from, copy.scale, copy.to);
+    }
+
+    private static DustColorTransitionParticleData read(NetworkBuffer buffer) {
+        RGBLike from = new Color(
```
mworzala, line 28: unnecessary separate read function

src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
iam4722202468: we're passing default dust, in `getIndex` if the entry is null it will return the default.

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
iam4722202468: Fixed. Going to to use stateCondition, since argument feels like something that should be thrown when creating something

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
```diff
@@ -4,24 +4,71 @@
...
 
+import java.util.Objects;
+
 import static net.minestom.server.network.NetworkBuffer.*;
 
-public record ParticlePacket(int particleId, boolean longDistance,
-                             double x, double y, double z,
-                             float offsetX, float offsetY, float offsetZ,
-                             float particleData, int particleCount, byte[] data) implements ServerPacket {
+public record ParticlePacket(int particleId, boolean longDistance, double x, double y, double z, float offsetX, float offsetY, float offsetZ, float maxSpeed, int particleCount, ParticleData data) implements ServerPacket {
```
mworzala, line 17: @Nullable on data

src/main/java/net/minestom/server/network/packet/server/play/ParticlePacket.java
```diff
@@ -4,24 +4,71 @@
...
+
     public ParticlePacket(@NotNull NetworkBuffer reader) {
-        this(reader.read(VAR_INT), reader.read(BOOLEAN),
-                reader.read(DOUBLE), reader.read(DOUBLE), reader.read(DOUBLE),
-                reader.read(FLOAT), reader.read(FLOAT), reader.read(FLOAT),
-                reader.read(FLOAT), reader.read(INT), reader.read(RAW_BYTES));
+        this(readPacket(reader));
+    }
+
+    public ParticlePacket(Particle particle, boolean longDistance, double x, double y, double z, int offsetX, int offsetY, int offsetZ, int maxSpeed, int particleCount) {
```
mworzala, line 43: annotations here and below

src/main/java/net/minestom/server/particle/data/DustColorTransitionParticleData.java
```diff
@@ -0,0 +1,46 @@
...
+import net.kyori.adventure.util.RGBLike;
+import net.minestom.server.color.Color;
+import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.particle.Particle;
+import net.minestom.server.utils.validate.Check;
+import org.jetbrains.annotations.NotNull;
+
+public record DustColorTransitionParticleData(@NotNull RGBLike from, float scale, @NotNull RGBLike to) implements ParticleData {
+    public DustColorTransitionParticleData {
+        Check.argCondition(scale < 0.01 || scale > 4, "scale must be between 0.01 and 4");
```
mworzala, line 12: ": was 5555555"

src/main/java/net/minestom/server/particle/data/FallingDustParticleData.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    FallingDustParticleData() {
+        this(Block.STONE);
+    }
+
+    private static Block read(NetworkBuffer reader) {
+        short blockState = reader.read(NetworkBuffer.VAR_INT).shortValue();
+        Block block = Block.fromStateId(blockState);
+        Check.stateCondition(block == null, "Block state " + blockState + " is invalid");
```
mworzala, line 21: again use format arg on stateCondition

PR #2005, opened 2024-02-24
Allow offsetting the BoundingBox

cosrnic:
Makes the offset constructor public and adds a method to add one to an existing bounding box


iam4722202468:
LGTM

PR #2023, opened 2024-03-14
Permissions rework 

### LINKED ISSUE #2021: Rework the permission system
Problems:
- There is no way to replace a player's PermissionHandler without extending Player class
- The net.minestom.server.permission.Permission class is useless
- No tristate system (True/False/Undefined)

MelonHell:
resolves #2021


LooFifteen (approved):
lgtm

MelonHell:
Why do entities need permissions? It seems to me that they are only needed by CommandSender

MelonHell:
_TODO rename Permissions back to PermissionHandler and add a setter_

MelonHell:
Permissions removed in https://github.com/Minestom/Minestom/pull/2445

PR #2025, opened 2024-03-15
Fix: disable time updates when not needed

reosfire:
When time rate is zero we do not need to send TimeUpdatePackets to users.

### DIFF
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -470,6 +470,7 @@ public abstract class Instance implements Block.Getter, Block.Setter,
      */
     public void setTimeRate(int timeRate) {
         Check.stateCondition(timeRate < 0, "The time rate cannot be lower than 0");
+        if (timeRate == 0) setTimeUpdate(null);
         this.timeRate = timeRate;
     }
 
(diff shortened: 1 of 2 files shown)


emortaldev:
What about when you do `setTimeRate(0)` and `setTimeRate(1)`? TimeUpdate will stay as null

mworzala (changes requested):
This should not alter the time update. Avoiding sending the time packet should be handled when sending it during tick, here: https://github.com/Minestom/Minestom/blob/4fea03fe3296b1531199f86b67f2dd1fbb6b78cd/src/main/java/net/minestom/server/instance/Instance.java#L666

I was under the impression the client would predict the time rate so you had to continuously send the time packet, but apparently that is not the case (anymore?).

reosfire:
> This should not alter the time update. Avoiding sending the time packet should be handled when sending it during tick, here:
> 
> https://github.com/Minestom/Minestom/blob/4fea03fe3296b1531199f86b67f2dd1fbb6b78cd/src/main/java/net/minestom/server/instance/Instance.java#L666
> 
> I was under the impression the client would predict the time rate so you had to continuously send the time packet, but apparently that is not the case (anymore?).

You right client some how uses this information. But it is not clear how exactly. If continuous packets sending is disabled, everything what I can imagine works properly. Probably this changes must wait for clarification of how client is affected by them

DeidaraMC:
> You right client some how uses this information. But it is not clear how exactly. If continuous packets sending is disabled, everything what I can imagine works properly. Probably this changes must wait for clarification of how client is affected by them

The client uses the world age for shaders

kermandev:
No longer applicable in 26.1

PR #2026, opened 2024-03-17
feat: add leash holding support to the entity api

DeidaraMC:


mworzala:
Is there any reason to limit an entity to only being leashed by a single entity? In vanilla you cannot leash an entity already held by someone else, but I assume this is just a server check?

DeidaraMC:
> Is there any reason to limit an entity to only being leashed by a single entity? In vanilla you cannot leash an entity already held by someone else, but I assume this is just a server check?

The client only supports 1 leash holder, the most recent packet sent overrides who is holding the leash on the client

mworzala (approved):

PR #2028, opened 2024-03-17
chore: move packet flush to after ServerTickMonitorEvent

DeidaraMC:
ServerTickMonitorEvent is a very useful event for running logic on final state the server is in on a given tick
packets should flushed after this event to reflect any logic here immediately


mworzala:
To me this feels like a weird abuse of the tick monitor event (to run logic at end of tick), I would prefer some dedicated way to run logic at the end of a tick.

PR #2029, opened 2024-03-17
feat: added player#teleportWithFlags for relative teleportation

DeidaraMC:


mworzala:
I have no idea the use case for this, more info required.

DeidaraMC:
> I have no idea the use case for this, more info required.

The flags let you teleport the client with relative values rather than absolute values. 
They also let you preserve the clients' velocity/look angles which let you smoothly teleport a player without forcing their yaw/pitch to be a specific way or stopping their momentum.
An excellent use case for this would be an endless staircase which teleports you back to the bottom without being noticeable because your yaw/pitch was never forcefully set nor momentum lost.

mworzala (commented):
Thanks for the info. Some comments:

- I think this API should just be an overload to the normal teleport function, not its own function (just a rename, but it would be better if there was not duplication, ie teleport can just call relative teleport with no flags set)
- I would prefer if the flags were just int constants OR-ed together instead of enum varargs (with some extras for X|Y|Z and YAW|PITCH perhaps)

DeidaraMC:
> Thanks for the info. Some comments:
> 
> * I think this API should just be an overload to the normal teleport function, not its own function (just a rename, but it would be better if there was not duplication, ie teleport can just call relative teleport with no flags set)
> * I would prefer if the flags were just int constants OR-ed together instead of enum varargs (with some extras for X|Y|Z and YAW|PITCH perhaps)

> We can modify this code to simply work with the teleport function instead
> Original I did have them as int constants, however from an API standpoint I couldn't figure out an intuitive place to put the constants for people to find (any ideas?). Other places use Enum flags like item metadata as well

DasBabyPixel:
> Thanks for the info. Some comments:
> 
> - I think this API should just be an overload to the normal teleport function, not its own function (just a rename, but it would be better if there was not duplication, ie teleport can just call relative teleport with no flags set)
> - I would prefer if the flags were just int constants OR-ed together instead of enum varargs (with some extras for X|Y|Z and YAW|PITCH perhaps)

Not a fan of int flags. Neither future-proof nor intuitive. varargs are better suited for an api imo.

mworzala:
Alright im convinced. It should not be declared halfway though the Player class however. Should be on its own preferably or at the top maybe.

DeidaraMC:
@mworzala 
Big changes here
- Refactored to simply be a part of the entity's teleport method
- Removed the includeSelf argument from synchronizePosition (we no longer need it) and replaced with Player#synchronizePositionAfterTeleport
- Moved the enum to its own file
- Added a way to get common flag cases in the RelativeFlag enum to reduce code bloat when including a bunch of flags
- Tidied up the teleport function (also no longer need to explicitly send the head packet because synchronize does that now)

src/main/java/net/minestom/server/entity/RelativeFlag.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static RelativeFlag @NotNull [] allFlags() {
+        return new RelativeFlag[] { X, Y, Z, YAW, PITCH };
+    }
+
+    public static RelativeFlag @NotNull [] axisFlags() {
+        return new RelativeFlag[] { X, Y, Z };
+    }
+
+    public static RelativeFlag @NotNull [] lookFlags() {
```
mworzala, line 39: should probably be called `viewFlags` since we refer to yaw/pitch as view elsewhere (eg `Pos#withView`)

src/main/java/net/minestom/server/entity/RelativeFlag.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return bits;
+    }
+
+    public static RelativeFlag @NotNull [] allFlags() {
+        return new RelativeFlag[] { X, Y, Z, YAW, PITCH };
+    }
+
+    public static RelativeFlag @NotNull [] axisFlags() {
```
mworzala, line 35: should probably be called coordFlags since we refer to x/y/z as coords elsewhere (eg Pos#withCoords)

src/test/java/net/minestom/server/entity/player/PlayerIntegrationTest.java
```diff
@@ -4,6 +4,7 @@
 import net.minestom.server.coordinate.Pos;
 import net.minestom.server.entity.GameMode;
 import net.minestom.server.entity.Player;
+import net.minestom.server.entity.RelativeFlag;
```
mworzala, line 7: Weird added imports

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2370,10 +2372,18 @@ protected void sendChunkUpdates(Chunk newChunk) {
...
+     * @see #teleport(Pos, long[], RelativeFlag...) 
+     */
+    @Override
+    public @NotNull CompletableFuture<Void> teleport(@NotNull Pos position, long @Nullable [] chunks, RelativeFlag... flags) {
+        chunkUpdateLimitChecker.clearHistory();
+        return super.teleport(position, chunks, flags);
+    }
+
     @Override
     public @NotNull CompletableFuture<Void> teleport(@NotNull Pos position, long @Nullable [] chunks) {
```
mworzala, line 2385: Unnecessary override

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -311,33 +312,40 @@ public <TMeta extends EntityMeta> void editEntityMeta(Class<TMeta> metaClass, Co
...
-            return instance.loadOptionalChunk(position).thenRun(endCallback);
+            return instance.loadOptionalChunk(globalPosition).thenRun(endCallback);
         } else {
             // Position is in the same chunk, keep it sync
             endCallback.run();
             return AsyncUtils.empty();
         }
     }
 
+    public @NotNull CompletableFuture<Void> teleport(@NotNull Pos position, long @Nullable [] chunks) {
```
mworzala, line 345: This is an unnecessary overload I think? No varargs should call the same

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -311,33 +312,40 @@ public <TMeta extends EntityMeta> void editEntityMeta(Class<TMeta> metaClass, Co
...
+        final Pos globalPosition = PositionUtils.getPositionWithRelativeFlags(this.position, position, relativeBits);
         final Runnable endCallback = () -> {
             this.previousPosition = this.position;
-            this.position = position;
-            refreshCoordinate(position);
-            synchronizePosition(true);
-            sendPacketToViewers(new EntityHeadLookPacket(getEntityId(), position.yaw()));
+            this.position = globalPosition;
+            refreshCoordinate(globalPosition);
+            if (this instanceof Player player) player.synchronizePositionAfterTeleport(position, relativeBits);
```
mworzala, line 326: This feels worse than how it was before. There shouldn't be player specific logic in entity if it can be handled by an override in Player (which keeps it self contained there)

DeidaraMC: > This feels worse than how it was before. There shouldn't be player specific logic in entity if it can be handled by an override in Player (which keeps it self contained there)

I agree that instanceof is a codesmell but can we also agree that needing to pass in a position & relative bits for a method only used by player twice and other/custom entities way more is smelly as well?

instanceof is used in multiple places throughout the entity class and I think this is the lesser of two evils when considering other developers who will be extending entity in the future

mworzala (approved):

PR #2030, opened 2024-03-17
Adventure NBT to replace Hephaistos

mworzala:
**This is not final, nor is it definitely getting merged**

For now this is just an experiment. I have not ported an Anvil loader, and none of the tests pass. Adventure NBT has some not-niceties which I will try to discuss with them.

1. No way to get an `BinaryTagType` from an int ID (its private)
2. The reader/writer is inflexible (it's not being used in this PR). It also only supports reading and writing compounds, which is not protocol-accurate.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -70,10 +70,7 @@ dependencies {
     api(libs.jetbrainsAnnotations)
     api(libs.bundles.adventure)
     api(libs.hydrazine)
-    api(libs.bundles.kotlin)
-    api(libs.bundles.hephaistos)
     implementation(libs.minestomData)
-    implementation(libs.dependencyGetter)
 
     // Performance/data structures
     implementation(libs.caffeine)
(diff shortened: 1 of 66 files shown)


TheMode (approved):

UserUNP:
What's the difference between the adventure nbt API and hephaistos?

mworzala:
This is blocked by the improvements being made to the NBT Reader/Writer in Adventure 4.17.0
https://github.com/KyoriPowered/adventure/milestone/27

mworzala:
> What's the difference between the adventure nbt API and hephaistos?

Well they are both NBT at the end of the day. Adventure doesn't need to be maintained by us, is more standard theoretically (if people have experience with adventure otherwise), and is in Java not Kotlin (Hephaistos required bringing all of Kotlin along with Minestom also).

mworzala:
This branch is being included in the 1.20.5 release (#2089) with our own shims for reading and writing (properly) until adventure-nbt gets this functionality (hopefully with 4.17.0 / their 1.20.5 release)

Issue #1881, opened 2023-06-15
Entity head rotation doesn't update on teleport


ZakShearman:
The entity head rotation packet (EntityHeadLookPacket) on teleport is only sent by the setView method. However, this never gets sent because the position has already been updated so it fast exits.

Don't have time to do this right now so if anyone wants to look at it :)

TheMeinerLP:
Have you code to repdroduce it ?

GreatWyrm:
Could you try to reproduce it with the latest commit? It may be fixed now

iam4722202468:
This is fixed

PR #2031, opened 2024-03-17
fix: send EntityHeadLookPacket on teleport

DeidaraMC:
setView will always return immediately before sending the EntityHeadLookPacket
there's no reason to call it, rather, we should send the necessary packet explicitly


ZakShearman:
I believe this should resolve #1881?

GreatWyrm:
@ZakShearman It could resolve that issue, would you mind retesting with this branch?

@DeidaraMC I'd recommend you fix the test as well that broke with the change

mworzala (approved):
Nice catch, thanks

PR #2032, opened 2024-03-18
feat: add weather to instances

DeidaraMC:


mworzala:
Weather changes need to be interpolated over rainLevel to players already in the instance at time of change. I understand this can be used for static weather effects, but it feels like a very incomplete api.

DeidaraMC:
Weather now does a linear interpolation from its current state to its target state passed in Instance#setWeather over the course of the specified transition ticks.
Setting transition ticks to 1 will prompt an immediate transition on the next instance tick.
Passing in a state like (false, 1, 1) will set the instance's rain and thunder levels to 0 after the transition is complete to allow for a smooth transition once rain is turned back on.

mworzala (approved):

PR #2033, opened 2024-03-18
feat: add an end of tick scheduler

DeidaraMC:
a dedicated way to run logic at the end of tick in response to #2028

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -161,6 +161,10 @@ public final class MinecraftServer {
         return serverProcess.scheduler();
     }
 
+    public static @NotNull SchedulerManager getEndOfTickScheduler() {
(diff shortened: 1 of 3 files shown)


mworzala:
This needs to become a part of the Scheduler api, not a global concept. A different threading model may tick two instances on different threads meaning this is no longer a valid way to execute tasks at end of tick.

I am not sure how to incorporate it into schedulers exactly however

DeidaraMC:
> Reference in

We can integrate new methods into the scheduler manager i.e, scheduleAtEndOfTick()... etc
Ultimately you're still going to have to call another method to do the end of tick scheduling in a similar manner to how it's being done right now. Determining when a tick ends would be up to different threading model implementations. 
The current scheduler would also have these same challenges in a hypothetical multi-threaded instance ticking system so this shouldn't be a reason to exclude the feature

PR #2039, opened 2024-03-22
Properly handle HTTP requests

MrGeoTech:
Right now, Minestom just throws errors if you try to connect to the server port with a web client. This is stupid. We have a HTTP status code for this: 418. I propose that we fix Minestom bug and make it handle this common case properly.

Also, funny.


mworzala:
Works as intended

PR #2044, opened 2024-03-25
Lighting fix 35

reosfire:
Some light compute improvements + fix section `y` light send threshold.


txfsjyqm (approved):
Works a bit better for me than version 34 from @iam4722202468 

At least I can't find any issues with broken skylight now.

src/main/java/net/minestom/server/instance/heightmap/HeightMapContainer.java
```diff
@@ -0,0 +1,13 @@
+package net.minestom.server.instance.heightmap;
+
+import net.minestom.server.instance.block.Block;
+import org.jglrxavpok.hephaistos.nbt.NBTCompound;
+
+public interface HeightMapContainer {
```
mworzala, line 6: should be sealed permitting only the implementation

src/main/java/net/minestom/server/instance/heightmap/HeightMapContainerImpl.java
```diff
@@ -0,0 +1,58 @@
+package net.minestom.server.instance.heightmap;
+
+import net.minestom.server.instance.Chunk;
+import net.minestom.server.instance.block.Block;
+import org.jglrxavpok.hephaistos.nbt.NBT;
+import org.jglrxavpok.hephaistos.nbt.NBTCompound;
+
+import java.util.Map;
+
+public class HeightMapContainerImpl implements HeightMapContainer {
```
mworzala, line 10: should be package private

src/main/java/net/minestom/server/instance/heightmap/HeightMapContainer.java
mworzala: also mark apistatus internal

PR #2047, opened 2024-03-25
feat: entity synchronization overhaul

DeidaraMC:
Summary of changes
- Changed entity synchronization to use server ticks rather than duration/absolute time
- Stopped duplicate position packets from being sent by refreshPosition+synchronizePosition
- synchronizePosition now updates player head position when the yaw is different
- Synchronization period defaults to 20 ticks (regular Minecraft behaviour) 
- minestom.entity-synchronization-ticks system property to modify the default synchronization period
- setSynchronizationTicks to increase/decrease synchronization time on entities that need it
- renamed askSynchronization to synchronizeNextTick to increase clarity


mworzala (approved):

mworzala:
This change was long overdue, thanks.

PR #2052, opened 2024-03-26
feat: factor data support for darkness effects

DeidaraMC:
The current behaviour of darkness is to blind the player entirely, this is because factor data is essential for darkness to work correctly.
I've created a FactorData record for ease of use, additionally I've created a Potion#darkness helper method to give developers factor values akin to what vanilla does

### DIFF
--- a/src/main/java/net/minestom/server/potion/Potion.java
+++ b/src/main/java/net/minestom/server/potion/Potion.java
@@ -76,7 +76,7 @@ public record Potion(@NotNull PotionEffect effect, byte amplifier,
 
     public Potion(@NotNull NetworkBuffer reader) {
         this(Objects.requireNonNull(PotionEffect.fromId(reader.read(VAR_INT))), reader.read(BYTE),
-                reader.read(VAR_INT), reader.read(BYTE), reader.read(BOOLEAN) ? FactorData.fromNBT((NBTCompound) reader.read(NBT)) : null);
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/potion/Potion.java
```diff
@@ -54,17 +57,26 @@ public record Potion(@NotNull PotionEffect effect, byte amplifier,
...
      */
     public Potion(@NotNull PotionEffect effect, byte amplifier, int duration) {
-        this(effect, amplifier, duration, (byte) 0);
+        this(effect, amplifier, duration, (byte) 0, null);
     }
 
     public Potion(@NotNull NetworkBuffer reader) {
         this(Objects.requireNonNull(PotionEffect.fromId(reader.read(VAR_INT))), reader.read(BYTE),
-                reader.read(VAR_INT), reader.read(BYTE));
+                reader.read(VAR_INT), reader.read(BYTE), reader.read(BOOLEAN) ? FactorData.fromNBT((NBTCompound) reader.read(NBT)) : null);
```
TheMode, line 79: Can use `NetworkBuffe#readOptional`

mworzala (commented):
I think it would be better if there was a generic "PotionData" interface like we have with ParticleData in case Mojang adds more potion datas, and to be a bit consistent with the particle api.

In the same vein it would be nice if Potion had some similar alternate constructors like particles.

DeidaraMC:
1.20.5 solves this issue automatically

PR #2053, opened 2024-03-26
feat: add aerodynamics record and the capability to set custom horizontal air resistance

DeidaraMC:
Takes some of the bloat out of Entity and lets us further customize air resistance


DeidaraMC:
@TheMode the logic for when velocity packets get sent is slightly different, the onGround setting method matches more closely with how vanilla does it too

iam4722202468 (approved):

PR #2054, opened 2024-03-26
feat: improve server ticking accuracy, transition to Thread#sleep and…

DeidaraMC:
… compensate for ticks that take too long

- Transitioned from LockSupport#parkNanos to Thread#sleep
- Prevent sleeping when there's 16 or less ms remaining until the next tick on windows, this is because windows has very bad timer resolution when it tries to save battery (out of our control)
- When a tick takes more than 50ms, the additional time taken will be subtracted from the next tick

I suspect we may need nanoTime rather than timeMillis to handle cases where the tickrate doesn't divide evenly?
Want to get the ball rolling on this change

### DIFF
--- a/src/main/java/net/minestom/server/ServerProcess.java
+++ b/src/main/java/net/minestom/server/ServerProcess.java
@@ -143,6 +143,6 @@ public interface ServerProcess extends Snapshotable {
 
     @ApiStatus.NonExtendable
     interface Ticker {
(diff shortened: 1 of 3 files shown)


DeidaraMC:
I found clocks on Linux accurate enough to sleep 2ms before the next tick, no resolution concerns unlike windows

PR #2057, opened 2024-03-27
chore: remove legacy EntityUtils, refactor Entity#hasVelocity

DeidaraMC:
checking for a solid block, 1 block below the player is not necessary and bad behaviour when movementTick handles setting onGround with a proper physics check


kermandev:
Fixed in 5c050c66ca80cd1b5f97e3558f4ebbbea2ae3fa1

PR #2060, opened 2024-03-28
feat: proper passenger position handling for most entities, fix passengers not always being applied to the client, don't sync passengers

DeidaraMC:
Fix description: Passengers would get their updateNewViewer() forcibly called by the passenger loop in Entity#updateNewViewer, this applies the required packets correctly in theory. However, after this loop runs, the viewer engine still adds all the entities automatically which calls updateNewViewer again: sends a new spawn packet, overriding the passenger packets. To fix this I opted to use addViewer rather than updateNewViewer. Then, in the view engine I ensured that if the player can already see an entity, it returns early (senseless to call this method twice).

Comments on proper passenger position handling: Unfortunately, almost every vanilla entity has specific hardcoded values/actions for where to offset passenger entities. I've implemented a baseline for getting the offset values based on the passed in entity.
- Only a few entities are missing proper implementation, in general the server now has a better grip on where passengers are (for applying particles and collision handling to them, this change doesn't affect entity position visually on the client)

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -467,7 +467,9 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
         final Set<Entity> passengers = this.passengers;
         if (!passengers.isEmpty()) {
             for (Entity passenger : passengers) {
-                if (passenger != player) passenger.updateNewViewer(player);
+                if (passenger != player && !passenger.isViewer(player)) {
+                    passenger.addViewer(player);
+                }
             }
             player.sendPacket(getPassengersPacket());
         }
@@ -577,7 +579,7 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
             effectTick();
         }
         // Scheduled synchronization
-        if (ticks >= nextSynchronizationTick) {
+        if (vehicle == null && ticks >= nextSynchronizationTick) {
             synchronizePosition();
         }
     }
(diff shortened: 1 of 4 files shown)


DeidaraMC:
Ditching the complex solution after a discussion with matt about the new changes coming with 1.20.5 to make this much easier.
Just making a slight change to the passenger height logic for now to reflect the general changes made in 1.20 to vehicle heights

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1310,31 +1310,15 @@ public void refreshPosition(@NotNull final Pos newPosition) {
...
-     * @param passenger   The passenger to be moved
+     * @param newPosition the new position of this vehicle
+     * @param passenger   the passenger to be moved
      */
     private void updatePassengerPosition(Point newPosition, Entity passenger) {
         final Pos oldPassengerPos = passenger.position;
-        final Pos newPassengerPos = oldPassengerPos.withCoord(newPosition.x(),
-                newPosition.y() + getPassengerHeightOffset(),
-                newPosition.z());
+        final Pos newPassengerPos = oldPassengerPos.withCoord(newPosition.x(), newPosition.y(), newPosition.z())
```
TheMode, line 1320: Shouldn't the mutation be done in a single `withCoord` call like before? This version isn't more readable and does one more allocation

DeidaraMC: > Shouldn't the mutation be done in a single `withCoord` call like before? This version isn't more readable and does one more allocation

Yeah, that was a mistake from reverting the big change to passenger offsets I had initially made

mworzala (approved):

Issue #2061, opened 2024-03-29
BUG: Entity damage display


Swofty-Developments:
Minestom is using the wrong packet by default to display the red outline damage effect;

A DamageEventPacket should be used instead of the EntityAnimationPacket currently sent out in LivingEntity.java.

GreatWyrm:
@Swofty-Developments Should be fixed by d5c6126a6b8260dfc06033dc08bf6603bb8f1b57, can you confirm that it now works for you?

TheMeinerLP:
@Swofty-Developments Is it now fixed ?

PR #2063, opened 2024-03-29
fix: damaged entities not playing their animation, animation type for source-less damage

DeidaraMC:
We need to send a DamageEventPacket rather than an EntityAnimationPacket.
Additionally, an AnimationType needed to be added to let the damage animation be played in a similar style to pre-1.20 (damage tilt remains constant regardless of source)

fix for #2061

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -20,10 +20,7 @@ import net.minestom.server.inventory.EquipmentHandler;
 import net.minestom.server.item.ItemStack;
 import net.minestom.server.network.ConnectionState;
 import net.minestom.server.network.packet.server.LazyPacket;
-import net.minestom.server.network.packet.server.play.CollectItemPacket;
-import net.minestom.server.network.packet.server.play.EntityAnimationPacket;
-import net.minestom.server.network.packet.server.play.EntityPropertiesPacket;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -335,8 +332,13 @@ public boolean damage(@NotNull Damage damage) {
 
             float remainingDamage = entityDamageEvent.getDamage().getAmount();
 
-            if (entityDamageEvent.shouldAnimate()) {
-                sendPacketToViewersAndSelf(new EntityAnimationPacket(getEntityId(), EntityAnimationPacket.Animation.TAKE_DAMAGE));
+            EntityDamageEvent.AnimationType animation = entityDamageEvent.getAnimationType();
+            if (animation != EntityDamageEvent.AnimationType.NONE) {
+                int entitySourceId = damage.getSource() == null ? 0 : damage.getSource().getEntityId();
+                boolean withSource = animation == EntityDamageEvent.AnimationType.WITH_SOURCE;
+                DamageEventPacket damageEventPacket = withSource ? new DamageEventPacket(getEntityId(), damage.getType().id(), entitySourceId, entitySourceId, damage.getSourcePosition()) :
```
mworzala, line 339: use an if statement please, also some tests would be great

mworzala:
Closing as out of date, though some work is still needed here. See discussion on #2707

PR #2074, opened 2024-03-31
Add unstable API for custom Login Plugin Messages

Minikloon:
At present there is no way to inject custom Login Plugin Messages which can be useful for additional proxy<->server features negotiated at login time. https://wiki.vg/Protocol#Login_Plugin_Request

The usage looks like this:
image

Discussion points:
- I didn't want to add a breaking change to the API, so I didn't touch the existing pluginRequestMap, but it would make sense to shove that map inside the message box.
- The timeouts in ConnectionManager should probably be customizable (including the new LOGIN_PLUGIN_MESSAGE_TIMEOUT)
- I really don't mind how the API looks. It could be 1-2 new events instead, but I think this is the easiest to use for someone who is familiar with CompletableFutures.
- The handle method should probably not be exposed for users
- Since the plugin messages are part of the Minecraft protocol, the map and the message box should/could be moved to PlayerConnection directly.
- I'm myself not a fan of the name LoginPluginMessageBox
- I tested with Velocity's ServerLoginPluginMessageEvent

tldr: Custom Login Plugin Request/Responses, I don't care how, this is one way that works.


src/main/java/net/minestom/server/listener/preplay/LoginListener.java
```diff
@@ -163,47 +163,57 @@ private static SecretKey getSecretKey(byte[] sharedSecret) {
...
+        String channel = socketConnection.getPluginRequestChannel(packet.messageId());
+
+        LoginPluginMessageBox messageBox = socketConnection.getLoginPluginMessageBox();
+        try {
+            messageBox.handle(packet.messageId(), channel, packet.data());
+        } catch (Throwable t) {
+            MinecraftServer.LOGGER.error("Error handling Login Plugin Response", t);
+            LoginDisconnectPacket disconnectPacket = new LoginDisconnectPacket(ERROR_DURING_LOGIN);
+            socketConnection.sendPacket(disconnectPacket);
+            socketConnection.disconnect();
```
Minikloon, line 216: The original code doesn't have this disconnect, but I feel just sending the disconnect packet and relying on the client isn't enough?

Minikloon:
New usage when adding to a list within the AsyncPlayerPreLoginEvent:
image

src/main/java/net/minestom/server/event/player/AsyncPlayerPreLoginEvent.java
```diff
@@ -60,6 +66,24 @@ public void setPlayerUuid(@NotNull UUID playerUuid) {
         this.playerUuid = playerUuid;
     }
 
+    /**
+     * Changes the player uuid.
```
GreatWyrm, line 70: Copy paste error, please fix

PR #2077, opened 2024-04-02
feat: raycasting api

DeidaraMC:
A raycasting API integrated into Minestom.

The design philosophy behind this API is to collect as much information as possible from the raycast, then give the user tools in the resulting CastResult to easily grab the information they need. 

Cast a `Ray` against blocks, entities, or both simultaneously with an additional `Configuration` option to apply filters and more.

Code examples:
```java
// Grab the entities the ray intersected before it collided with a block
MinecraftServer.getGlobalEventHandler().addListener(PlayerUseItemEvent.class, e -> {
    Player player = e.getPlayer();
    Instance instance = e.getInstance();
    Pos pos = player.getPosition();

    // origin, direction, distance
    Ray ray = new Ray(pos.add(0, player.getEyeHeight(), 0), pos.direction(), 16, config -> {
        config.blockFilter(Block::isSolid);
        config.entityFilter(entity -> !(entity instanceof Player));
    });

    EntityBlockCastResult result = ray.cast(instance, instance.getEntities());
    result.findEntitiesBeforeBlockCollision().forEach(collision -> {
        //do something with collision.entity()
    });
}

// Shoot a flame at the point of ray intersection for each block, moving in the
// the direction of the surface normals
MinecraftServer.getGlobalEventHandler().addListener(PlayerUseItemEvent.class, e -> {
    Player player = e.getPlayer();
    Instance instance = e.getInstance();
    Pos pos = player.getPosition();

    Ray ray = new Ray(pos.add(0, player.getEyeHeight(), 0), pos.direction(), 16, config -> config.blockFilter(Block::isSolid));
    BlockCastResult result = ray.cast(instance);
    result.blockCollisions().forEach(collision -> {
        ParticlePacket entryFlame = new ParticlePacket(Particle.FLAME, collision.entry(), collision.entrySurfaceNormal(), 1, 0);
        ParticlePacket exitFlame = new ParticlePacket(Particle.SOUL_FIRE_FLAME, collision.exit(), collision.exitSurfaceNormal(), 1, 0);
        player.sendPacket(entryFlame);
        player.sendPacket(exitFlame);
    });
}
```

### DIFF
--- a/src/main/java/net/minestom/server/collision/BlockCastResult.java
+++ b/src/main/java/net/minestom/server/collision/BlockCastResult.java
@@ -1,0 +1,64 @@
+package net.minestom.server.collision;
+
+import net.minestom.server.coordinate.Point;
+import net.minestom.server.coordinate.Vec;
+import net.minestom.server.instance.block.Block;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.NoSuchElementException;
+
+/**
+ *  The result of a {@link Ray#cast(Block.Getter)}
+ */
+public interface BlockCastResult extends CastResult {
+    /**
+     * Whether this {@link BlockCastResult} has a block collision or not
+     *
+     * @return true if there is a block collision, otherwise false
+     */
+    boolean hasBlockCollision();
+
+    /**
+     * Returns the first block collision from the proceeding {@link Ray#cast}
+     *
+     * @return the first block collision
+     * @throws NoSuchElementException when the block cast result has no {@link BlockRayCollision}
+     */
+    @NotNull BlockRayCollision firstBlockCollision();
+
+    /**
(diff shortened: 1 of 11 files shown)


Tofaa2:
Is there a reason it takes an instance and the instance entities? It can be accessed by the method itself

DeidaraMC:
> Is there a reason it takes an instance and the instance entities? It can be accessed by the method itself

It takes any `Block.Getter` and any `Collection<Entity>`
This helps with making the API more versatile (preprocessing your list of entities when needed for example), take a look at the Ray record, depending on the cast method used you get a different interface back

TogAr2:
Are you aware there is already some raycasting done in Minestom physics (to check for collisions)? See `EntityCollision` and `BlockCollision` classes. I agree exposing things in an API could be helpful, but maybe it's better to adapt some of the existing code.

PR #2080, opened 2024-04-04
fix: virtual thread pinning (event node)

maximbreitman:
avoid synchronized block in favour of lock.


KermanIsPretty:
Virtual thread synchronization pinning should be fixed when minestom updates to java 25 LTS ([fixed in 24](https://openjdk.org/jeps/491)), unsure if its worth implementing a lock since after the first iteration its should be fast enough.

kermandev:
No longer a issue since release https://github.com/Minestom/Minestom/releases/tag/2025.10.11-1.21.10

PR #2082, opened 2024-04-05
Add a Scheduler::scheduleTask with period supplier AND delay

Minikloon:
At the moment there's no easy API to schedule a task which is in charge of its own cancellation.

The functionality is in Task::Builder so might aswell offer it.


mworzala (approved):
I don't really love `repeatOverride`, but I didnt come up with a better option.

PR #2095, opened 2024-04-13
chore: eof instead when disconnected

maximbreitman:


TheMode:
Fixed in `next` branch

PR #2097, opened 2024-04-14
Fix: Server does not stop

### LINKED ISSUE #2096: Server doesn't stop correctly
Minestom server doesn't stop correctly when using `MinecraftServer#stopCleanly`. Server's threads are still running.

I recorded a video showing the issue:
https://www.youtube.com/watch?v=QJkS8tqTKpA

devreluije:
Problem: Server doesn't stop when using `MinecraftServer#stopCleanly` inside a command

Resolves #2096 

Why this bug happens ?
During the server lifecycle commands are executed in `connection().tick(msTime)` (`ServerProcessImpl.TickerImpl` class)
image

When you use the _/shutdown_ command, `MinecraftServer#stopCleanly` method is called which stops all `TickThread`
After all commands are executed, `serverTick(msTime)` is called which start (stopped TickThread) and wait until threads are finished

Solution: set a flag to true when stop method is called and check this flag at end of tick

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -194,6 +194,10 @@ public final class MinecraftServer {
         return !isStarted();
     }
 
+    public static boolean isStopped() {
+        return serverProcess.isStopped();
+    }
+
     /**
      * Gets the chunk view distance of the server.
      * <p>
--- a/src/main/java/net/minestom/server/ServerProcess.java
+++ b/src/main/java/net/minestom/server/ServerProcess.java
@@ -137,9 +137,28 @@ public interface ServerProcess extends Snapshotable {
 
     void start(@NotNull SocketAddress socketAddress);
 
+    /**
+     * Request the server to stop
+     */
     void stop();
 
+    /**
+     * Shutdown all sub processes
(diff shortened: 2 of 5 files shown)


thebadlorax:
i concur, when i made my /stop command i had to just put a System.exit(0) at the end.

cowtorode:
Why don't you just schedule the call to stopCleanly?

LoboMetalurgico:
I agree, this would be very helpful. In fact, recently, the reason for this strange behavior was being discussed on discord.

cowtorode:
Yeah. And you don't necessarily have to do it within stopCleanly, but there should be an API note that you can't call this method from within a method that's called during server ticks without scheduling.

LoboMetalurgico:
> Yeah. And you don't necessarily have to do it within stopCleanly, but there should be an API note that you can't call this method from within a method that's called during server ticks without scheduling.

I'm sorry, but if a method that should shut down the server doesn't shut it down, that's a bug, not a feature.

cowtorode:
I understand this, but you don't have to schedule the call to stopCleanly for every call. It therefore doesn't make sense to change the stopCleanly method to make this implementation.

mworzala:
Fixed in #2445

Issue #2076, opened 2024-04-01
Ability to hide player from Tab


MelonHell:
We are making Tab with custom data and real players on the list are interfering with us. We're currently using this hack https://github.com/SkolkovoLab/Minestom/commit/0ec435bb3180e4ceec49b36551f22e4739da734c, but we'd like Minestom to be able to hide players from the Tab

PR #2113, opened 2024-04-29
feat: player predicate to control who to send info entries to

DeidaraMC:
Addresses issue #2076
Currently Minestom has no way to prevent info entries from being sent to players through the API.

This change introduces a predicate to implement custom logic against who should receive entry packets from a player

Simple example, only send entries to other players who are in a specific list
`player.setInfoEntryUpdateRule(receiver -> list.contains(receiver))`

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -79,7 +79,6 @@ import net.minestom.server.snapshot.SnapshotUpdater;
 import net.minestom.server.statistic.PlayerStatistic;
 import net.minestom.server.timer.Scheduler;
 import net.minestom.server.utils.MathUtils;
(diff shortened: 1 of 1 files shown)


Steanky:
Would it be better to raise an event instead of using a predicate? Benefits are multiple listeners can modify the tablist recipients. Could be something like `PlayerTablistAddEvent`, `PlayerTablistRemoveEvent`, both would include a modifiable List<Player> representing the players receiving the packets. Also wouldn't have to evaluate a predicate against every player on the network every time someone joins.

TheMeinerLP:
@DeidaraMC Can you please update your PR ?

SorkoPiko:
this would be much better with an event, as @Steanky suggested. as this pr is dead anyway, i'll re-implement these changes using events instead.

SorkoPiko:
superseded by #2980

PR #2115, opened 2024-04-30
Use int in Block#fromStateId

SoSeDiK:
Changes `Block#fromStateId` to accept `int` instead of `short` (protocol uses VarInt) and removes now unnecessary casts.


iam4722202468:
but why though

MelonHell:
I agree with this PR. Why is there a short cast here?

SoSeDiK:
> but why though

The question is rather opposite, why is there short?
Every single other place uses an int (including protocol) except this one single redirect that passes into a method that also accepts an int.
I was irritated by always having to cast to short (only for it to be passed as int again), so I've asked on Discord why is this the case. I've got a response from @TheMode that it should be changed to an int.

PR #2122, opened 2024-05-01
feat: LivingEntity fire overhaul

DeidaraMC:
- Fire duration uses ticks
- Setting fire through the entity meta will not expire it immediately
- Renamed EntityFireEvent > EntitySetFireEvent
- Added an EntityFireExtinguishEvent for when fire is extinguished
- Removed fire damage period, it was was not being used anyway
- Fire related unit tests


mworzala:
it feels like we should not lose the behavior of an event while the player is on fire. That seems like the most common case (like ticking health down)

mworzala (approved):

PR #2123, opened 2024-05-03
fix: automatically set instance when using Player.spectate

oglassdev:
This PR makes Player.spectate automatically set the instance, just like how addPassenger sets the instance. This should help with uniformity


mworzala:
I dont think this behavior should exist, it should instead just be an error if they are not in the same instance. Passengers should also work this way, its odd to have this implicit switch to me.

PR #2125, opened 2024-05-03
Fix Aerodynamics#withGravity

TogAr2:
Previously, the method would change the horizontal air resistance instead of the gravity value


DeidaraMC (approved):
Looks good

PR #2089, opened 2024-04-10
1.20.5

mworzala:
Work in progress 1.20.5 branch. All of the item components have a (largely untested) implementation and most functionality seems to be there for items and other packet changes.

This branch also includes the following:
- #1992 (will merge just before this branch)
- #2030

Immediately pressing issues:
- Tests needed for the component implementations
- Dynamic registries all need to be changed to work more like Biome, as well as load the vanilla data and track modifications to decide what data to send (new known packs exchange).

Transfer Packet:
- I am aiming to finish and merge this branch _without_ support for transfer packets. This will come in a future update.

Backwards compatibility/upgrading:
- Currently no backwards compatibility is maintained in this branch. I could make the item component changes backwards compatible, however I don't believe it's worth the effort. There are significant other changes here which are non backwards compatible.
- I will be writing an upgrade guide for this version bump.

This branch is now published on Maven Central at `net.minestom:minestom-snapshots:1_20_5-{first 10 chars of a commit hash}`

PR #2128, opened 2024-05-04
Item usage duration

TogAr2:
This pr generalizes the food code in `Player` and `UseItemListener` to also allow for non-food items to have a usage duration specified beforehand.

While this does make `PlayerPreEatEvent` and `PlayerEatEvent` redundant (the same can be done with `PlayerUseItemEvent` and `ItemUpdateStateEvent`), I left those in for clarity and backwards compatibility. For the same reason I left `Player#isEating`, which now just checks if the currently used item is a food. Please let me know if these should be removed instead.

Along with those changes I added spyglass, horn and brush item animation types, since Minestom already had support for bow, crossbow, trident, etc. I also removed the deprecated `Player#callItemUpdateStateEvent(boolean, Hand)` since it has been deprecated for a long time and I think it was time for it to be removed.


mworzala:
I merged this into #2089

PR #1992, opened 2024-02-11
Inventory Rework

GoldenStack:
Rework of the inventory system.
Recreation of https://github.com/hollow-cube/minestom-ce/pull/132

Main changes:
- Accurate click handling (except for special slots like result slots)
- Improved event API: allows accessing all click information via Click.Info, as well as providing a list of all changes via Click.Result
- Removed InventoryCondition in favor of three events: InventoryPreClickEvent, modifying only click info, like slot IDs; InventoryClickEvent, modifying the list of changes that will occur as a result of the click; and InventoryPostClickEvent, which can view all the click information but happens after handling is finished

Changes:
- Added InventoryButtonClickEvent
- Moved a lot of behaviour into AbstractInventory, renamed AbstractInventory to Inventory and Inventory to ContainerInventory
- Inventory events return the relevant PlayerInventory instead of null in #getInventory
- Added the tagged union Click.Info containing all necessary click info (e.g. slots)
- Made all inventories implement Viewable
- Simplified click handling and implemented vanilla parity other than special slots (like result slots) in ClickProcessors
- Moved click packet parsing & drag handling to each player's Click.Preprocessor
- Added Click.Result which stores a map of all changes and side effects that could occur
- Refactored InventoryCondition into InventoryClickEvent, and renamed what was formerly InventoryClickEvent into InventoryPostClickEvent
- More comprehensive tests
- Use ReentrantLock instead of Java synchronization
- Made cursor items PlayerInventory-based
- Greatly simplified inventory transaction logic
- Creative click logic and drop/swap logic in PlayerDiggingHandler is handled correctly (also removed PlayerSwapItemEvent)

PR #2137, opened 2024-05-11
Better shift click button differentiation

ItsFelix5:
This change allows for more control over clicks in UIs.
Also replaced an if jungle in the window click listener with switch.


mworzala:
I believe this is addressed by https://github.com/Minestom/Minestom/pull/1992, which is being merged into the 1.20.5 branch already.

PR #2142, opened 2024-05-18
Hand handling improvement

Codestech1:
This pull request:
- Allows to swing a hand for a living entity
- Calls an `EntitySwingHandEvent` in `LivingEntity#swingHand(Hand)` method **instead of** in a packet listener.

I think that it would be nice if this pull request was merged, because we could avoid messy code like [this](https://github.com/Hypejet/Minestom-Combat/blob/6e5a1b0588120b1008d99efaa31ad9ac2d86c341/src/main/java/net/hypejet/combat/player/CombatPlayer.java#L60-L65).


src/main/java/net/minestom/server/event/player/EntitySwingHandEvent.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minestom.server.event.trait.CancellableEvent;
+import net.minestom.server.event.trait.EntityInstanceEvent;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Called when a living entity swings its hand.
+ *
+ * @see LivingEntity#swingHand(LivingEntity.Hand)
+ */
+public final class EntitySwingHandEvent implements EntityInstanceEvent, CancellableEvent {
```
mworzala, line 13: Should not be in the `event.player` package.

Though more generally, I am unclear on the need for this to be an event. Whats the use case?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2411,14 +2411,6 @@ protected void sendChunkUpdates(Chunk newChunk) {
         return super.teleport(position, chunks, flags);
     }
 
-    /**
-     * Represents the main or off hand of the player.
-     */
-    public enum Hand {
```
mworzala, line 2417: I think this should stay on Player. Even if it is slightly incorrect (eg for LivingEntity#swingHand to take a Player.Hand), I believe most people will look for player hand. Also makes this change less breaking

src/main/java/net/minestom/server/event/player/EntitySwingHandEvent.java
Codestech1: I will move the event to another package. But the use case of this is to cancel behavior, such as sending a hand swing packet to viewers. Another use case might be my minestom combat implementation, we need this event to reset the attack strength for a player.

src/main/java/net/minestom/server/entity/Player.java
Codestech1: I don't know if it's that breaking, using `Player.Hand` is still correct. Do you still want this enum to be in a player class?

PR #2143, opened 2024-05-20
Add PlayerList API

Minikloon:
Currently Minestom shares the player list (the tablist) with the whole instance.
This PR is an example of an API allowing devs to customize this behavior.


Obyvante:
Are you planning to add Javadoc to your class and methods?

PR #2144, opened 2024-05-22
Add BlockVec as a Point implementation

Minikloon:
From Discord talk: https://discord.com/channels/706185253441634317/706186227493109860/1242848057578618941

Summarized, the purpose is to leverage the type system to offer a guarantee that a Point is grid-aligned.


mworzala:
Please add a note in the javadoc about how it is semantically only returned to indicate a floored position, never taken as a parameter. then im good with it.

PR #2146, opened 2024-05-23
Acquirable improvements

TheMode:
Random improvements, ideally with more example to ease usage.


mworzala:
(closing since we have follow up now)

PR #2149, opened 2024-05-25
Retain offset in BoundingBox::contract and expand

Minikloon:
This is a breaking change to those methods but I expected those to behave like this so much that I think it's a bug.


mworzala:
This is a really 'sneaky' breaking change and would be frustrating to debug. I think there should simply be some new methods added to work this way and javadocs added explaining the difference rather than breaking here.

PR #2156, opened 2024-05-29
Offer more control over LightingChunk invalidations

Minikloon:
I'm doing a lot of block updates at once and calling the invalidation every setBlock shows up on the profiler.
This allows doing the invalidation later all at once.

### DIFF
--- a/src/main/java/net/minestom/server/instance/LightingChunk.java
+++ b/src/main/java/net/minestom/server/instance/LightingChunk.java
@@ -112,7 +112,7 @@ public class LightingChunk extends DynamicChunk {
         this.freezeInvalidation = freezeInvalidation;
     }
 
-    public void invalidateSection(int coordinate) {
+    public void invalidateNeighborsSection(int coordinate) {
         if (freezeInvalidation) {
             return;
         }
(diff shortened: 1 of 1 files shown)


mworzala:
seems fine to me, but ill defer to @iam4722202468

src/main/java/net/minestom/server/instance/LightingChunk.java
```diff
@@ -107,7 +108,15 @@ private boolean checkSkyOcclusion(Block block) {
         return occludesBottom || occludesTop;
     }
 
-    private void invalidateSection(int coordinate) {
+    public void setFreezeInvalidation(boolean freezeInvalidation) {
+        this.freezeInvalidation = freezeInvalidation;
+    }
+
+    public void invalidateSection(int coordinate) {
+        if (freezeInvalidation) {
```
demkom58, line 116: Why to do check here if this one already is in `setBlock`? 
Also freeze sounds to me, as it should automatically invalidate all things after being unfreezed.

Minikloon: 1. because it's now public
2. That's a large side effect for a setter, plus the caller maybe decided to do no changes, but i don,t mind that much

PR #2167, opened 2024-05-30
Move static Entity lookup maps to EntityTracker

Minikloon:
EntityTrackerEntry exists to ensure the same behavior as before in regards to the recorded position in the tracker.

### DIFF
--- a/src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
+++ b/src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
@@ -8,6 +8,7 @@ import net.minestom.server.coordinate.Vec;
 import net.minestom.server.entity.Entity;
 import net.minestom.server.entity.Player;
 import net.minestom.server.utils.chunk.ChunkUtils;
+import net.minestom.server.utils.validate.Check;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.annotations.Unmodifiable;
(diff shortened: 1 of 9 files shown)


mworzala:
Putting an entity in an instance which already contains that ID or UUID should throw an exception

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
```diff
@@ -106,17 +138,23 @@ public void remove(@NotNull T removed) {
         }
     }
 
+    @Override
+    public void changeUuid(@NotNull Entity entity, UUID oldUuid) {
```
mworzala, line 142: this is not thread safe and missing an annotation, but honestly we should just remove the setUuuid method from entity and this functionality and make it final. It doesn't make sense to change an entity uuid after creating it I dont think.

Minikloon:
> Putting an entity in an instance which already contains that ID or UUID should throw an exception

It will from the (using Checks) assertion in EntityTracker.

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
Minikloon: The only use within Minestom of setUuid within Minestom is transitionLoginToConfig
The commit message said thread-safety, I meant safer than register/unregister.

I agree with removing setUwuid, but I think it's beyond the scope of this PR.

mworzala (approved):

PR #2179, opened 2024-06-09
[RE] Update PlayerChangeHeldSlotEvent.java

unjoinable:
Slight changes in the docs since they were mis-leading.. Took me a while to realise what was wrong in my algorithm

the held slot -> future slot

info ->  old post got deleted because it was on master branch

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerChangeHeldSlotEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerChangeHeldSlotEvent.java
@@ -25,7 +25,7 @@ public class PlayerChangeHeldSlotEvent implements PlayerInstanceEvent, Cancellab
     /**
      * Gets the slot which the player will held.
      *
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/event/player/PlayerChangeHeldSlotEvent.java
```diff
@@ -25,7 +25,7 @@ public PlayerChangeHeldSlotEvent(@NotNull Player player, byte slot) {
     /**
      * Gets the slot which the player will held.
```
MatzHilven, line 26:
```suggestion
     * Gets the slot which the player will hold.
```
You didn't change this part, but is this grammar intended 🤔

unjoinable: Sorry i dont have the best english but yes thats better

PR #2184, opened 2024-06-10
Add getTag overload with fallback value

mudkipdev:


mworzala:
The expected way to handle this is setting the defaultvalue on the tag. I don't think it makes sense to have two ways of the same behavior (which are kind of confusing with each other)

PR #2188, opened 2024-06-11
feat: added packet preprocessor

sweet333:
It is often necessary to replace packets before sending them. The preprocessor solves this problem.


Codestech1:
> Please do not open pull requests from the `master` branch, create a new branch instead.

💀

PR #2189, opened 2024-06-11
1.21

mworzala:
Working branch for 1.21 support.

Known missing changes:
* Enchantment (dynamic) registry


src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -159,6 +175,46 @@ public ItemStack read(@NotNull NetworkBuffer buffer) {
...
+    }
+
+    @Contract(value = "_ -> new", pure = true)
+    default @NotNull ItemStack withCustomModelData(int customModelData) {
+        return with(ItemComponent.CUSTOM_MODEL_DATA, customModelData);
+    }
+
+    @Contract(value = "_ -> new", pure = true)
+    default @NotNull ItemStack withGlowing(boolean glowing) {
+        return with(ItemComponent.ENCHANTMENT_GLINT_OVERRIDE, false);
```
FluxCapacitor2, line 200:
```suggestion
        return with(ItemComponent.ENCHANTMENT_GLINT_OVERRIDE, glowing);
```
Should this use the parameter value instead of `false`?
https://minecraft.wiki/w/Data_component_format#enchantment_glint_override

mworzala: Good catch, thanks

PR #2190, opened 2024-06-12
Add Instance::getPlayerByUuid for convenience

Minikloon:
It's a common query.


TheMode:
Nitpick but it seems like you could use a ternary operator instead

Minikloon:
image

⛔ It's worse!

mworzala:
For the record, I think its better too :)

PR #2191, opened 2024-06-13
fix full light data is not sent

bea4dev:
Fixed an issue where full light data was not sent when re-logging in.
> Before

> After


iam4722202468:
Is there a reason to store a cached full light packet in dynamic chunk? Looks like the only place it's used is in the chunk packet, where it will be cached

bea4dev:
> Is there a reason to store a cached full light packet in dynamic chunk? Looks like the only place it's used is in the chunk packet, where it will be cached

You're right, I don't think that's necessary.
I'll fix it.

bea4dev:
Fixed

PR #2193, opened 2024-06-14
Add feature flags to AsyncPlayerConfigurationEvent

Flopgop:
PR 2 because github actions is a big meany and I forgot to create a new branch on my PR.

(Adds the ability to add and remove feature flags during the AsyncPlayerConfigurationEvent on a per basis without requiring the user to send packets. By default includes the "minecraft:vanilla" feature flag, to keep vanilla-esque behavior. No feature flag namespaces are hardcoded anywhere besides vanilla, mainly because I did not want to deal with data generator or keeping that maintained.)


mworzala (approved):
The feature flags probably should be generated, but it can be done in the future.

PR #2194, opened 2024-06-14
Addition: PlayerOnGroundChangeEvent

hapily04:
This pull request allows people to more easily track the on-ground value of a player, which can be helpful for fall damage implementations. Additionally, it provides context for whether the client claims they are on the ground or if the server does.


GreatWyrm:
I'd recommend against making this a cancellable event, as I think that could lead to a lot of movement jitter.

hapily04:
> I'd recommend against making this a cancellable event, as I think that could lead to a lot of movement jitter.

I think I made this cancellable when I thought you could set the value yourself (kind of like how you can set if a player is sprinting or not), but you can't. I also didn't look far enough to see that onGround is always just based on what the client says, so I'll be removing that part of the event too.

PR #2195, opened 2024-06-16
Update Player.java

unjoinable:
Removed 3 not needed  @ApiStatus.Experimental from Player.class not sure about interpretPacketQueue();


unjoinable:
Removed the 2 from PlayerConnection as well

mworzala (approved):

PR #2198, opened 2024-06-16
Fix add transaction

iam4722202468:
Currently `Inventory::addItemStack` will fill air slots with the entire item stack size, ignoring stacking rules.
This fix makes item stack rules not get ignored when using `addItemStack`

It's worth noting that `addItemStack` doesn't ignore stacking rules if it finds the item and is adding to it.


GoldenStack:
Merged after hours of stressful and heated delibration.

PR #2201, opened 2024-06-17
Autogenerate Feature Flags

Flopgop:
Do not pull until MinestomDataGenerator 1.21-rc2 is released; this depends on a JSON generated by it and has that version set in the `libs.versions.toml`. Without it __this code will not work__.

Adds autogeneration of Feature Flags into a static constant file at `net.minestom.server.FeatureFlags`. 
~~TODO: Check if referencing `net.minestom.server.FeatureFlags` in the Javadoc of AsyncPlayerConfigurationEvent breaks it.~~

~~Also, can someone verify that I'm autogenerating the class correctly and if there's not a better way for me to do it? Thanks.~~
edit: Rewrite removes the possibility of Javadoc breaking and I now believe I am autogenerating the code correctly.


Flopgop:
It has come to my attention that the naming convention for data gen is not `rv1` for the first revision of the data gen and it's actually based on minestom versioning.
still my update needs to be pushed lol

Flopgop:
For merge conflicts, please accept my changes over master unless something new has been added to these files since [f1d5940](https://github.com/Minestom/Minestom/commit/f1d5940855b1370c24ef0ac54cd10768e8ba34ee).

mworzala:
this needs to be **rebased** on the latest master, also please add a factory method to FeatureFlag for people to make their own (mods could add them I guess)

Flopgop:
I will do that in a moment

code-generators/src/main/java/net/minestom/codegen/FeatureFlagGenerator.java
TheMeinerLP, line 1: Is that still needed ?

code-generators/src/main/java/net/minestom/codegen/Generators.java
```diff
@@ -25,6 +25,7 @@ public static void main(String[] args) {
         new RecipeTypeGenerator(resource("recipe_types.json"), outputFolder).generate();
         new ParticleGenerator(resource("particles.json"), outputFolder).generate();
         new ConstantsGenerator(resource("constants.json"), outputFolder).generate();
+        //new FeatureFlagGenerator(resource("feature_flags.json"), outputFolder).generate();
```
TheMeinerLP, line 28: Remove comment out code

src/main/java/net/minestom/server/FeatureFlagImpl.java
```diff
@@ -0,0 +1,33 @@
+package net.minestom.server;
+
+import net.minestom.server.registry.Registry;
+import net.minestom.server.utils.NamespaceID;
+import org.jetbrains.annotations.NotNull;
+
+public record FeatureFlagImpl(@NotNull Registry.FeatureFlagEntry registry) implements FeatureFlag {
```
TheMeinerLP, line 7: Add some documentation

TheMeinerLP (changes requested):

src/main/java/net/minestom/server/FeatureFlagImpl.java
Flopgop: None of the other data-generated impl classes have documentation. This is also an internal class that I forgot to make package-private.

Flopgop:
Everything in order :+1:

Flopgop:
> > this needs to be **rebased** on the latest master, also please add a factory method to FeatureFlag for people to make their own (mods could add them I guess)
> 
> Both parts of this comment still need to be done

I will not be adding a builder as mods can add an entry to the json (as they would for a Block or Entity.)
I rebased with intellij and pushed, however it doesn't seem to have done anything.

mworzala:
This is not a strictly static protocol entry like blocks or entities, I don't see a good reason to restrict people to only entries declared in the registry json. It's definitely a grey zone because it's also not sent to the client dynamically, but I'd prefer to avoid adding new static registries based only on the json files. They aren't very intuitive.

PR #2202, opened 2024-06-17
Add fields for kicking players in AsyncPlayerConfigurationEvent

GreatWyrm:
This updates the AsyncPlayerConfigEvent with 2 new fields, a 'ConfigurationResult' field and a 'KickMessage' field.

This brings the behavior closer to [Spigot's equivalent event](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/player/AsyncPlayerPreLoginEvent.html), and I believe the Spigot version is more useful because it provides a more obvious way of kicking people during the join process, where you can set the status in the event rather than the unintuitive method of calling 'player.kick()' during the event.

The previous behavior of kicking during the event will still work, or you could set the ConfigurationResult so Minestom will kick them for you, and an optional field to provide a custom kick message or use the default one.

### DIFF
--- a/src/main/java/net/minestom/server/event/player/AsyncPlayerConfigurationEvent.java
+++ b/src/main/java/net/minestom/server/event/player/AsyncPlayerConfigurationEvent.java
@@ -2,6 +2,8 @@ package net.minestom.server.event.player;
 
 import it.unimi.dsi.fastutil.objects.ObjectArraySet;
 import it.unimi.dsi.fastutil.objects.ObjectSets;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.format.NamedTextColor;
 import net.minestom.server.entity.Player;
 import net.minestom.server.event.trait.PlayerEvent;
 import net.minestom.server.instance.Instance;
@@ -31,6 +33,8 @@ public class AsyncPlayerConfigurationEvent implements PlayerEvent {
     private boolean clearChat;
     private boolean sendRegistryData;
     private Instance spawningInstance;
+    private ConfigurationResult configurationResult;
+    private Component kickMessage;
 
     public AsyncPlayerConfigurationEvent(@NotNull Player player, boolean isFirstConfig) {
         this.player = player;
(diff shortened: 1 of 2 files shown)


iam4722202468:
Not a fan of this, since it implies that using `KickMessage` and `ConfigurationResult` are the correct ways to kick the player, when all they do is provide another way of kicking the player.
To me it feels more intuitive to use `Player::kick` instead of setting a field in the event. I've never used that api in spigot, seems like a weird design choice...

GreatWyrm:
I'd argue it's easier to understand from a coder that's just jumped into Minestom, as they'd search for a way to stop the player during the config phase and they see some provided fields to use. I certainly was caught off guard that there's no way to inform the player login process that it should halt, and that you have to kick the player yourself.

As for it being implied that it's the 'correct' way to do so, I guess? It feels a bit more library sanctioned with this change, but as I noted in my PR description, both ways will work.

Also, it's easier to tell what exactly disconnected the player if you have multiple listeners on to the Async Event, if one of them kicks the player, the other one would be hard pressed to know exactly why it happened (they could tell it happened with Player::isOnline()).

DasBabyPixel:
I am also against this, because players can enter configuration phase during game and not only when joining. Some documentation on the event saying "use Player#kick" should suffice imo.

GreatWyrm:
> I am also against this, because players can enter configuration phase during game and not only when joining. Some documentation on the event saying "use Player#kick" should suffice imo.

I don't see how that's an issue? Even if you re-enter configuration phase while the player has already connected, this PR works perfectly fine. You could set the event to kick them, but that would be a bit strange (I'd use Player#kick() in that instance).

At minimum there should be that documentation. I think providing some convenience methods in the event itself so it's more clear to developers how to prevent someone from joining is a worthwhile improvement.

mworzala:
I believe I would also prefer to simply document that `Player#kick` is the appropriate way to handle this. Unless there is some use case where kick is insufficient, to me 1 api for this is better than adding a new one.

PR #2204, opened 2024-06-19
Improve notification API

mudkipdev:
Delete `NotificationCenter` and move send notification methods to `Player` and `PacketGroupingAudience` (breaking)


mworzala:
I'm probably good with this change (may have minor nitpicks) but ill look more thoroughly when closer to 1.21.1 which is when this will be merged most likely.

DeidaraMC:
I'm happy with this api change

PR #2209, opened 2024-06-23
Fix LivingEntity Damage Event Packet

GreatWyrm:


mworzala (approved):
Oops this slipped through 1.20.6, thanks.

PR #2210, opened 2024-06-24
Separate lib module

TheMode:
Isolate independent modules from the Minestom framework and its opinionated control flow.

Files would have to be slowly moved over (Palette, Registry, Packet). Shouldn't be done in one go.

Goals are:
* Encourage more features to be independent from `Entity`/`Instance`
* Better organize our utils by considering them as their own library
* Make the future `scratch` module/repo more appealing


MelonHell:
<3

mworzala:
Marking ready for review so the ci triggers

PR #2217, opened 2024-06-25
feat: proper passenger handling for entities

DeidaraMC:
Quite a few changes being made in this PR.
- Exposing vehicle/passenger attachment offsets through the registry
   - EntityUtil#getPassengerAttachment and EntityUtil#getVehicleAttachment handle more entity specific/niche logic to give you the exact expected attachment points for an entity
- Renamed AgeableMobMeta to AbstractAgeableMeta to better fit the naming scheme
- Added an AgeableMeta interface to encompass is/setBaby on all ageable mobs (not all ageable mobs extend AbstractAgeableMeta)
- AgeableMeta#setBaby now forcefully overrides custom set bounding boxes to the bounding box the client expects

A small handful of entities aren't handled yet, camels, boats, item frames.
Custom mob scaling handling will come in a near soon pr


src/main/java/net/minestom/server/entity/metadata/AgeableMeta.java
```diff
@@ -0,0 +1,12 @@
+package net.minestom.server.entity.metadata;
+
+public interface AgeableMeta {
+    boolean isBaby();
+
+    /**
+     * Forcefully sets the bounding box of this entity to the bounding box
+     * which the client expects.
```
GreatWyrm, line 8: Javadocs seem somewhat incomplete? It should also mention that it changes the entity's baby status, if I looked at this method and it only said that it changed the bounding box I'd be a bit confused

src/main/java/net/minestom/server/entity/metadata/AbstractAgeableMeta.java
```diff
@@ -5,31 +5,26 @@
...
-            BoundingBox bb = entity.getBoundingBox();
-            if (value) {
-                double width = bb.width() / 2;
-                entity.setBoundingBox(width, bb.height() / 2, width);
-            } else {
-                double width = bb.width() * 2;
-                entity.setBoundingBox(width, bb.height() * 2, width);
-            }
+            BoundingBox bb = entity.getEntityType().registry().boundingBox();
+            if (value) entity.setBoundingBox(bb.width() * 0.5, bb.height() * 0.5, bb.depth() * 0.5);
```
GreatWyrm, line 26: I think this could be better done with something along the lines of `protected abstract double boundingBoxMultiplier();`, and then certain metas override that instead of repeating this chunk of code.

DeidaraMC: I'm not a fan of this. Not all entities extend AbstractAgeableMeta for a reason unfortunately (thanks mojang). Making metadata extend this arbitrary class would add api confusion imo. Could be delegated to the EntityUtils class but I think the current refactor is already a good start

PR #2218, opened 2024-06-25
chore: send keepalive packets immediately, simplify PacketProcessor#process

DeidaraMC:


MelonHell:
It would be cool to see the ability to send any packet immediately. For example, sometimes my animations have more than 20 fps, but I don’t want to increase the server’s TPS for them

mworzala (approved):

PR #2219, opened 2024-06-25
chore: fix EntityLiving#damage animation not cancellable, function/event refactor

DeidaraMC:
A fix for cancelling the EntityDamageEvent's animation properly (DamageEventPacket needs to not be sent.)
Cleaned up some of the code nesting in this function as well.
Changed EntityDamageEvent#setSound to supply an actual sound instead of just the SoundEvent with fixed fields.
Removed the concept of Sound from the Damage class, I don't think sound belongs there and it wasn't providing correct sound behaviour anyway.


mworzala:
Unless the sound change is extracted to a different PR, this is going in the 1.21.1 milestone because it's breaking. But I would prefer to do the animation part earlier

PR #2221, opened 2024-06-26
feat: Improvements to AttributeInstance API

Steanky:
Some small non-breaking changes to AttributeInstance:
* `addModifier(AttributeModifier)` returns the modifier that was replaced by the new one, or null if none exists
* Both `removeModifier(AttributeModifier)` and `removeModifier(NamespaceID)` return the modifier that was removed, or null if none was removed
* New method, `getValueWithBase(double)` for computing what the value of the AttributeInstance would be if it had a given base, without actually needing to set the base 

Justification:
* First two changes are mostly self-explanatory, it's useful to be able to know what a modifier did if you're removing or replacing it.
* Last change is useful for custom attributes that don't have a set "base value" that needs to be stored. For example, a damage type resistance API where the "base value" is just the damage that's being received, which is then dynamically modified based on AttributeModifiers.

### DIFF
--- a/src/main/java/net/minestom/server/entity/attribute/AttributeInstance.java
+++ b/src/main/java/net/minestom/server/entity/attribute/AttributeInstance.java
@@ -96,31 +96,40 @@ public final class AttributeInstance {
      * Add a modifier to this instance.
      *
      * @param modifier the modifier to add
+     * @return the old modifier, or null if none
      */
-    public void addModifier(@NotNull AttributeModifier modifier) {
-        if (modifiers.putIfAbsent(modifier.id(), modifier) == null) {
+    public AttributeModifier addModifier(@NotNull AttributeModifier modifier) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/entity/attribute/AttributeInstance.java
```diff
@@ -133,11 +142,18 @@ public double getValue() {
...
     /**
-     * Recalculate the value of this attribute instance using the modifiers.
+     * Gets the value of this instance, calculated assuming the given {@code baseValue}.
+     *
+     * @param baseValue the value to be used as the base for this operation, rather than this instance's normal base
+     *                  value
+     * @return the attribute value
      */
-    private void refreshCachedValue() {
+    public double getValueWithBase(double baseValue) {
```
mworzala, line 151: I think something like `applyModifiers(double value)` would be a clearer name

PR #2222, opened 2024-06-26
Ability to teleport player without teleport confirm

### LINKED ISSUE #2220: Teleport confirmation issue
On our server there is weapon recoil, which moves the player's camera with relative teleports. While shooting, the player model will often suddenly "teleport" as some movements are ignored

The problem is in these lines
https://github.com/Minestom/Minestom/blob/d5c6126a6b8260dfc06033dc08bf6603bb8f1b57/src/main/java/net/minestom/server/listener/PlayerPositionListener.java#L47-L49

If you disable the teleport confirmation system (that is, simply delete these lines), this will greatly aggravate the [problem of falling through the world #2017](https://github.com/Minestom/Minestom/issues/2017)

MelonHell:
resolves #2220


DeidaraMC:
I'd like to see the variable named (should)confirmTeleport rather than noConfirm, don't want to introduce double negatives into the api.
I also don't think this elegantly handles a situation where you teleport a player with confirmation and then immediately teleport a player without confirmation... it looks like it would override the confirmation it's waiting on and start showing the player's client position when it shouldn't.

I don't think noConfirm should not overwrite a valid pending teleport confirmation
```java
player.teleport(pos1, confirm); // Player is frozen at their destination
player.teleport(pos2, noConfirm); // Player is no longer frozen at their destination and running around doing whatever without being teleported, even though their client has yet to confirm teleport 1.
```

MelonHell:
> I don't think noConfirm should not overwrite a valid pending teleport confirmation
It does not overwrite valid pending teleport confirmation. It sends a teleport with ID -1, and when receiving a response with ID -1, this response is simply ignored

mworzala (approved):

PR #2224, opened 2024-06-27
Address some todos, create IntProvider for DimensionType

GreatWyrm:
* Changed Javadoc of LivingEntity#heal()
* Made STATUS_PERMISSION_LEVEL_OFFSET public to remove a magic number from a test
* Removed todo variant from WolfMeta since it has it
* Add IntProvider class for Dimension Type, adapted from https://minecraft.wiki/w/Dimension_type and Fabric's https://maven.fabricmc.net/docs/yarn-1.17+build.13/net/minecraft/util/math/intprovider/IntProvider.html


DeidaraMC:
I don't agree with the void threshold field, there's no game logic backing it and killing entities below that point (and that's a good thing, i don't think this should be default minestom behaviour)
To me it seems like sticking your own tag on instance for void height if necessary would be a better solution.

PR #2228, opened 2024-06-27
Automatically update an entity's attributes when changing equipment

GreatWyrm:


mworzala (commented):
Player overrides all of these and needs a way too still handle the slot updates

mworzala (approved):

Issue #2225, opened 2024-06-27
AttributeInstance addModifier unintuitive behavior


mworzala:
Currently when using `AttributeInstance#addModifier` the modifier will use a `putIfAbsent` call to add the modifier, meaning that it will not overwrite a previous value. To me this was unintuitive where I would have expected a new addModifier call to overwrite the previous value.

Just want to check I am not the only one who reads this as unintuitive behavior.

Steanky:
Also seemed a bit weird to me. If it's changed, I think  that `addModifier` will also (sometimes) have to re-compute the cached value (unless the replaced modifier is identical to the new one).

PR #2234, opened 2024-07-01
fix: AttributeInstance#addModifier not refreshing cachedValue for new modifier values

DeidaraMC:


TogAr2:
Doesn’t this do effectively nothing? Since putIfAbsent won’t change the maps contents if the key is already present, the cached value will be the same

mworzala:
Still not good. operation can also change which should invalidate it. also please add a test or two

mworzala:
Resolves #2225

PR #2235, opened 2024-07-01
Add Player#getCurrentItemUseTime

TogAr2:
This adds a method to get the current amount of ticks which have passed since the start of the item usage. Useful for determining the strength of a bowshot, etc.


DeidaraMC:
I'd prefer this be independent of instance, can you tie it to the player's tick instead?

mworzala (approved):

PR #2237, opened 2024-07-02
feat: Acquirable, ThreadDispatcher usage improvements

### LINKED ISSUE #2014: Some improvements to ThreadDispatcher API
The `ThreadDispatcher` class is a non-internal and potentially useful part of the API. It would allow users to implement their own systems that emulate Minestom's parallel entity ticking, and leverage `Acquirable` on their own objects to guarantee thread-safety. However, there are problems with how it is currently implemented:

* The class is hardcoded to only work with `Entity` (and subclasses). 
* Users cannot control the names of the threads. They will always look like "Ms-tick-n", which could cause potential confusion if they are used in other contexts.

To address these concerns, I've committed some improvements [to this branch on my fork](https://github.com/Minestom/Minestom/compare/master...Steanky:Minestom:thread-dispatcher-improvements?expand=1), which are summarized below. They are designed to be drop-in and should not break existing code.

* Instead of hardcoding `Entity`, `ThreadDispatcher` supports any implementation of the new functional interface `AcquirableSource<T>`, which specifies `Acquirable<? extends T> getAcquirable`
* `Entity` now implements `AcquirableSource<Entity>`
* `TickThread` has a new constructor which accepts a string for its name, rather than generating a hardcoded one from its integer ID
* A new static constructor method has been added to `ThreadDispatcher` that accepts an `IntFunction<String>` to generate unique names for each tick thread
* `ThreadDispatcher` has additional/clarified documentation in some places
* All non-demo subclasses of `Entity` override `getAcquirable` to return their own, more specific, type. This makes working with the acquirable API a bit easier, as users will not have to cast to more specific Entity subclasses when calling methods like `Acquirable#sync`

If these changes are desirable, I will make a pull request. Feedback is appreciated and I would be happy to make further modifications if necessary!

Additional note: on my branch, I opted to make the new `ThreadDispatcher` constructor accept an `IntFunction<String>` rather than `IntFunction<TickThread>`. Although the latter is strictly more flexible, I opted with the former to avoid exposing a class marked as `@ApiStatus.Internal`. However, I would argue that `TickThread` should be made part of the public API, considering that `ThreadDispatcher` is, and it would allow for more fine-grained control in combination with the `ThreadDispatcher` enhancements.

Steanky:
Closes #2014. 

Adds more convenient `getAcquirable` methods to Entity (and subclasses), more documentation, modifies ThreadDispatcher to be usable with any partition type, makes it no longer specific to just ticking entities, and includes a new static factory method for creating dispatchers with custom thread names. 

These shouldn't be breaking changes.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Entity.java
+++ b/src/main/java/net/minestom/server/entity/Entity.java
@@ -41,6 +41,7 @@ import net.minestom.server.snapshot.Snapshotable;
 import net.minestom.server.tag.TagHandler;
 import net.minestom.server.tag.Taggable;
 import net.minestom.server.thread.Acquirable;
+import net.minestom.server.thread.AcquirableSource;
 import net.minestom.server.timer.Schedulable;
 import net.minestom.server.timer.Scheduler;
 import net.minestom.server.timer.TaskSchedule;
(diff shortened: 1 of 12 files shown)


TheMode:
I'd prefer renaming the method to `acquirable()`. You can keep a separate `getAcquirable` in `Entity` for backward compatibility

mworzala (approved):

PR #2240, opened 2024-07-04
chore: add AlphaColor, convert Color to class, fix entity_effect particle

DeidaraMC:


src/main/java/net/minestom/server/particle/Particle.java
```diff
@@ -255,21 +256,26 @@ public void writeData(@NotNull NetworkBuffer writer) {
         }
     }
 
-    record EntityEffect(@NotNull NamespaceID namespace, int id, @NotNull RGBLike color) implements Particle {
+    record EntityEffect(@NotNull NamespaceID namespace, int id, @NotNull AlphaColor color) implements Particle {
 
         @Contract(pure = true)
-        public @NotNull EntityEffect withColor(@NotNull RGBLike color) {
+        public @NotNull EntityEffect withColor(@NotNull AlphaColor color) {
```
mworzala, line 262: Should have withColor(RGBLike) also which defaults to 1 alpha

src/main/java/net/minestom/server/color/Color.java
```diff
@@ -138,4 +146,42 @@ public int getGreen() {
     public int getBlue() {
```
mworzala, line 146: Should add withAlpha returning an alpha color

mworzala (changes requested):

PR #2243, opened 2024-07-04
fix: call PlayerDisconnectEvent on Configuration

### LINKED ISSUE #2242: PlayerDisconnectEvent not called if player in Configuration
image
image

Spliterash:
If the player was disconnected during configuration, the PlayerDisconnectEvent event will not be thrown. Even though PlayerConnection calls `player.scheduleNextTick(Entity::remove);`, it will never be executed because the player is not ticked off

resolves #2242

### DIFF
--- a/src/main/java/net/minestom/server/network/player/PlayerConnection.java
+++ b/src/main/java/net/minestom/server/network/player/PlayerConnection.java
@@ -4,6 +4,8 @@ import net.minestom.server.MinecraftServer;
 import net.minestom.server.crypto.PlayerPublicKey;
 import net.minestom.server.entity.Entity;
 import net.minestom.server.entity.Player;
+import net.minestom.server.event.EventDispatcher;
+import net.minestom.server.event.player.PlayerDisconnectEvent;
 import net.minestom.server.network.ConnectionState;
 import net.minestom.server.network.packet.server.SendablePacket;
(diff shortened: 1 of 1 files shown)


mworzala (commented):
I believe `isSpawned` is unnecessary. You can check the state of the player during the event to see if they are preplay when being disconnected.

MelonHell:
What else is missing in order to complete this pull request?

src/main/java/net/minestom/server/network/player/PlayerConnection.java
```diff
@@ -119,8 +121,12 @@ public void disconnect() {
         this.online = false;
         MinecraftServer.getConnectionManager().removePlayer(this);
         final Player player = getPlayer();
-        if (player != null && !player.isRemoved()) {
-            player.scheduleNextTick(Entity::remove);
+        if (player != null) {
+            if (connectionState == ConnectionState.PLAY && !player.isRemoved())
+                player.scheduleNextTick(Entity::remove);
+            else {
+                EventDispatcher.call(new PlayerDisconnectEvent(player));
```
mworzala, line 128: This should only be called if player is not removed, and should call remove itself. Currently if 3 different handlers of `AsyncPlayerConfigurationEvent` kicked the player, `PlayerDisconnectEvent` would be called 3 times.

Andrei965:
What else needs to be done for this to get merged?

PR #2253, opened 2024-07-09
Add BlockHandler::setPlacedBy to allow dealing with doors and beds

Minikloon:


Minikloon:
Following discussion, it has been deemed unpopular.

PR #2254, opened 2024-07-10
Add a few useful methods to Direction

Minikloon:


src/main/java/net/minestom/server/utils/Direction.java
```diff
@@ -15,11 +16,13 @@ public enum Direction {
     private final int normalX;
     private final int normalY;
     private final int normalZ;
+    private final Vec normalVec;
```
MelonHell, line 19: I think we should remove normalVec

mworzala: Why? it seems ok to me

MelonHell: It just seems weird to me to store the same data twice

mworzala: There are 6 of these ever instantiated. It's negligible and avoids a lot of allocation

PR #2255, opened 2024-07-10
Add Block::original so it's easier to compare by block type

Minikloon:
The implementation is not very sophisticated.


mworzala (changes requested):
I would prefer calling this `defaultState()`, original seems like a confusing name to me. Also, `Block.fromBlockId(id())` is a slightly preferable implementation.

I am fine with this because its obvious to someone new, although given blocks `a` and `b`, `a.id() == b.id()` does the same comparison better.

PR #2261, opened 2024-07-12
fix: fix vector normalization

BiConsumer:
Vec#normalize used to return NaN valued vectors when normalizing a 0 length vector and re-normalized already normalized vectors.


iam4722202468:
Inefficient. This is computing the length three times - twice in `normalize` and once in `isNormalized`.

Steanky:
Normalizing the zero-length vector is undefined, so I think having it return (NaN, NaN, NaN) makes sense.

PR #2264, opened 2024-07-12
feat: Inventory eventNodes

oglassdev:
Inventories now have event nodes. This should alleviate some of the annoyances with working with inventories.


kermandev:
Added in d5cb6111a1692be0b4a28af969657022b5ea2e74, after inventory click changes.

PR #2265, opened 2024-07-13
Fix biome sounds + particles

iam4722202468:
Biome sounds and particles don't work. This fixes sounds, and starts fixing particles. Not all particles are implemented, I was unable to figure out how some of them work.


mworzala (approved):
Does not fix everything, but is a good start

PR #2266, opened 2024-07-13
add a lot of badpackets checks

Tofaa2:
PR adds more compensated content for the client and a lot of basic badpackets stuff that the client server protocol should account for. Also removes the unnecesary repeat creation of the anticheat player object every loop

### DIFF
--- a/demo/src/main/java/net/minestom/demo/Main.java
+++ b/demo/src/main/java/net/minestom/demo/Main.java
@@ -1,6 +1,7 @@
 package net.minestom.demo;
 
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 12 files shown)


Tofaa2:
TODO Before moving on to reach and other checks, stuff that is left is compiling a set of entities per player, and its respective checks. Inventory and chunk data aswell

lolo45654:
Note that you can sneak while sending messages using toggle sneak in the accessibility settings. At least visually. I don't know about sprinting though.

mworzala:
Branch is being deleted. Was just an experiment not to be continued.

PR #2271, opened 2024-07-16
Increased support for 1.21 data driven enchantments

TogAr2:
This pull request will add support for more enchantment effects and other functions which have recently been added in the 1.21 update.


oliviathevampire:
What’s the progress on this?

TogAr2:
I originally planned to implement `DataPredicate`s as well, but they are already available at https://github.com/GoldenStack/trove, along with more loot table stuff.

I imagine we would eventually want Minestom to support these as well, but this PR is probably not the right place

TogAr2:
I don't know why the build failed but I don't think it is an issue with the PR

Issue #2249, opened 2024-07-06
feat: Events for block, chunk changes


Steanky:
I'm proposing several new events to be added that relate to blocks being modified by the server. These would function in addition to the player-related block events (`PlayerPlaceBlockEvent`, etc). 

- ~~`PreBlockChangeEvent`~~
  - ~~Broadcast from an `InstanceContainer`, before any chunk data is changed~~
  - ~~Includes the position, the old block, the new block, and the instance~~
  - ~~Cancellable, which would have the obvious effect of preventing the chunk from being modified~~
  - ~~Can be set to not inform the client of the block change~~
  - ~~Caveat: a `PlayerBlockPlaceEvent` that causes a `PreBlockChangeEvent` that is then cancelled will theoretically need to notify the client to prevent desync~~
  - **Problem:**
    - Since this event is broadcast from an InstanceContainer, it does not cover other places where blocks can be modified; e.g. through Chunk#setBlock, as is done by the Batch API.
  - **Proposed solution:**
    - Use an entirely separate event broadcast before a `BlockChangePacket` is sent. The event would be called in both `BlockPlacementListener` and `InstanceContainer`, supporting the goal of client-side blocks by allowing cancellation of the `BlockChangePacket`, as well as changing the block type. Then, use a different event that announces block changes from `Chunk#setBlock`, but does not support cancellation.
      - Not a breaking change, does not require substantial modification of the chunk API.
      - Would work for 3 of the 4 goals (excepting "selectively preventing block modifications" as the event is not cancellable).

- ~~`BlockChangeEvent`~~
  - ~~Broadcast after `PreBlockChangeEvent` and the chunk has been modified~~
  - ~~Otherwise includes all the same data as `PreBlockChangeEvent`~~
  - ~~Not cancellable~~
  - ~~Useful to run code once the user is sure the block has been changed (i.e. no `PreBlockChangeEvent` handler cancelled the modification)~~
  - **Likely unnecessary complexity, can probably be removed in favor of having just one event without losing much functionality.**

- ~~`SendChunkEvent`~~
  - ~~Broadcast before a chunk is sent to a player, in `Chunk#sendChunk`~~
  - ~~The actual chunk that is sent can be changed through the event~~
  - ~~Not cancellable~~
  - **Covered by #2274, which accomplishes the same goal**

These events could be useful for:
* Logging block modification
* Selectively preventing block modifications
* Implementing client-only blocks
[...]

PR #2274, opened 2024-07-19
feat: allow changing the chunk in PlayerChunkLoadEvent

Steanky:
Partially addresses #2249 (third bullet point); does not include any new events, only modifies PlayerChunkLoadEvent to support setting the chunk that is sent to the player. No breaking changes.


mworzala:
I am not a fan of allowing people to change the Chunk, it feels clunky. I am thinking it would be better to allow replacing the ChunkData and LightData objects in the event instead.

Steanky:
> I am not a fan of allowing people to change the Chunk, it feels clunky. I am thinking it would be better to allow replacing the ChunkData and LightData objects in the event instead.

Hmm, that doesn't seem like a great idea to me. There's currently no API to get `ChunkData` or `LightData` from a `Chunk`. Adding that would be a breaking change. 

`ChunkData`/`LightData` are also not very useful to expose to users through an event like this since they're basically just packet internals (with most of the interesting stuff represented as an opaque byte array). With `Chunk` in the event, you'd be able to query the properties of the chunk, and it doesn't incur a breaking change.

TheMeinerLP (changes requested):
Can you rebase it?

PR #2276, opened 2024-07-19
feat: proof-of-concept block events

### LINKED ISSUE #2249: feat: Events for block, chunk changes

Steanky:
There are several problems with this, hence why it's a draft. Will eventually close #2249, though!

Issues with packet caching should be fixed. Main concerns are now:
* API viability 
* Performance (but I doubt these events make a meaningful impact unless you have slow listeners); needs profiling
* Names (current ones are probably bad and/or confusing)

### DIFF
--- a/src/main/java/net/minestom/server/event/instance/BlockSetEvent.java
+++ b/src/main/java/net/minestom/server/event/instance/BlockSetEvent.java
@@ -1,0 +1,40 @@
+package net.minestom.server.event.instance;
+
+import net.minestom.server.coordinate.BlockVec;
+import net.minestom.server.coordinate.Point;
+import net.minestom.server.event.trait.BlockEvent;
+import net.minestom.server.event.trait.InstanceEvent;
+import net.minestom.server.instance.Instance;
+import net.minestom.server.instance.block.Block;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Called before a block is set on a chunk. {@link BlockSetEvent#getBlock()} to see what the new block is. The new block
+ * may or may not differ from the old block.
+ */
+public class BlockSetEvent implements InstanceEvent, BlockEvent {
+    private final Instance instance;
+    private final Block block;
+    private final BlockVec blockPosition;
+
+    public BlockSetEvent(@NotNull Instance instance, @NotNull Block block, @NotNull Point blockPosition) {
+        this.instance = instance;
+        this.block = block;
+        this.blockPosition = new BlockVec(blockPosition);
+    }
+
+    @Override
+    public @NotNull Block getBlock() {
+        return block;
+    }
+
+    @Override
+    public @NotNull BlockVec getBlockPosition() {
+        return blockPosition;
+    }
+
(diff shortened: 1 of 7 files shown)


UserUNP:
What could be the purpose of PlayerBlockUpdateEvent?

Steanky:
> What could be the purpose of PlayerBlockUpdateEvent?

Controlling `BlockUpdatePacket`, for when you want certain blocks on clients to differ from the server, or different clients to see different blocks. Naming is WIP at this point, could probably be changed to something more descriptive.

Steanky:
Likely not going to be added in its present form, and the same effect should (in theory) be possible with subclassing DynamicChunk.

PR #2281, opened 2024-07-20
Minestom 1.next

mworzala:
This branch and PR is for the pending changes for the next Minecraft update.

Breaking changes should be directed to this branch, but only if they must. Other PRs should always go to master.


mworzala:
Was for 1.21.3 so closing, may bring back for future updates :)

PR #2291, opened 2024-07-23
Remove BinaryReader/Writer

TheMode:


src/main/java/net/minestom/server/command/builder/arguments/number/ArgumentInteger.java
```diff
@@ -1,11 +1,12 @@
...
 
-import net.minestom.server.utils.binary.BinaryWriter;
+import net.minestom.server.network.NetworkBuffer;
 
 public class ArgumentInteger extends ArgumentNumber<Integer> {
 
     public ArgumentInteger(String id) {
-        super(id, "brigadier:integer", Integer::parseInt, Integer::parseInt, BinaryWriter::writeInt, Integer::compare);
+        super(id, "brigadier:integer", Integer::parseInt, Integer::parseInt,
+                (buffer, number) -> buffer.write(NetworkBuffer.INT, number), Integer::compare);
```
mworzala, line 9: could just take the networkbuffer.type here instead of a full function right?

PR #2293, opened 2024-07-23
feat: Allow disabling "block does not have any corresponding handler" warning

Steanky:
When using `AnvilLoader`, some blocks need a handler, which if missing will generate a warning in the console. However, this warning is not always useful, as in some cases it is not necessary (or desirable) to implement one. This PR allows users to disable this warning by setting the server flag `minestom.missing-block-handler-warning` to `false` (defaults to `true`).


mworzala:
I am pretty opposed to this change. If we open the "config option for specific logs" can of worms, there will be a lot of config options. You can already disable this log using the logger settings of your slf4j impl, and should just do that.

PR #2294, opened 2024-07-23
Remove cursor items from individual inventories

GoldenStack:
Each inventory, as well as the player inventory, having its own inventory was rather complicated. Now, only player inventories have a cursor item.

Breaking changes:
- Removed a few functions from `Inventory`
- Clearing an `Inventory` no longer clears the cursor item (you must clear the player inventory for that)

All tests pass, clicking in the demo server works fine, and tests were added/modified to verify this behaviour.

### DIFF
--- a/src/main/java/net/minestom/server/inventory/Inventory.java
+++ b/src/main/java/net/minestom/server/inventory/Inventory.java
@@ -16,7 +16,6 @@ import org.jetbrains.annotations.NotNull;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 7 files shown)


src/main/java/net/minestom/server/inventory/Inventory.java
```diff
@@ -156,48 +153,18 @@ public boolean addViewer(@NotNull Player player) {
...
         return result;
     }
 
-    /**
-     * Gets the cursor item of a viewer.
-     *
-     * @param player the player to get the cursor item from
-     * @return the player cursor item, air item if the player is not a viewer
-     */
-    public @NotNull ItemStack getCursorItem(@NotNull Player player) {
```
mworzala, line 170: Both of these should come back, be deprecated, and proxy to the player inventory

src/main/java/net/minestom/server/inventory/Inventory.java
```diff
@@ -105,7 +103,6 @@ public byte getWindowId() {
 
     @Override
     public synchronized void clear() {
-        this.cursorPlayersItem.clear();
         super.clear();
```
mworzala, line 106: not a necessary override anymore if empty

mworzala (commented):
I think I can live with the clear change going into master, but the cursor functions need to stay deprecated.

PR #2296, opened 2024-07-24
feat: add ability to clear modifiers from an AttributeInstance

Steanky:
Also, add a method to get all `AttributeInstance`s on a `LivingEntity`. Useful if you want to inspect the player's current attribute modifiers, or remove all of them without needing to know every kind of `Attribute` they might have.

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -38,6 +38,7 @@ import net.minestom.server.utils.time.TimeUnit;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.UnmodifiableView;
 
 import java.time.Duration;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -495,6 +492,15 @@ public void heal() {
                 s -> new AttributeInstance(attribute, this::onAttributeChanged));
     }
 
+    /**
+     * Retrieves all {@link AttributeInstance}s on this entity.
+     *
+     * @return a collection of all attribute instances on this entity
+     */
+    public @NotNull Collection<AttributeInstance> getAttributeInstances() {
```
mworzala, line 500: Should just be getAttributes() in line with getAttribute

src/main/java/net/minestom/server/entity/attribute/AttributeInstance.java
```diff
@@ -114,6 +116,14 @@ public AttributeModifier removeModifier(@NotNull AttributeModifier modifier) {
         return removeModifier(modifier.id());
     }
 
+    /**
+     * Clears all modifiers on this instance.
+     */
+    public void clearModifiers() {
+        this.modifiers.clear();
```
mworzala, line 123: This should not clear `LivingEntity.SPRINTING_SPEED_MODIFIER`, that is almost always invalid behavior.

mworzala (changes requested):

mworzala (approved):

PR #2301, opened 2024-07-24
Remove NamespaceID class

MelonHell:
The NamespaceID class is no longer needed and is just a nuisance. I think it should be replaced with Key


Steanky:
I strongly agree with this PR and I'd like to add additional justification for its removal: the equals/hashCode implementation of NamespaceID must be *identical* to that of Adventure's KeyImpl, or otherwise incorrect behavior will result when storing NamespaceID and KeyImpl in the same set or map (ex. in a Set<Key>). 

This would not be a problem if Adventure specified a contract for `Key` implementations, but I do not believe that they currently do. Therefore, their equals/hashCode implementation could change at any time.

This may sound like a niche edge case, and it is, but I actually ran into this bug in my own codebase a while ago (fixed by this old pull request https://github.com/Minestom/Minestom/issues/1883). Removing NamespaceID would prevent this issue reappearing in the future.

PR #2302, opened 2024-07-24
Remove permissions

MelonHell:
### Why I want to remove permissions:
1) The permissions system is not part of Minecraft. We don't add built-in PlaceholderAPI or WorldGuard to Minestom, so why did we add permissions?

2) The built-in permissions system is too raw, it doesn't contain tristate and doesn't allow you to conveniently connect an external provider like LuckPerms

3) Checking permissions is not the responsibility of the Player class! It would be correct to do `PermissionSystem.hasPermission(player, "test")` instead of `Player.hasPermission("test")`

4) "If you don't like this system so much, just ignore it" - If this code exists, there is always a chance that some library will access it, for example an external Command Framework will ask the system, which is not even used, which can create problems

😁😁😁
image


cody-quinn (approved):
LGTM! :+1:

mudkipdev (approved):

PR #2321, opened 2024-08-04
Improve collision performance

xZeGoe:


iam4722202468:
Have you done performance testing to see if this actually improves performance? I'm not convinced this would be faster, it might even be slower because more branching is used.

In practice, most checks will have an x, y and z component. This new code would require 9 branches for the most common collision check, while the current method requires 1.

Your code has a max stack size of 3 due to nesting, while the current code has a max stack size of 1. This means more overhead.

I'm unsure why you removed final from the points.

xZeGoe:
Thanks for the feedback. I've did performance testing, and it shows improvements of 5-90% (on average 40%) in CPU usage and 550% in RAM usage. If the changes are indeed slower, I'd be glad to close this PR.

iam4722202468:
After checking performance, this does actually appear to be a significant performance improvement. Not sure how, but it's definitely better. Good work, I'll merge this

PR #2322, opened 2024-08-05
Revert "Improve collision performance"

AidanMars2:
Reverts Minestom/Minestom#2321. While the PR did improve performance significantly, it did so by only checking collisions when an entity crossed a block border. This can result in entities intersecting blocks.


TheMode:
the `pointAfter` improvement is probably worth keeping. Thats still one less allocation

iam4722202468:
I'm surprised no tests caught that... Going to merge without the pointAfter improvement because working physics is critical.

iam4722202468:
Can you write a test case which fails on the improvement PR but works here?

PR #2325, opened 2024-08-07
Add EquipmentSlot.BODY

TogAr2:
Vanilla recently added `EquipmentSlot.BODY`, and since the equipment packet works by using enum ordinals, it has to be present in the Minestom enum as well.

This new equipment slot is in vanilla only used by horses, wolfs and llama's, but it is still generalized and every living entity except for players can have equipment in this slot.


src/main/java/net/minestom/server/entity/EquipmentSlot.java
```diff
@@ -25,7 +26,7 @@ public enum EquipmentSlot {
     }
 
     public boolean isHand() {
-        return !armor;
+        return !armor && this != BODY;
```
mworzala, line 29: this should just be == MAIN_HAND || == OFF_HAND

mworzala (changes requested):

mworzala (approved):

PR #2327, opened 2024-08-10
Add chunk loading as an option to Batches

Flopgop:
yeah two-line change, ikr
I'm PRing this because block batches will just kill themselves if they can't place blocks because the chunk is loaded. This is really stupid behavior and I think it should be optional. Hence this PR.

Can someone review ChunkBatch.java @ line 106 and see if my implementation is sufficient? I can't get a straight answer on whether or not batches are supposed to block so I don't know if blocking while loading the chunk is a good idea.

### DIFF
--- a/src/main/java/net/minestom/server/instance/batch/BatchOption.java
+++ b/src/main/java/net/minestom/server/instance/batch/BatchOption.java
@@ -13,6 +13,7 @@ public class BatchOption {
     private boolean calculateInverse = false;
     private boolean unsafeApply = false;
     private boolean sendUpdate = true;
+    private boolean loadChunks = false;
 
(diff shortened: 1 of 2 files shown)


Flopgop:
This will stay drafted until I get feedback on whether my ChunkBatch implementation of chunk loading is deemed sufficient.

mworzala:
I am probably opposed to this -- We don't have a good way for these chunks to be unloaded currently so if people want to load chunks to place blocks I would prefer they do it themselves (acknowledging that they would not unload if done like this pr).

About the actual implementation: the join on batch thread doesn't make sense. It should do a `thenRunAsync` passing the batch pool as executor.

PR #2329, opened 2024-08-13
Improve chat API

mudkipdev:
Having the second layer of setting the message format function which takes in a chat event is confusing and unnecessary. This pull request removes that. I also renamed `message` to `content` which should make the usage clearer. (breaking)

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerChatEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerChatEvent.java
@@ -87,7 +87,6 @@ public class PlayerChatEvent implements PlayerInstanceEvent, CancellableEvent {
                 .arguments(
                         Component.text(player.getUsername())
                                 .insertion(player.getUsername())
-                                .clickEvent(ClickEvent.suggestCommand("/msg " + player.getUsername() + " "))
                                 .hoverEvent(player),
                         Component.text(content));
     }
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/listener/ChatMessageListener.java
```diff
@@ -40,30 +34,20 @@ public static void chatMessageListener(ClientChatMessagePacket packet, Player pl
...
             }
         });
     }
-
-    private static @NotNull Component buildDefaultChatMessage(@NotNull Player player, @NotNull String message) {
-        final String username = player.getUsername();
-        return Component.translatable("chat.type.text")
-                .args(Component.text(username)
-                                .insertion(username)
-                                .clickEvent(ClickEvent.suggestCommand("/msg " + username + " "))
```
OneLiteFeather, line 63: I don't really understand why you remove the clickEvent line here. Can you tell me the reason?

mudkipdev: The command is not implemented in minestom and it might not be the desired behavior

src/main/java/net/minestom/server/event/player/PlayerChatEvent.java
```diff
@@ -1,85 +1,70 @@
...
+ * The event can be cancelled to not send anything, and the final message can be changed.
  */
 public class PlayerChatEvent implements PlayerInstanceEvent, CancellableEvent {
-
     private final Player player;
     private final Collection<Player> recipients;
-    private String message;
-    private Function<PlayerChatEvent, Component> chatFormat;
-
+    private final String content;
```
mworzala, line 20: `rawMessage` and `formattedMessage`.

PR #2332, opened 2024-08-15
Integrate PlayerInventory

GoldenStack:
Integrates PlayerInventory and Inventory, making them much more similar than they are now by moving a lot of existing logic to AbstractInventory.

\[this is a portion of my old inventory PR\]

Primary changes:
- Inventory events (and all other relevant places that were using it) now use `@NotNull AbstractInventory` instead of `@Nullable Inventory` (which used `null` to indicate the player inventory)
- `PlayerInventoryItemChangeEvent` no longer exists; use `InventoryItemChangeEvent`
- `AbstractInventory` now implements `Viewable`. This means `PlayerInventory` supports (and has to deal with) multiple potential viewers, too.
- `PlayerInventory` can no longer guarantee that it knows the held item because each viewer may hold a different item.

Note:
`PlayerInventory#UNSAFE_itemInsert` is pretty ugly (for example it must iterate through its viewers two separate times). Having PlayerInventory store the held slot instead of the player would simplify it a bit, but I don't really like the sound of that.


GoldenStack:
(marking as a draft because it definitely needs discussion + I am still not super happy with some of the implementation)

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -754,6 +756,7 @@ private void spawnPlayer(@NotNull Instance instance, @NotNull Pos spawnPosition,
         }
 
         if (dimensionChange || firstSpawn) {
+            inventory.addViewer(this);
```
mworzala, line 759: why re add on dimension change

src/main/java/net/minestom/server/utils/mojang/MojangUtils.java
```diff
@@ -1,32 +1,25 @@
...
 import java.io.IOException;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Utils class using mojang API.
  */
 public final class MojangUtils {
     private static final String FROM_UUID_URL = "https://sessionserver.mojang.com/session/minecraft/profile/%s?unsigned=false";
     private static final String FROM_USERNAME_URL = "https://api.mojang.com/users/profiles/minecraft/%s";
```
mworzala, line 18: why is this change here? it shouldn't be here

mworzala (commented):
looks good besides the two minor things

src/main/java/net/minestom/server/utils/mojang/MojangUtils.java
GoldenStack: artifact of my obvious skill issue with regards to merging. i'll fix that

src/main/java/net/minestom/server/entity/Player.java
GoldenStack: good catch, was in the wrong place

GoldenStack:
Would it make sense to rename `AbstractInventory` -> `Inventory` and `Inventory` -> `ContainerInventory` (or something like that)?
Now that AbstractInventory is much less abstract and in fact contains most of the logic

mworzala (approved):

Issue #289, opened 2021-05-15
Instabreaking blocks doesn't run BlockBreakEvent


emortaldev:
https://cdn.discordapp.com/attachments/809040362835935242/838618645898067969/2021-05-03_04-29-19-1.mp4

Once you re-log only the ones you didn't instabreak actually appear as broken

Uraxys:
Took a look at it and checked how much it would be to get it to work, from my adventure into the code it looks like we need to check each block mined, check if it requires a special tool, if it does, then also get the speed bonus we get for using the right tool with the given block (For example, shears, you get a 15.0 speed bonus against leaves, but only 5.0 against wool).
Then calculate the efficiency enchantment + dig speed + bonus dig speed, with this we check if the mining speed is above the instant mine "level", if it is we instantly break the block & fire the event.

I started on it before I knew the things above and found out that it wasn't as easy as I first thought
Not sure what is the best solution, but I think adding a "require tool" value to the Block enum or something like that is the best way, if not then we need to add each block that requires a special tool to a set or list, but I could be wrong.

emortaldev:
Well I'd love to check if the block broken was instabroken, but how am I supposed to know if any blocks are broken if  blockbreakevent isn't triggered

TheMode:
> Well I'd love to check if the block broken was instabroken, but how am I supposed to know if any blocks are broken if blockbreakevent isn't triggered

The issue is that the client doesn't tell us when a block has been instabroken, and we currently do not have the math to check it ourselves.

mworzala:
This issue has actually regressed a bit, you now cannot even instabreak blocks.

DasBabyPixel:
I need this for a project, so I wanted to ask if anyone has started/tried some sort of fix for me to go off of?

mworzala:
Fixed in above

PR #2340, opened 2024-08-19
Fix Instamine

DasBabyPixel:
WIP to fix instamine issues.
At this point this contains the breaking speed calculation.
At this time uses a custom version of the data generator, which is published on [my repo](https://nexus.darkcube.eu/repository/dasbabypixel/).
These changes are gonna have to be dropped when this is ready to merge, and the official data generator will have to update to support everything required by these changes.

Please check the calculation, feedback and suggestions are appreciated!

**Some relevant information for testing:**
Added command in demo: /attribute and /testinstabreak
/testinstabreak without any arguments gives some nice default configuration

Required player.mining_efficiency attribute
wool with shears: 19
stone with diamond_pickaxe: 37

Required player.mining_efficiency attribute when not on ground
wool with shears: 115
stone with diamond_pickaxe: 217

Required player.mining_efficiency attribute when not on ground and in water
wool with shears: 595

Required player.block_break_speed attribute
wool: 24
wool with shears: 4.8
stone: 150
stone with diamond_pickaxe: 5.625

Required player.block_break_speed attribute when not on ground
wool: 120
wool with shears: 24
stone: 750
stone with diamond_pickaxe: 28.125

Required player.block_break_speed attribute when not on ground and in water
wool: 600
wool with shears: 120
stone with diamond_pickaxe: 140.625


DasBabyPixel:
Should fix #289 when done

DasBabyPixel:
Currently, minestom assumes the instant break ability allows a player to instantly break all blocks. The vanilla client doesn't though. The entire instantbreak/instabuild ability is very unclear, not affecting block breaking, but changing environment sounds, UI features (removes anvil "Too Expensive!" and changes Enchanting Table to "creative mode"), and infinite material checks.
But unless there is a reason not to reflect the client behaviour, I'd change the calculation to only assume instantbreak in case of creative.

mworzala:
MinestomDataGenerator PR is merged & published as `1.21-rv4`.

demo/src/main/java/net/minestom/demo/PlayerInit.java
```diff
@@ -117,15 +117,15 @@ public class PlayerInit {
...
-                class A {
-                    static boolean b = false;
-                }
-                if (A.b) {
-                    event.getPlayer().getAttribute(Attribute.GENERIC_MOVEMENT_SPEED).removeModifier(NamespaceID.from("test"));
-                } else {
-                    event.getPlayer().getAttribute(Attribute.GENERIC_MOVEMENT_SPEED).addModifier(new AttributeModifier(NamespaceID.from("test"), 0.5, AttributeOperation.ADD_VALUE));
-                }
-                A.b = !A.b;
+//                class A {
```
mworzala, line 120: can just be deleted.

DasBabyPixel:
waiting for build to finish, then dropping commit

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
```diff
@@ -69,31 +70,30 @@ public static void playerDiggingListener(ClientPlayerDiggingPacket packet, Playe
 
     private static DiggingResult startDigging(Player player, Instance instance, Point blockPosition, BlockFace blockFace) {
         final Block block = instance.getBlock(blockPosition);
-        final GameMode gameMode = player.getGameMode();
+
+        System.out.println(" Packet Start Digging");
```
mworzala, line 74: remove

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
```diff
@@ -69,31 +70,30 @@ public static void playerDiggingListener(ClientPlayerDiggingPacket packet, Playe
...
         }
-        // Client only send a single STARTED_DIGGING when insta-break is enabled
+        // Client only sends a single STARTED_DIGGING when insta-break is enabled
         return breakBlock(instance, player, blockPosition, block, blockFace);
     }
 
     private static DiggingResult cancelDigging(Player player, Instance instance, Point blockPosition) {
         final Block block = instance.getBlock(blockPosition);
+
+        System.out.println(" Packet Cancel Digging");
```
mworzala, line 95: remove

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
```diff
@@ -102,10 +102,23 @@ private static DiggingResult cancelDigging(Player player, Instance instance, Poi
     private static DiggingResult finishDigging(Player player, Instance instance, Point blockPosition, BlockFace blockFace) {
         final Block block = instance.getBlock(blockPosition);
 
+        System.out.println(" Packet Finish Digging");
```
mworzala, line 105: remove

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
```diff
@@ -102,10 +102,23 @@ private static DiggingResult cancelDigging(Player player, Instance instance, Poi
...
 
+        final int breakTicks = BlockBreakCalculation.breakTicks(block, player);
+        // Realistically shouldn't happen, but a hacked client can send any packet, also illegal ones
+        // If the block is unbreakable, prevent a hacked client from breaking it!
+        if (breakTicks == BlockBreakCalculation.UNBREAKABLE) {
+            PlayerCancelDiggingEvent playerCancelDiggingEvent = new PlayerCancelDiggingEvent(player, block, new BlockVec(blockPosition));
+            EventDispatcher.call(playerCancelDiggingEvent);
+            return new DiggingResult(block, false);
+        }
+        // TODO maybe add a check if the player has spent enough time mining the block.
```
mworzala, line 119: I don't believe its Minestoms job to check this.

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
```diff
@@ -174,6 +187,8 @@ private static DiggingResult breakBlock(Instance instance,
                                             Player player,
                                             Point blockPosition, Block previousBlock, BlockFace blockFace) {
         // Unverified block break, client is fully responsible
+        System.out.println("  Break block");
```
mworzala, line 190: remove

src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
```diff
@@ -0,0 +1,192 @@
...
+     * Calculates the block break time in ticks
+     *
+     * @return the block break time in ticks, -1 if the block is unbreakable
+     */
+    public static int breakTicks(@NotNull Block block, @NotNull Player player) {
+        if (player.getGameMode() == GameMode.CREATIVE) {
+            // Creative can always break blocks instantly
+            return 0;
+        }
+        // Taken from minecraft wiki Breaking#Calculation
```
mworzala, line 43: Add link

src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
```diff
@@ -0,0 +1,192 @@
...
+        Tool tool = item.get(ItemComponent.TOOL);
+        boolean isBestTool = canBreakBlock(tool, block);
+        float speedMultiplier;
+
+        if (isBestTool) {
+            speedMultiplier = getMiningSpeed(tool, block);
+
+            // Also, the wiki's canBreak seems to be this >1
+            if (speedMultiplier > 1F) {
+                // Wiki tells us about calculations for efficiency, but enchantments don't apply these calculations in minestom.
```
mworzala, line 67: Since data driven enchantments, efficiency uses the mining efficiency attribute anyway so this is correct (though minestom doesn't apply currently)

src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
```diff
@@ -0,0 +1,192 @@
...
+
+        double damage = speedMultiplier / blockHardness;
+
+
+        if (isBestTool) {
+            damage /= 30;
+        } else {
+            damage /= 100;
+        }
+        System.out.println(block.key().asString() + " Hardness: " + blockHardness);
```
mworzala, line 104: remove x 3

src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
```diff
@@ -0,0 +1,192 @@
...
+
+        if (player.hasEffect(PotionEffect.MINING_FATIGUE)) {
+            speedMultiplier *= getMiningFatigueMultiplier(player);
+        }
+
+        speedMultiplier *= (float) player.getAttributeValue(Attribute.PLAYER_BLOCK_BREAK_SPEED);
+
+        ItemStack helmet = player.getInventory().getHelmet();
+
+        if (isInWater(player) && !helmet.get(ItemComponent.ENCHANTMENTS, EnchantmentList.EMPTY).has(Enchantment.AQUA_AFFINITY)) {
```
mworzala, line 88: Do not check aqua affinity. The definition for the aqua affinity enchantment applies the submerged mining speed attribute.

```
    "minecraft:aqua_affinity": {
        effects: {
            "minecraft:attributes": [
                {
                    amount: {
                        base: 4b,
                        per_level_above_first: 4b,
                        type: "minecraft:linear"
                    },
                    attribute: "minecraft:player.submerged_mining_speed",
                    id: "minecraft:enchantment.aqua_affinity",
                    operation: "add_multiplied_total"
                }
            ]
        },
```

src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
```diff
@@ -0,0 +1,192 @@
...
+        // Taken from minecraft wiki Breaking#Calculation
+        // More information to mimic calculations taken from minecraft's source
+        Registry.BlockEntry registry = block.registry();
+        double blockHardness = registry.hardness();
+        if (blockHardness == -1) {
+            // Bedrock, barrier, and unbreakable blocks
+            return UNBREAKABLE;
+        }
+        ItemStack item = player.getItemInMainHand();
+        // Bamboo is hard-coded in client
```
mworzala, line 52: classic stuff, really.

DasBabyPixel: ? anything I should change?

src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
DasBabyPixel: Should I change comment? Tried to explain deviation from wiki, in case someone looks at it in the future

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
DasBabyPixel: Actually, I believe it is. Basic illegal behaviour should be checked, that includes a sent FINISHED_DIGGING packet on an unbreakable block. Otherwise players can go around breaking any block (bedrock, barrier) they like if the break isn't cancelled by the server owner.
Similar sanity checks are already there, IMO not harmful adding another:
https://github.com/Minestom/Minestom/blob/480a7dcaa2b26a074c97583f8177d2f7a3677876/src/main/java/net/minestom/server/listener/PlayerDiggingListener.java#L115

DasBabyPixel:
@mworzala ?

src/main/java/net/minestom/server/listener/PlayerDiggingListener.java
DasBabyPixel: vanilla has a similar time check, which is why blocks pop up again after broken if the server is lagging or you have a bad internet connection. I haven't added the check because it is not trivial to do it right, I just left the comment because it is something that could be considered in the future.

mworzala (approved):
I know this took a long time, my apologies for that. Thank you for making the change, it has been too long with this issue.

PR #2344, opened 2024-08-21
Add inversion methods to Pos

JXSnack:
I've recently been working a lot with inverted pitch and yaw values, and it's really annoying to go `withYaw(-entity.getYaw())` so I decided to make a fix myself


JXSnack:
I'm dumb, this doesn't correctly invert it like one would expect.

PR #2350, opened 2024-08-24
give more control of collisions to the entities colliding

DasBabyPixel:
I noticed that I couldn't place a block where a snowball entity is, so I investigated and found that this part of the logic is hardcoded and impossible to override without going to extreme lengths.
This PR gives entities more control over whether they should prevent block placement and whether should be used as collision targets for other entities (say a projectile can't be hit by another projectile)


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1718,8 +1722,29 @@ public boolean intersectBoxSwept(@NotNull Point rayStart, @NotNull Point rayDire
...
+
+    public boolean isPreventBlockPlacement() {
+        // EntityMeta can change at any time, so initializing this during #initCollisions is not an option
+        // Can be overridden to allow for custom behaviour
+        if (entityMeta instanceof ArmorStandMeta armorStandMeta && armorStandMeta.isMarker()) return false;
+        return preventBlockPlacement;
+    }
+
+    protected void initCollisions() {
+        preventBlockPlacement = entityType != EntityType.ARROW
```
TheMode, line 1737: You should probably make constants `Set<EntityType>` instead of chained expressions

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -504,6 +510,7 @@ public synchronized void switchEntityType(@NotNull EntityType entityType) {
         EntitySpawnType type = entityType.registry().spawnType();
         this.aerodynamics = aerodynamics.withAirResistance(type == EntitySpawnType.LIVING ||
                 type == EntitySpawnType.PLAYER ? 0.91 : 0.98, 1 - entityType.registry().drag());
+        initCollisions();
```
iam4722202468, line 513: name this `updateCollisions` since you are calling it multiple times

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -87,7 +88,11 @@ public class Entity implements Viewable, Tickable, Schedulable, Snapshotable, Ev
             EntityType.FISHING_BOBBER, EntityType.SNOWBALL, EntityType.EGG, EntityType.ENDER_PEARL, EntityType.POTION,
             EntityType.EYE_OF_ENDER, EntityType.DRAGON_FIREBALL, EntityType.FIREBALL, EntityType.SMALL_FIREBALL,
             EntityType.TNT);
-
+    private static final Set<EntityType> ALLOW_BLOCK_PLACEMENT_ENTITIES = Set.of(EntityType.ARROW, EntityType.ITEM,
```
iam4722202468, line 91: Do these come from the vanilla client?

src/main/java/net/minestom/server/collision/EntityCollision.java
```diff
@@ -26,7 +26,7 @@ final class EntityCollision {
             SweepResult sweepResult = new SweepResult(minimumRes, 0, 0, 0, null, 0, 0, 0, 0, 0, 0);
 
             if (!entityFilter.apply(e)) continue;
-            if (!e.hasCollision()) continue;
+            if (!e.isCollidesWithEntities()) continue;
```
iam4722202468, line 29: `hasEntityCollisions` would be a better name

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -1718,8 +1726,20 @@ public boolean intersectBoxSwept(@NotNull Point rayStart, @NotNull Point rayDire
         return boundingBox.relativeEnd();
     }
 
-    public boolean hasCollision() {
-        return hasCollision;
+    public boolean isCollidesWithEntities() {
+        return collidesWithEntities;
+    }
+
+    public boolean isPreventBlockPlacement() {
```
iam4722202468, line 1733: `preventBlockPlacement` is better

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -766,6 +773,7 @@ public CompletableFuture<Void> setInstance(@NotNull Instance instance, @NotNull
         this.lastSyncedPosition = spawnPosition;
         this.previousPhysicsResult = null;
         this.instance = instance;
+        initCollisions();
```
iam4722202468, line 776: Could this be called in the constructor instead?

src/main/java/net/minestom/server/entity/Entity.java
DasBabyPixel: No they don't, I just added the ones that I thought of. Probably missing some

src/main/java/net/minestom/server/entity/Entity.java
DasBabyPixel: done

src/main/java/net/minestom/server/collision/EntityCollision.java
DasBabyPixel: done

src/main/java/net/minestom/server/entity/Entity.java
DasBabyPixel: done

src/main/java/net/minestom/server/entity/Entity.java
iam4722202468: The only way to set `entityType` is through the constructor and `switchEntityType`, and the field is marked with `UNSAFE to change, modify at your own risk`.

It's safe to call `updateCollisions` in the constructor after the field `entityType` is set and updating collisions in the constructor makes a lot more sense than in `setInstance`

PR #2357, opened 2024-08-26
Fix server freeze at extreme positions

CoPokBl:
This pull request fixes a critical exploit in the library which allows clients to crash servers by setting their positions to above the 32bit integer limit, which is possible via a cheat client.

You don't have to keep this, but at least have it until a solution can be found to the problem of integer limits as a whole.


GreatWyrm:
I'd recommend you also add bounds to the other coordinates, not just the Y-coordinate.

TheMode:
Should be fixed by 789befee315a8d10ce104a0919c731a7501517fc

PR #2360, opened 2024-08-28
Implement Graph Cache in CommandManager

KermanIsPretty:
Implemented the graph cache TODO in CommandManager#getCache

I changed the implementation sightly to make use of CommandManager#getCommands to ensure that any Graph implementation is not mutating commands.

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandManager.java
+++ b/src/main/java/net/minestom/server/command/CommandManager.java
@@ -32,6 +32,7 @@ public final class CommandManager {
     private final Set<Command> commands = new HashSet<>();
 
     private CommandCallback unknownCommandCallback;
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/command/CommandManager.java
```diff
@@ -188,9 +193,18 @@ public CommandParser.Result parseCommand(@NotNull CommandSender sender, String i
         return parser.parse(sender, getGraph(), input);
     }
 
-    private Graph getGraph() {
-        //todo cache
-        return Graph.merge(commands);
+    private @NotNull Graph getGraph() {
```
TheMode, line 196: this is not thread-safe and can still return null

PR #2364, opened 2024-08-30
fix: chunk history desyncing from actual state of loaded chunks

qxeii:
This PR fixes an issue with the player auto chunk loader where chunks would be marked as loaded but never marked as unloaded.

**This bug can be reproduced by**:
* Enabling automatic chunk loading
* Adding code to unload chunks within the events provided by the Player class

**Expected behaviour** (should also be provided by this patch):
* A chunk that has been unloaded due to exiting a player's view distance is able to be reloaded when the player goes near it again.

**Actual behaviour**:
* Loading, unloading, and then attempting to load again any given chunk with a player will fail when trying to reload the chunk after an unload.

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -167,6 +167,7 @@ public class Player extends LivingEntity implements CommandSender, Localizable, 
     };
     final IntegerBiConsumer chunkRemover = (chunkX, chunkZ) -> {
         // Unload old chunks
+        this.chunkUpdateLimitChecker.removeFromHistory(ChunkUtils.getChunkIndex(chunkX, chunkZ));
         sendPacket(new UnloadChunkPacket(chunkX, chunkZ));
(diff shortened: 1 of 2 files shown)


qxeii:
Upon further inspection it seems like the system causing this issue is entirely unnecessary (at least in my opinion) so I've removed all of the code related to it.

If anyone still believes there is a use for delaying chunk reloading after unloading chunks, it should be replaced with a better implementation as the current one would have worked strangely with low render distances (and doesn't work as intended currently).

mworzala (approved):
I'm good with this, we discussed it a bit ago and didnt know why it's present now. It needs a rebase tho

TheMeinerLP:
@qxeii Can you rebase your PR ?

qxeii:
@TheMeinerLP just saw this, sorry for the late reply. i've not used minestom in a while but i am not entirely certain this PR actually fixes the issue i thought it did. i would strongly suggest looking into this code again and producing a new fix (and closing this PR)

PR #2377, opened 2024-09-04
Tick alive threads only

xZeGoe:
Resolves https://github.com/Minestom/Minestom/issues/2096

I assume it's not best solution, but it should be simpler than https://github.com/Minestom/Minestom/pull/2097


mworzala:
Fixed in #2445

PR #2382, opened 2024-09-06
fix: saturation not working in food item component

SLH335:
I am using the food implementation of MinestomPvP and noticed that eating food only applied nutrition, not saturation, to the player.

While searching for the cause of the issue, I found a typo in the Food ItemComponent, where the `saturation_modifier` tag was misspelled as `saturation_odifier`. Correcting the typo did not fix the issue, however.
Upon further inspection, I noticed that the tag is actually labeled as just `saturation` in the item data:

screenshot

I renamed the tag to `saturation` in the Food ItemComponent and now saturation gets correctly applied to the player, so this seems to resolve the issue.


mworzala (approved):
Yep, not sure where saturation_modifier came from let alone saturation_odifier. thanks

PR #2384, opened 2024-09-08
Rework LoginListener

KermanIsPretty:
Currently the old Mojang login listener uses HTTPClient (Only use in the entire project, and doesn't even use it correctly)

This commit also moves the implementation of Player#kick to PlayerConnection#kick, I would like to know the feedback on this change or if this should be a separate PR.

### DIFF
--- a/src/main/java/net/minestom/server/utils/mojang/MojangUtils.java
+++ b/src/main/java/net/minestom/server/utils/mojang/MojangUtils.java
@@ -2,7 +2,6 @@ package net.minestom.server.utils.mojang;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
-import net.minestom.server.MinecraftServer;
 import net.minestom.server.ServerFlag;
 import net.minestom.server.utils.url.URLUtils;
 import org.jetbrains.annotations.ApiStatus;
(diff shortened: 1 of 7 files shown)


LooFifteen (approved):

src/main/java/net/minestom/server/utils/mojang/MojangUtils.java
```diff
@@ -90,6 +102,29 @@ public final class MojangUtils {
...
+        ) {
+            url = String.format(PREVENT_PROXY_CONNECTIONS_AUTH_URL, username, serverId, address.getHostAddress());
+        } else {
+            url = String.format(BASE_AUTH_URL, username, serverId);
+        }
+
+        try {
+            return retrieve(url);
+        } catch (IOException e) {
+            MinecraftServer.getExceptionManager().handleException(e);
```
TheMode, line 123: Would prefer if you just threw a `RuntimeException`. The exception manager is already used in the listener

KermanIsPretty: Do you want the Login Listener to handle the exception, or should we let it be managed by the main packet listener try-catch?

If we handle it in the Login Listener, it might be best to let it throw an `IOException`. This way, we can use custom error handling for it, treating it as a Mojang error.

TheMode: In the listener seems fine. Also maybe explicitly define in `MojangUtils` signature the exception thrown

TheMode (approved):

PR #2387, opened 2024-09-08
Refactor GameMode to use functional abstraction

0x7CD6:
Replaced enum with a class to support dynamic predicates. Added methods for converting between names and IDs to improve flexibility. Updated references throughout the codebase for compatibility and extensibility.


TheMode:
Sorry I am against, I don't really see any reason for it to be a class

PR #2389, opened 2024-09-09
Add Block::nbtOrEmpty

Minikloon:
It's a common pattern when implementing BlockHandlers.

Only somewhat related rant (so matt sees):
image

From: https://discord.com/channels/706185253441634317/706186227493109860/1282587137681784882

### DIFF
--- a/src/main/java/net/minestom/server/instance/block/Block.java
+++ b/src/main/java/net/minestom/server/instance/block/Block.java
@@ -92,6 +92,16 @@ public sealed interface Block extends StaticProtocolObject, TagReadable, Blocks 
     @Contract(pure = true)
     @Nullable CompoundBinaryTag nbt();
 
+    /**
+     * Returns an unmodifiable view of the block nbt or an empty compound.
+     *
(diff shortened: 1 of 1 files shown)


TheMode:
Can be implemented as `Objects.requireNonNullElse(nbt(), CompoundBinaryTag.empty())` instead

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -92,6 +92,16 @@ public sealed interface Block extends StaticProtocolObject, TagReadable, Blocks
     @Contract(pure = true)
     @Nullable CompoundBinaryTag nbt();
 
+    /**
+     * Returns an unmodifiable view of the block nbt or an empty compound.
+     *
+     * @return the block nbt or an empty compound if not present
+     */
+    default @Nullable CompoundBinaryTag nbtOrEmpty() {
```
TheMode, line 100: Also, shouldnt it be `NotNull`?

PR #2394, opened 2024-09-10
feat: register multiple commands

DebitCardz:
Makes it easier to register multiple commands at once.

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandManager.java
+++ b/src/main/java/net/minestom/server/command/CommandManager.java
@@ -61,13 +61,13 @@ public final class CommandManager {
     }
 
     /**
-     * Register multiple {@link Command}.
+     * Register multiple {@link Command}s.
      *
      * @param commands the array of commands
      * @throws IllegalStateException if a command with the same name already exists
      */
-    public synchronized void register(@NotNull Command ...commands) {
-        for(Command command : commands) {
+    public synchronized void register(@NotNull Command... commands) {
+        for (Command command : commands) {
             register(command);
         }
     }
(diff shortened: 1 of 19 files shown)


mudkipdev:
I think it would be better if you added the vararg to the main register method instead of making an overload.

DebitCardz:
> I think it would be better if you added the vararg to the main register method instead of making an overload.
Maybe, but I really don't want to potentially mess with someones reflection garbage for a simple QoL change.

src/main/java/net/minestom/server/command/CommandManager.java
```diff
@@ -60,6 +60,18 @@ public synchronized void register(@NotNull Command command) {
         invalidateGraphCache();
     }
 
+    /**
+     * Register multiple {@link Command}.
```
mudkipdev, line 64: Should be plural "commands".

DebitCardz: Links to the actual Command class, would you prefer `Commands` or `{@link Command}s`?

mworzala: the second one

src/main/java/net/minestom/server/command/CommandManager.java
```diff
@@ -60,6 +60,18 @@ public synchronized void register(@NotNull Command command) {
...
+     *
+     * @param commands the array of commands
+     * @throws IllegalStateException if a command with the same name already exists
+     */
+    public synchronized void register(@NotNull Command ...commands) {
```
mworzala, line 69: ... should be before the space, and there should be a space after `for`

mworzala:
im fine with the overload

PR #2409, opened 2024-09-20
fix position used in the 'onPlace' method when loading

QuillDev:
When loading a chunk, the BlockHandler#onPlace() method returns the incorrect coordinates (chunk space instead of world space), this pr adds the chunk offset to correct the position to world space.

Before:
image
After:
image


mworzala (approved):
Nice catch, thanks

PR #2410, opened 2024-09-21
Fix javadoc

xZeGoe:
UuidProvider was replaced in favor of PlayerProvider and Acquirable#async(Consumer) was removed, both in dcb87355c6cb11784c7d20301f12fd0efc7cd2ef


mworzala:
Nice catch, thanks!

mworzala (approved):

PR #2411, opened 2024-09-21
Expose iteration methods of PPath

xZeGoe:


mworzala:
I am generally opposed to this. Those methods are definitely internal details. I will need a good justification for this.

xZeGoe:
This is primarily intended for manual pathfinding in [scratch](https://github.com/Minestom/Minestom/blob/d2cfbd2ad6f0b1331e7a66a111e2f7976f15141c/scratch/src/main/java/template/SurvivalTemplate.java#L624). If there is a nice way to do it, without exposing internals, I'd gladly close this PR

iam4722202468:
If you're implementing your own pathfinding why would you be using any internal minestom classes? This class is an implementation detail and can change in the future, relying on it instead of having your own class is a bad idea.

xZeGoe:
That's a fair point. I was trying to avoid duplicating effort and reuse existing functionality and mostly was just referring to Mode's code for it.

PR #2421, opened 2024-09-29
Fix Entity#switchEntityType method

N4TH4NOT:
Fix Entity#switchEntityType method to work for LivingEntity with not LivingEntity.
Basically fix disguises ^^


N4TH4NOT:
When I hit a player when I disguised in not living entity, the target receive again my properties (attributes) that cause the same error that pull request should fix 😵‍💫

PR #2429, opened 2024-10-06
Fix EffectChance network writer

GreatWyrm:
Relates to #2423.

Effect Chances were not written properly when sending them through a packet, resulting in a decoder exception whenever this list was not empty.


mworzala (approved):
Oops, awkward

PR #2433, opened 2024-10-07
Duplicate block property ID fix + unit test

boxic:
[This earlier commit](https://github.com/Minestom/Minestom/commit/a3dec124aaf20a0f4dadf0df7caeaa0aee4ff90a) does not properly account for the number of properties of blocks like noteblocks.

This hopefully fixes it.


boxic:
HOLY SHIT I DID IT, THANKS MODE!!!

boxic:
Okay, basically I had to make the properties array a long, due to this system not accounting for blocks with so many states. Should work now!

mworzala (approved):

PR #2437, opened 2024-10-08
Fix a few item component issues

wiidotmom:
The `color` field doesn't actually exist, it's called `rgb`
https://minecraft.wiki/w/Data_component_format#dyed_color
image

Book pages don't have a `text` field, but they do have a `raw` field
https://minecraft.wiki/w/Data_component_format#written_book_content
image


mworzala (approved):
Nice catches, thank you!

PR #2443, opened 2024-10-21
Update Pos method names for clarity

ArcticRaven:
renamed `Pos.direction()` to `Pos.getLookVector()` for clarity; consideration for Direction object.

### DIFF
--- a/src/main/java/net/minestom/server/coordinate/Pos.java
+++ b/src/main/java/net/minestom/server/coordinate/Pos.java
@@ -142,13 +142,17 @@ public record Pos(double x, double y, double z, float yaw, float pitch) implemen
     }
 
     /**
-     * Gets a unit-vector pointing in the direction that this Location is
-     * facing.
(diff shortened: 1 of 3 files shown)


GreatWyrm:
A couple of things:
1. You didn't adjust the tests that use this method, so they're failing now. Please fix them.
2. I'd recommend instead you deprecate the old `direction()` method and create a new one called `lookVector()` (Sidenote but the new javadocs you wrote say it's `lookVector()` and not `getLookVector()`, I'd recommend using `lookVector()` instead to fit with the other functions' style in Pos).

GreatWyrm:
LGTM

mworzala:
I think I'm opposed to this. It is clearer for some people less so for others, and I am not keen on getting into a habit of making breaking changes like this for the sake of a name. I would prefer to simply add some extra terminology to the direction javadoc to make it more accessible.

Happy to discuss more

PR #2446, opened 2024-10-23
AsyncPlayerPreLoginEvent getConnection

MelonHell:
image


src/main/java/net/minestom/server/event/player/AsyncPlayerPreLoginEvent.java
```diff
@@ -16,11 +17,14 @@
 public class AsyncPlayerPreLoginEvent implements Event {
 
     private GameProfile gameProfile;
+    private final PlayerConnection connection;
```
mworzala, line 20: nitpick, but connection should be ordered before gameprofile

MelonHell: I don't understand why, but okay

PR #2448, opened 2024-10-24
Batch rework

AidanMars2:
Batches now use completable futures.


KrystilizeNevaDies:
What's the performance impact of this pr? i.e. does it improve or degrade?
Also instead of a `long getLastBlockChangeTime` why not use a hash instead?
Could be something like `int getBlockHash` or `void blockHash(MessageDigest)`.
If all we need to know is that the blocks have changed, a hash tells that, and there won't be issues with placing a block back after it being broken within the same tick for example.

AidanMars2:
I haven't benchmarked it yet because I'm a bit busy, but I expect performance to degrade slightly. Partially because it will _actually_ try to send light updates now, and partially because `AbsoluteBlockBatch` uses one `CompletableFuture` in a way that doesn't quite sit right with me.

TheMeinerLP (changes requested):
Can you update your PR?

src/main/java/net/minestom/server/instance/batch/AbsoluteBlockBatch.java
```diff
@@ -60,130 +51,84 @@ public void setBlock(int x, int y, int z, @NotNull Block block) {
...
-        return inverse;
-    }
+        return CompletableFuture.supplyAsync(() -> {
+            // await chunkBatch futures and collect inverse batches
+            final AbsoluteBlockBatch inverse = this.options.shouldCalculateInverse() ?
+                    new AbsoluteBlockBatch(inverseOption) : null;
+
+            for (int index = 0; index < chunkBatchIndexes.size(); index++) {
+                final long chunkIndex = chunkBatchIndexes.getLong(index);
+                final ChunkBatch chunkInverse = chunkBatchFutures.get(index).join();
```
TheMeinerLP, line 96: This blocks the thread from the future. I would prefer https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#allOf-java.util.concurrent.CompletableFuture...- or make a Queue.

src/main/java/net/minestom/server/instance/batch/ChunkBatch.java
```diff
@@ -99,69 +87,43 @@ public ChunkBatch apply(@NotNull Instance instance, @Nullable ChunkCallback call
...
-
-        final ChunkBatch inverse = this.options.shouldCalculateInverse() ? new ChunkBatch(options, false) : null;
-        BLOCK_BATCH_POOL.execute(() -> singleThreadFlush(instance, chunk, inverse, callback, safeCallback));
-        return inverse;
+    protected @NotNull CompletableFuture<@Nullable ChunkBatch> apply(
+            @NotNull Instance instance, @NotNull Chunk chunk, boolean subBatch) {
+        return CompletableFuture.supplyAsync(() -> {
+            final ChunkBatch inverse = this.options.shouldCalculateInverse() ? new ChunkBatch(inverseOption) : null;
+            synchronized (chunk) {
+                singleThreadFlush(instance, chunk, inverse, subBatch);
```
TheMeinerLP, line 117: That looks very dirty inside a supply async

src/main/java/net/minestom/server/utils/location/LocationUtils.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param z the block Z
+     * @return an index which can be used to store and retrieve later data linked to a block position
+     */
+    public static long getGlobalBlockIndex(int x, int y, int z) {
+//        Check.argCondition(((x & HORIZONTAL_BIT_MASK) << 6 >> 6) != x,
+//                "x position may not be more than 26 bits long (value: {})", x);
+//        Check.argCondition(((y & VERTICAL_BIT_MASK) << 20 >> 20) != y,
+//                "y position may not be more than 12 bits long (value: {})", y);
+//        Check.argCondition(((z & HORIZONTAL_BIT_MASK) << 6 >> 6) != z,
+//                "z position may not be more than 26 bits long (value: {})", z);
```
TheMeinerLP, line 31: Please remove out comment code

src/main/java/net/minestom/server/utils/location/LocationUtils.java
```diff
@@ -0,0 +1,89 @@
+package net.minestom.server.utils.location;
+
+import net.minestom.server.coordinate.BlockVec;
+import net.minestom.server.coordinate.Point;
+import net.minestom.server.utils.validate.Check;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+
+@ApiStatus.Internal
+public class LocationUtils {
```
TheMeinerLP, line 10: Nitpick: Add final

src/main/java/net/minestom/server/utils/location/LocationUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Converts a block index to a position Y.
+     *
+     * @param index an index computed from {@link #getGlobalBlockIndex(int, int, int)}
+     * @return the position Y of the index
+     */
+    public static int globalBlockIndexToPositionY(long index) {
+        return (int) (index << 52 >> 52); // 0-12 bits
```
TheMeinerLP, line 77: What means the magic values here ?
Is there some proof or documentation about it ?

src/main/java/net/minestom/server/utils/location/LocationUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Converts a block index to a position X.
+     *
+     * @param index an index computed from {@link #getGlobalBlockIndex(int, int, int)}
+     * @return the position X of the index
+     */
+    public static int globalBlockIndexToPositionX(long index) {
+        return (int) (index >> 38); // 38-64 bits
```
TheMeinerLP, line 67: Nitpick: Magic Value

src/main/java/net/minestom/server/utils/location/LocationUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Converts a block index to a position Z.
+     *
+     * @param index an index computed from {@link #getGlobalBlockIndex(int, int, int)}
+     * @return the position Z of the index
+     */
+    public static int globalBlockIndexToPositionZ(long index) {
+        return (int) (index << 26 >> 38); // 12-38 bits
```
TheMeinerLP, line 87: Nitpick: Magic value

src/test/java/net/minestom/server/instance/batch/AbsoluteBlockBatchTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+// These tests also cover ChunkBatch
+@EnvTest
+public class AbsoluteBlockBatchTest {
```
TheMeinerLP, line 21: Nitpick: Remove public

src/test/java/net/minestom/server/instance/batch/AbsoluteBlockBatchTest.java
```diff
@@ -0,0 +1,108 @@
...
+    @Test
+    public void singleChunk(Env env) {
```
TheMeinerLP, line 23: Dry (Don't repeat yourself)

src/test/java/net/minestom/server/instance/batch/AbsoluteBlockBatchTest.java
```diff
@@ -0,0 +1,108 @@
...
+        final var inverse = batch.apply(instance).join();
+        assertNotNull(inverse, "Inverse batch null with BatchOption#calculateInverse true");
+        assertEquals(instance.getBlock(point), Block.GRASS_BLOCK);
+
+        inverse.apply(instance).join();
+        assertEquals(originalBlock, instance.getBlock(point));
+    }
+
+    @Test
+    public void multipleChunks(Env env) {
```
TheMeinerLP, line 46: Dry (Don't repeat yourself)

src/test/java/net/minestom/server/instance/batch/AbsoluteBlockBatchTest.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        inverse.apply(instance).join();
+        for (int index = 0; index < points.size(); index++) {
+            assertEquals(originalBlocks.get(index), instance.getBlock(points.get(index)));
+        }
+    }
+
+    @Test
+    public void withNbt(Env env) {
```
TheMeinerLP, line 89: Dry (Don't repeat yourself)

TheMeinerLP (changes requested):

src/test/java/net/minestom/server/instance/batch/AbsoluteBlockBatchTest.java
TheMeinerLP: It's against the JUnit 5 recommendations.
Ref: https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

src/main/java/net/minestom/server/utils/location/LocationUtils.java
AidanMars2: The bit packing is equal to what minecraft does over the network as described [here](https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol#Type:Position) and implemented by Minestom [here](https://github.com/Minestom/Minestom/blob/0d47d974177cdb277aad3b4d450e380d8e659eb6/src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java#L391).

src/main/java/net/minestom/server/instance/batch/AbsoluteBlockBatch.java
AidanMars2: I didn't know that existed, thanks!

src/test/java/net/minestom/server/instance/batch/AbsoluteBlockBatchTest.java
AidanMars2: I removed the `public` classifier from the Batch test classes for now, it might be worth discussing whether we want to make that the case for all tests.

src/main/java/net/minestom/server/utils/location/LocationUtils.java
KermanIsPretty, line 1: Any reason this exists instead of using CoordConversion?

PR #2449, opened 2024-10-24
fix LoginListener.enterConfig

MelonHell:
Blocking the AsyncPlayerPreLoginEvent event blocks receiving a cookie response from the client


mworzala (approved):
Seems to be an oversight in the vthread change. thanks!

PR #2458, opened 2024-10-30
Fix multi login with offline accounts

Athishh:
image

### DIFF
--- a/src/main/java/net/minestom/server/listener/preplay/LoginListener.java
+++ b/src/main/java/net/minestom/server/listener/preplay/LoginListener.java
@@ -62,13 +62,14 @@ public final class LoginListener {
             }
         }
 
+        if (CONNECTION_MANAGER.getOnlinePlayerByUsername(packet.username()) != null) {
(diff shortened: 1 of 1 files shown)


MelonHell:
This code does not take into account the possibility to change the username in AsyncPreLoginEvent. For example, we ignore usernames from packages, and use our own system of usernames, in this case, if there is a player with the nickname A on the server, another player whose nickname A in the launcher will not be able to log on to the server, although he would have been assigned nickname B using a custom nickname system

MelonHell:
This code does not take into account the possibility to change the username in AsyncPreLoginEvent. For example, we ignore usernames from packages, and use our own system of usernames, in this case, if there is a player with the nickname A on the server, another player whose nickname A in the launcher will not be able to log on to the server, although he would have been assigned nickname B using a custom nickname system

I think if you do this, then after the event. But I believe that this code should not be part of Minestom, but should be part of your server software implementation, or at least be an optional part of Minestom

PR #2459, opened 2024-10-31
Make entity interaction dependent on range

GreatWyrm:
- Makes entity interaction work with the interaction attribute
- Adds a server flag to disable checking of the range limit
- Change the MAP_RGB_REDUCTION flag to be an int so there is no int parsing messiness in MapColors.


iam4722202468:
Can you move the map change to another PR?

GreatWyrm:
@iam4722202468 Done, other PR is here: https://github.com/Minestom/Minestom/pull/2460

PR #2467, opened 2024-11-05
Fix damage invulnerablity checks

### LINKED ISSUE #2466: Player#damage does nothing in 1.21.2
In the 1.21.2 branch, calling `Player#damage` doesn't do anything, from a quick glance it seems to be related to [this method](https://github.com/Minestom/Minestom/blob/1_21_2/src/main/java/net/minestom/server/entity/Player.java#L1017) returning the wrong value even when the player is in survival mode.

GreatWyrm:
There are a couple of problems with the current damage checks:

1. The invulnerable field was named `canTakeDamage`, guaranteeing confusion, as if this was true, that means they cannot take damage.
2. The `damage()` function checked invulnerability before checking immunity, skipping the special case of OUT_OF_WORLD damage being able to damage players while invulnerable.

I have renamed the field and adjusted the damage function so it only performs the isImmune check, and moved the logic to the isImmune check.
Resolves #2466.

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -326,7 +326,7 @@ public class LivingEntity extends Entity implements EquipmentHandler {
     public boolean damage(@NotNull Damage damage) {
         if (isDead())
             return false;
-        if (isInvulnerable() || isImmune(damage.getType())) {
+        if (isImmune(damage.getType())) {
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -384,7 +384,10 @@ public boolean damage(@NotNull Damage damage) {
      * @return true if this entity is immune to the given type of damage
      */
     public boolean isImmune(@NotNull DynamicRegistry.Key<DamageType> type) {
-        return false;
+        if (isInvulnerable() && type == DamageType.OUT_OF_WORLD) {
```
mworzala, line 387: should use `#equals` here. it is a constant but it is not referentially unique.

mworzala, line 387: I think this logic would be clearer if it was simply:

```
if (type.equals(OUT_OF_WORLD) return false;
return isInvulnerable()
```

Checking isInvulnerable twice seems like it just adds confusion

GreatWyrm: Changed to match.

PR #2468, opened 2024-11-05
public MetadataHolder.createMeta

MelonHell:
This PR allows you to create an EntityMeta instance manually, which is useful for scratch projects

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataHolder.java
+++ b/src/main/java/net/minestom/server/entity/MetadataHolder.java
@@ -132,7 +132,7 @@ public final class MetadataHolder {
 
     static final Map<String, BiFunction<Entity, MetadataHolder, EntityMeta>> ENTITY_META_SUPPLIER = createMetaMap();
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/entity/MetadataHolder.java
```diff
@@ -132,7 +132,7 @@ public void setNotifyAboutChanges(boolean notifyAboutChanges) {
 
     static final Map<String, BiFunction<Entity, MetadataHolder, EntityMeta>> ENTITY_META_SUPPLIER = createMetaMap();
 
-    static EntityMeta createMeta(EntityType entityType, Entity entity, MetadataHolder metadata) {
+    public static EntityMeta createMeta(EntityType entityType, Entity entity, MetadataHolder metadata) {
```
mworzala, line 135: ApiStatus.Internal, NotNull, Nullable, NotNull.

I am fine with this as a temporary solution for doing weird things with metadata, but I do not want to encourage it.

mworzala (changes requested):

mworzala (approved):

PR #2469, opened 2024-11-05
Ability to get registry packets without MinecraftServer

MelonHell:
This PR allows creating registry packets without accessing MinecraftServer, this is necessary for scrach servers. Does not break compatibility in any way

### DIFF
--- a/src/main/java/net/minestom/server/registry/DynamicRegistry.java
+++ b/src/main/java/net/minestom/server/registry/DynamicRegistry.java
@@ -213,10 +213,11 @@ public sealed interface DynamicRegistry<T> permits DynamicRegistryImpl {
      * <p>Returns a {@link SendablePacket} potentially excluding vanilla entries if possible. It is never possible to
      * exclude vanilla entries if one has been overridden (e.g. via {@link #register(NamespaceID, T)}.</p>
      *
+     * @param registries Registries provider
      * @param excludeVanilla Whether to exclude vanilla entries
      * @return A {@link SendablePacket} containing the registry data
      */
     @ApiStatus.Internal
-    @NotNull SendablePacket registryDataPacket(boolean excludeVanilla);
+    @NotNull SendablePacket registryDataPacket(Registries registries, boolean excludeVanilla);
 
 }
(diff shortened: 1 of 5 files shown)


src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -163,23 +164,32 @@ public boolean remove(@NotNull NamespaceID namespaceId) throws UnsupportedOperat
...
         }
     }
 
     @Override
-    public @NotNull SendablePacket registryDataPacket(boolean excludeVanilla) {
+    public @NotNull SendablePacket registryDataPacket(Registries registries, boolean excludeVanilla) {
         // We cache the vanilla packet because that is by far the most common case. If some client claims not to have
         // the vanilla datapack we can compute the entire thing.
-        return excludeVanilla ? vanillaRegistryDataPacket : createRegistryDataPacket(false);
+        if (excludeVanilla) {
```
mworzala, line 180: This feels very gross, and also doesn't handle the case where we get a different Registries instance on the second call (which is weird, but you are exposing that by adding this parameter).

Either this should also track which registries it was and invalidate if we get a different one, or just take registries when constructing it would probably be fine too.

If you keep this parameter, it needs a NotNull annotation.

src/main/java/net/minestom/server/registry/DynamicRegistry.java
```diff
@@ -213,10 +213,11 @@ default int getId(@NotNull Key<T> key) {
      * <p>Returns a {@link SendablePacket} potentially excluding vanilla entries if possible. It is never possible to
      * exclude vanilla entries if one has been overridden (e.g. via {@link #register(NamespaceID, T)}.</p>
      *
+     * @param registries Registries provider
      * @param excludeVanilla Whether to exclude vanilla entries
      * @return A {@link SendablePacket} containing the registry data
      */
     @ApiStatus.Internal
-    @NotNull SendablePacket registryDataPacket(boolean excludeVanilla);
+    @NotNull SendablePacket registryDataPacket(Registries registries, boolean excludeVanilla);
```
mworzala, line 221: NotNull, but also see later comment

mworzala (changes requested):

mworzala (approved):

PR #2471, opened 2024-11-07
Rewritten all EntityMeta classes to use MetadataDef

MelonHell:
I really don't like code duplication, which doubles the chance of errors, several errors in MetadataDef were fixed in the process of writing this PR.

I really hope this PR gets accepted, as I spent several days on it XD

Does not break backward compatibility, except for low-level calls to Metadata


MelonHell:
I managed to implement Wolf.VARIANT, it turned out I specified the generic type incorrectly :D

MelonHell:
Added ..._FLAGS fields to all MetadataDefs with flags, I think this can be useful in cases where someone wants to manually specify the entire byte

MelonHell:
All that's left is to decide what to do with Sheep.COLOR_ID and the PR will be completely ready. I need help with this, I have no idea how to make Sheep.COLOR_ID

mworzala (approved):
Very nice, thank you for doing the busy work :)

Issue #2196, opened 2024-06-16
Feature: accept transfer connections in HandshakeListener


paolobettelini:
Minestom does not accept connections resulting from transfers packets. Currently, an `UnsupportedOperationException` is thrown

https://github.com/Minestom/Minestom/blob/8736fc43d8e0790f6cc25d4c4a372c079b66fdaa/src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java#L118

ajh123:
Instead of accepting all transfer connections maybe an event could be queued that will:
* if not cancelled allow the transfer
* if cancelled stop the transfer

This will allow developers to have a greater control of when a transfer is allowed.

PR #2473, opened 2024-11-08
Transfer intent implementation

AllroundYT:
+ Accepts connections via transfer
+ Connection rules added: ALL, ONLY_DIRECT, ONLY_TRANSFER, NONE
+ Added Player#transfer Method
+ Added PrePlayerTransferEvent & PostPlayerTransferEvent

#2196

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -22,6 +22,7 @@ import net.minestom.server.message.ChatType;
 import net.minestom.server.monitoring.BenchmarkManager;
 import net.minestom.server.network.ConnectionManager;
 import net.minestom.server.network.PacketProcessor;
+import net.minestom.server.network.packet.client.handshake.ClientHandshakePacket;
 import net.minestom.server.network.packet.server.common.PluginMessagePacket;
 import net.minestom.server.network.packet.server.play.ServerDifficultyPacket;
 import net.minestom.server.network.socket.Server;
@@ -72,6 +73,7 @@ public final class MinecraftServer implements MinecraftConstants {
     private static int compressionThreshold = 256;
     private static String brandName = "Minestom";
     private static Difficulty difficulty = Difficulty.NORMAL;
+    private static ClientHandshakePacket.ConnectionRule connectionRule = ClientHandshakePacket.ConnectionRule.ALL;
 
     public static MinecraftServer init() {
         updateProcess();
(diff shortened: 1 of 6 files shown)


src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -318,6 +320,14 @@ public static Server getServer() {
         return serverProcess.server();
     }
 
+    public static ClientHandshakePacket.ConnectionRule getConnectionRule() {
```
mworzala, line 323: this should not exist. people can use an event to prevent transfers if they want that feature.

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -271,6 +275,18 @@ public Player(@NotNull UUID uuid, @NotNull String username, @NotNull PlayerConne
         metadata.setNotifyAboutChanges(false);
     }
 
+    public void transfer(@NotNull InetSocketAddress address) {
+        this.transfer(address.getAddress().getHostAddress(), address.getPort());
+    }
+
+    public void transfer(@NotNull String host, int port) {
+        TransferPacket packet = new TransferPacket(host, port);
+        EventDispatcher.call(new PrePlayerTransferEvent(this, host, port));
```
mworzala, line 284: I think this can just be PlayerTransfer event, having pre and post seems excessive. It probably should be cancellable also

src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
```diff
@@ -48,74 +51,40 @@ public static void listener(@NotNull ClientHandshakePacket packet, @NotNull Play
...
+                    disconnect(connection, INVALID_VERSION_TEXT);
+                }
+
+                if (!MinecraftServer.getConnectionRule().isAllowed(ClientHandshakePacket.Intent.TRANSFER)) {
+                    // Transfer not allowed
+                    disconnect(connection, INVALID_VERSION_TEXT);
+                }
+
+                // Bungee support (IP forwarding)
+                String bungeeAddress = bungeeConnect(address, connection);
```
mworzala, line 81: Is this a real case? If you transfer to a bungee proxy I would assume it does normal login flow not transfer flow. I would think it's just not possible to receive transfers if velocity or bungee is enabled?

src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
```diff
@@ -48,74 +51,40 @@ public static void listener(@NotNull ClientHandshakePacket packet, @NotNull Play
...
-                    }
+                if (bungeeAddress == null) {
+                    return;
                 }
+
+                address = bungeeAddress;
             }
             case TRANSFER -> {
-                throw new UnsupportedOperationException("Transfer intent is not supported in HandshakeListener");
+                connection.setConnectionState(ConnectionState.LOGIN);
```
mworzala, line 69: The login event should have an isTransfer field which would allow people to handle disallowing transfers if they choose.

TheMeinerLP:
Duplicated of #2484

mworzala:
Superseded by https://github.com/Minestom/Minestom/pull/2484

PR #2484, opened 2024-11-17
Implement transfer intent

### LINKED ISSUE #2196: Feature: accept transfer connections in HandshakeListener

mudkipdev:
Resolves #2196, closes #2473.

- Adds `OutgoingTransferEvent`.
- Adds `PlayerConnection#transfer(String, int)`.
- Adds `PlayerConnection#wasTransferred()`.
- Adds `/transfer` command to the demo project.
- I'm unsure whether the comment about Bungee in #2473. Needs to be tested.

### DIFF
--- a/src/main/java/net/minestom/server/event/player/OutgoingTransferEvent.java
+++ b/src/main/java/net/minestom/server/event/player/OutgoingTransferEvent.java
@@ -1,0 +1,73 @@
+package net.minestom.server.event.player;
+
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.trait.CancellableEvent;
+import net.minestom.server.event.trait.PlayerEvent;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Called when a {@link Player} is about to be redirected to another server.
+ */
+public class OutgoingTransferEvent implements PlayerEvent, CancellableEvent {
+    private final @NotNull Player player;
+    private @NotNull String host;
+    private int port;
+    private boolean cancelled;
+
+    public OutgoingTransferEvent(@NotNull Player player, @NotNull String host, int port) {
+        this.player = player;
+        this.host = host;
+        this.port = port;
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return this.cancelled;
+    }
+
+    @Override
+    public void setCancelled(boolean cancel) {
+        this.cancelled = cancel;
+    }
+
+    /**
+     * Returns the player who is being transferred.
+     * @return The player who is being transferred.
+     */
+    @Override
+    public @NotNull Player getPlayer() {
+        return this.player;
+    }
+
+    /**
+     * Returns the address of the server that the player is being transferred to.
+     * @return The address of the server that the player is being transferred to.
+     */
+    public @NotNull String getHost() {
+        return this.host;
+    }
+
+    /**
+     * Returns the port of the server that the player is being transferred to.
+     * @return The port of the server that the player is being transferred to.
+     */
+    public int getPort() {
+        return this.port;
+    }
+
+    /**
+     * Changes the address of the server that the player will be transferred to.
+     * @param host The address of the new server that the player will be transferred to
+     */
+    public void setHost(@NotNull String host) {
+        this.host = host;
+    }
+
+    /**
+     * Changes the port of the server that the player will be transferred to.
+     * @param port The port of the new server that the player will be transferred to
+     */
+    public void setPort(int port) {
+        this.port = port;
+    }
+}
--- a/src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
+++ b/src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
@@ -38,93 +38,99 @@ public final class HandshakeListener {
     private static final Component INVALID_BUNGEE_FORWARDING = Component.text("Invalid connection, please connect through the BungeeCord proxy. If you believe this is an error, contact a server administrator.", NamedTextColor.RED);
 
     public static void listener(@NotNull ClientHandshakePacket packet, @NotNull PlayerConnection connection) {
-        String address = packet.serverAddress();
         switch (packet.intent()) {
             case STATUS -> connection.setConnectionState(ConnectionState.STATUS);
-            case LOGIN -> {
+            case LOGIN, TRANSFER -> {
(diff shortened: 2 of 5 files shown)


src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
```diff
@@ -38,93 +38,99 @@ public final class HandshakeListener {
...
+            }
+        }
+
+        if (connection instanceof PlayerSocketConnection) {
+            // Give to the connection the server info that the client used
+            ((PlayerSocketConnection) connection).refreshServerInformation(
+                    packet.serverAddress(),
+                    packet.serverPort(),
+                    packet.protocolVersion());
+        }
```
LooFifteen, line 55: You can invert the condition to reduce nesting here.
```suggestion
        if (!(connection instanceof PlayerSocketConnection connection)) return;
        // Give to the connection the server info that the client used
        connection.refreshServerInformation(
                packet.serverAddress(),
                packet.serverPort(),
                packet.protocolVersion()
        );
```

LooFifteen (changes requested):

TheMeinerLP:
Duplicated of #2473

src/main/java/net/minestom/server/network/player/PlayerConnection.java
```diff
@@ -281,6 +285,37 @@ public void receiveKnownPacksResponse(List<SelectKnownPacksPacket.Entry> clientP
...
+     * @param port The port of the server to transfer the player to.
+     */
+    public void transfer(@NotNull String host, int port) {
+        OutgoingTransferEvent event = new OutgoingTransferEvent(this.player, host, port);
+        EventDispatcher.callCancellable(event, () -> this.sendPacket(new TransferPacket(event.getHost(), event.getPort())));
+    }
+
+    /**
+     * Returns whether the player was redirected from another server.
+     * @return Whether the player was redirected from another server.
```
kermandev: This doesnt seem a very helpful Javadoc, maybe its most akin to the client notified the server it was transferred

src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
```diff
@@ -35,95 +36,99 @@ public final class HandshakeListener {
...
-                                        bungeeDisconnect(socketConnection);
-                                        return;
-                                    }
-
-                                    foundBungeeGuardToken = true;
-                                }
-
-                                properties.add(new GameProfile.Property(nameString, valueString, signatureString));
-                            }
+        connection.markTransferred(packet.intent() == ClientHandshakePacket.Intent.TRANSFER);
```
kermandev: Shouldnt the intent be marked before this is handled? Maybe the switch should use the switch statement that falls through?

src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
```diff
@@ -35,95 +36,99 @@ public final class HandshakeListener {
...
-                        boolean hasProperties = split.length == 4;
-                        if (bungee.guard() && !hasProperties) {
-                            bungeeDisconnect(socketConnection);
-                            return;
-                        }
+    private static String handlePlayerLogin(@NotNull ClientHandshakePacket packet, @NotNull PlayerConnection connection,
+                                            @NotNull String address) {
+        if (packet.protocolVersion() != MinecraftServer.PROTOCOL_VERSION) {
+            // Incorrect client version
+            connection.kick(INVALID_VERSION_TEXT);
```
kermandev: This should probably return here, weird its missing

Also, why break this out into a separate function?

src/main/java/net/minestom/server/event/player/OutgoingTransferEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+import net.minestom.server.entity.Player;
+import net.minestom.server.event.trait.CancellableEvent;
+import net.minestom.server.event.trait.PlayerEvent;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Called when a {@link Player} is about to be redirected to another server.
+ */
+public class OutgoingTransferEvent implements PlayerEvent, CancellableEvent {
```
kermandev, line 14: We dont use @NotNull anymore as this should already be marked.

Also please use Objects.requireNonNull for this.player and this.host including reassignment.

src/main/java/net/minestom/server/event/player/OutgoingTransferEvent.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public void setCancelled(boolean cancel) {
+        this.cancelled = cancel;
+    }
+
+    /**
+     * Returns the player who is being transferred.
+     * @return The player who is being transferred.
```
kermandev: This is likely redundant due to the trait

src/main/java/net/minestom/server/event/player/OutgoingTransferEvent.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    @Override
+    public @NotNull Player getPlayer() {
+        return this.player;
+    }
+
+    /**
+     * Returns the address of the server that the player is being transferred to.
+     * @return The address of the server that the player is being transferred to.
```
kermandev: Probably should be worded as the current target address and port

src/main/java/net/minestom/server/listener/preplay/HandshakeListener.java
mudkipdev: I dont understand, can you elaborate on "the switch should use the switch statement that falls through"?

kermandev: instead of a switch statements with -> you use one with : to let it fall through. So you can set the intent and run the regular login path.

mworzala (changes requested):
Transfers should be disabled by default and enabled via a ServerFlag. Should send the vanilla disconnect message translation if disabled.

PR #2490, opened 2024-11-24
fix painting variants

### LINKED ISSUE #2479: Some painting variants causes client disconnection
**Summary**
I tried to write a painting loader which could load paintings from Anvil worlds, but some variants make my client disconnect instantly, when I can join the variants are incorrect.

Here's a link to my code: [repository](https://github.com/bluroch/elderwood-experiments/tree/base-code)

**Details**
Variants are incorrect on the loaded paintings, also some variants are making clients disconnect.
Paintings on Minestom
Paintings on Paper
I included a debug world which has all painting variants with all orientations, see `worlds/gallery` in the linked repository. This is the world which makes the clients disconnect, the images are from a separate world that I want to load.

I suspect that I can join when that world is loaded only because I didn't used all variants there.

The repository also contains one of my disconnect logs: `logs/client/painting_error.txt`

mesoscopic:
Fixes #2479.
Paintings appeared incorrectly and sometimes disconnected the client due to both the registry being incorrectly ordered (now ordered alphabetically with `Comparators.naturalOrder()` like vanilla) and the fact that, for whatever reason, the registry starts at 1 on the client (fixed by adding 1 to the value if the registry is `minecraft:painting_variant`).

### DIFF
--- a/src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
+++ b/src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
@@ -631,7 +631,8 @@ interface NetworkBufferTypeImpl<T> extends NetworkBuffer.Type<T> {
         public void write(@NotNull NetworkBuffer buffer, DynamicRegistry.Key<T> value) {
             Check.stateCondition(buffer.registries == null, "Buffer does not have registries");
             final DynamicRegistry<T> registry = selector.apply(buffer.registries);
-            // For whatever reason, the client indexes painting variants starting at 1, so we add 1 in that case.
+            // Painting variants may be sent in their entirety rather than a registry reference so the ID is offset by 1 to indicate this.
+            // FIXME: Support sending the entire registry object instead of an ID reference.
             final int id = registry.id().equals("minecraft:painting_variant")?registry.getId(value)+1:registry.getId(value);
             Check.argCondition(id == -1, "Key is not registered: {0} > {1}", registry, value);
             buffer.write(VAR_INT, id);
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
```diff
@@ -631,7 +631,7 @@ record RegistryTypeType<T extends ProtocolObject>(@NotNull Function<Registries,
         public void write(@NotNull NetworkBuffer buffer, DynamicRegistry.Key<T> value) {
             Check.stateCondition(buffer.registries == null, "Buffer does not have registries");
             final DynamicRegistry<T> registry = selector.apply(buffer.registries);
-            final int id = registry.getId(value);
+            final int id = (registry.id() == "minecraft:painting_variant")?registry.getId(value)+1:registry.getId(value);
```
mudkipdev, line 634:
```suggestion
            final int id = registry.id().equals("minecraft:painting_variant")?registry.getId(value)+1:registry.getId(value);
```

mudkipdev (changes requested):

iam4722202468:
Weird, can you add a comment in the code so people in the future understand why you're doing this

src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
```diff
@@ -631,7 +631,8 @@ record RegistryTypeType<T extends ProtocolObject>(@NotNull Function<Registries,
         public void write(@NotNull NetworkBuffer buffer, DynamicRegistry.Key<T> value) {
             Check.stateCondition(buffer.registries == null, "Buffer does not have registries");
             final DynamicRegistry<T> registry = selector.apply(buffer.registries);
-            final int id = registry.getId(value);
+            // For whatever reason, the client indexes painting variants starting at 1, so we add 1 in that case.
+            final int id = registry.id().equals("minecraft:painting_variant")?registry.getId(value)+1:registry.getId(value);
```
mworzala, line 635: Its not random, its because painting variants are encoded as a (mojang term) "holder", which means either its a painting id from the registry or you can send the entire painting variant data directly in which case you write an id of 0 and the data following it.

We don't have proper support for these holder types for now so I am ok leaving this with a fixme. Something like:
```
// Painting variants may be sent in their entirety rather than a registry reference so the ID is offset by 1 to indicate this.
// FIXME: Support sending the entire registry object instead of an ID reference.
final int id = registry.id().equals("minecraft:painting_variant") ? registry.getId(value) + 1 : registry.getId(value);
```

mworzala (changes requested):

mworzala (approved):

PR #2493, opened 2024-11-26
Add brightness helpers for AbstractDisplayMeta

leon0017:
`setBrightnessOverride(int value)` can be a bit confusing, as it could seem it takes in a light level but takes in the combined block light and sky light in a single int.


mudkipdev:
Why not have:
```
int getBlockLight();
int getSkyLight();
void setBrightness(int blockLight, int skyLight);
```

leon0017:
That seems like the better option

leon0017:
Got confused with the shifting there

mworzala (approved):

PR #2497, opened 2024-11-28
Rename Effects to World Events, update world event list

GreatWyrm:
Rename Effects to World Events to align with https://wiki.vg/Protocol#World_Event
Update list of World Events to match wiki.vg


mudkipdev:
I think we should follow the naming conventions of mojang, but the naming is the same in this case (`LevelEvent` vs `WorldEvent`)

mworzala:
The events and their IDs seem like they should be generated to me if its reasonable.

GreatWyrm:
I agree, I think they should be generated.
I did take a look and I unfortunately could not find an easy way to generate them.

mworzala:
Here you go! :) https://github.com/Minestom/MinestomDataGenerator/pull/36

GreatWyrm:
Whoa, appreciate it! Once that's merged and the data gen version is bumped, I can add that to the auto-generated data.

GreatWyrm:
Superseded by #2499.

Issue #2492, opened 2024-11-26
feat: API for custom, server-side attributes


Steanky:
It would be useful to be able to create non-vanilla attributes. This was possible before attributes were moved to the current Registry-based system.

Use cases vary, but I imagine this feature would be used most often to facilitate custom, adjustable (via modifiers) gameplay mechanics that exist server-side. It would also be useful in the case of supporting modded clients that recognize some non-vanilla attributes.

As far as I can tell from reviewing the code as-is, this would require:
* Separate storage for non-vanilla Attribute instances
* A user-friendly (i.e. non-`@ApiStatus.Internal`) method of creating new Attribute instances
* Ability to exclude certain attributes from being sent to players in an `EntityAttributesPacket`

If this feature gets positive feedback, I'd be willing to work on it alongside a fix for #2491, and if not, a standalone fix for #2491.

mworzala:
I think this is a reasonable feature to support, though it should be done separately from #2491 I think.

PR #2498, opened 2024-11-29
feat: make attributes use a DynamicRegistry + API for custom attributes

Steanky:
Closes https://github.com/Minestom/Minestom/issues/2492

Draft to indicate some lack of confidence in correctness, as I'm not overly familiar with DynamicRegistries. Tests pass and basic usage works fine, so it's probably not broken too badly :slightly_smiling_face:. Regardless, any feedback is appreciated & I'm happy to make modifications if needed.

General summary:

* Updated the datagen to write `DynamicRegistry.Key<Attribute>`s instead of `Attribute` instances
* A `DynamicRegistry<Attribute>` instance is now available from the `ServerProcess` alongside other registries
* Usages of `Attribute` have been updated to use `DynamicRegistry.Key<Attribute>` instead, where appropriate
* LivingEntity won't notify the client of attribute changes where the underlying attribute's `isSynced` field is false. This prevents "server-side" attributes from crashing clients.


imBuzz:
Hey! Any news about this ?

code-generators/src/main/java/net/minestom/codegen/Generators.java
```diff
@@ -49,7 +49,6 @@ public static void main(String[] args) {
         generator.generate(resource("potions.json"), "net.minestom.server.potion", "PotionType", "PotionTypeImpl", "PotionTypes");
         generator.generate(resource("sounds.json"), "net.minestom.server.sound", "SoundEvent", "BuiltinSoundEvent", "SoundEvents");
         generator.generate(resource("custom_statistics.json"), "net.minestom.server.statistic", "StatisticType", "StatisticTypeImpl", "StatisticTypes");
-        generator.generate(resource("attributes.json"), "net.minestom.server.entity.attribute", "Attribute", "AttributeImpl", "Attributes");
```
TheMeinerLP, line 52: Is there another solution?
I due to recent updates we need to look at a hybrid solution here

TheMeinerLP (changes requested):

Issue #2481, opened 2024-11-16
Incorrect Serialization of HoverEvent ShowItem


GreatWyrm:
There's two issues with using HoverEvent showItem in Minestom.

The first issue is that the BinaryTag serializes the whole item, not just the components (pictures below), meaning no components are stored in the Adventure ShowItem object.
_The code that serializes the item. Notice that the whole item NBT is displayed._
![Screenshot from 2024-11-15 23-12-36](https://github.com/user-attachments/assets/5f7d566b-651b-4845-a2dc-82096fd6d5ed)
_The Adventure ShowItem object. Notice there are no dataComponents attached._
![Screenshot from 2024-11-15 23-12-55](https://github.com/user-attachments/assets/8feb0e2d-9ce4-4aa2-97fd-0bf0b98f1098)

The second issue is that the ShowItem object is not written in the Component Serializer, due to it being unimplemented.
https://github.com/Minestom/Minestom/blob/bb7acc2e774944fb77f8b9d0dd4b23728408a920/src/main/java/net/minestom/server/network/ComponentNetworkBufferTypeImpl.java#L247-L250
A minor note, but [wiki.vg](https://wiki.vg/Text_formatting#Text_components) lists the part as *tags*, rather than *components*.

GreatWyrm:
Closed, this works as of the 1.21.5 merge

PR #2501, opened 2024-12-01
Serialize ItemStacks in system chat

### LINKED ISSUE #2227: Bug: Item hover event is not working
The current implementation of `NbtComponentSerializer` is using legacy NBT tags instead of the newer data components in the item hover event. As a result, the hover event is not functioning correctly.

GreatWyrm:
Currently, ItemStacks are not correctly serialized when using asHoverEvent().
This is due to various changes, but the important one is the switch to item components.

This PR implements this functionality.

Closes #2227, #2481.


src/main/java/net/minestom/server/network/ComponentNetworkBufferTypeImpl.java
```diff
@@ -247,7 +256,35 @@ private void writeHoverEvent(@NotNull NetworkBuffer buffer, @NotNull HoverEvent<
 
             buffer.write(BYTE, TAG_COMPOUND);
             writeUtf(buffer, "components");
-            //todo item components
+            BinaryTagWriter nbtWriter = impl(buffer).nbtWriter;
```
mworzala, line 259: this is duplicated from NbtTypeImpl, this logic should just be moved into NetworkBufferTypeImpl, eg `impl(buffer).nbtWriter()`, same for reader.

src/main/java/net/minestom/server/component/DataComponentMapImpl.java
```diff
@@ -68,6 +70,26 @@ public boolean has(@NotNull DataComponentMap prototype, @NotNull DataComponent<?
         return new DataComponentMapImpl(newComponents);
     }
 
+    @Override
+    public void forEach(@NotNull BiConsumer<DataComponent<?>, Object> consumer) {
```
mworzala, line 74: These are not good and need to be resolved before this is merged.

Some options (none of which I am not in love with currently):
* Change the consumer to return ints and let the caller handle it
* track the ID->DataComponent mapping internally

I think tracking the id mapping is the most obvious option, but I am not that excited about increasing the memory footprint of ItemStack :(

src/main/java/net/minestom/server/network/ComponentNetworkBufferTypeImpl.java
```diff
@@ -247,7 +254,28 @@ private void writeHoverEvent(@NotNull NetworkBuffer buffer, @NotNull HoverEvent<
 
             buffer.write(BYTE, TAG_COMPOUND);
             writeUtf(buffer, "components");
-            //todo item components
+            BinaryTagWriter nbtWriter = impl(buffer).nbtWriter();
+            BinaryTagSerializer.Context context = impl(buffer).registries() == null ? BinaryTagSerializer.Context.EMPTY : new BinaryTagSerializer.ContextWithRegistries(impl(buffer).registries(), true);
+            for (var entry : value.dataComponents().entrySet()) {
+                if (entry.getValue() instanceof MinestomDataComponentValue minestomDataComponentValue && minestomDataComponentValue.component().isSerialized()) {
+                    // If it's a removed value
+                    if (minestomDataComponentValue.value().equals(DataComponentValue.removed())) {
```
mworzala, line 262: this is wrong, it should be a check on `entry.getValue()`, not the value of our MinestomDataComponentValue which only holds added components.

src/main/java/net/minestom/server/adventure/provider/NBTLegacyHoverEventSerializer.java
```diff
@@ -15,6 +15,7 @@
 import java.util.Objects;
 import java.util.UUID;
 
+@Deprecated
```
TheMeinerLP, line 18: Can you add some reason why is deprecated as a javadoc comment ?

src/main/java/net/minestom/server/component/DataComponentMapImpl.java
```diff
@@ -68,6 +70,26 @@ public boolean has(@NotNull DataComponentMap prototype, @NotNull DataComponent<?
...
+            DataComponent<?> component = ItemComponent.fromId(integer);
+            if (component != null) {
+                consumer.accept(component, object);
+            }
+        });
+    }
+
+    @Override
+    public void patchForEach(@NotNull BiConsumer<DataComponent<?>, @Nullable Object> consumer) {
+        // TODO: This only works for itemComponent, fix
```
TheMeinerLP, line 86: Can you remove or open a issue for the todo ?

src/main/java/net/minestom/server/network/NetworkBufferImpl.java
```diff
@@ -10,8 +10,7 @@
 
 import javax.crypto.Cipher;
 import javax.crypto.ShortBufferException;
-import java.io.EOFException;
-import java.io.IOException;
+import java.io.*;
```
TheMeinerLP, line 13: Please prevent * imports

src/main/java/net/minestom/server/network/NetworkBufferImpl.java
```diff
@@ -555,6 +554,35 @@ static NetworkBufferImpl impl(NetworkBuffer buffer) {
...
+                    return NetworkBufferImpl.this.read(BYTE) & 0xFF;
+                }
+
+                @Override
+                public int available() {
+                    return (int) NetworkBufferImpl.this.readableBytes();
+                }
+            }));
+        }
+        return nbtReader;
```
TheMeinerLP, line 583: Looks dirty

src/main/java/net/minestom/server/network/NetworkBufferImpl.java
```diff
@@ -555,6 +554,35 @@ static NetworkBufferImpl impl(NetworkBuffer buffer) {
...
 
+    BinaryTagWriter nbtWriter() {
+        if (this.nbtWriter == null) {
+            this.nbtWriter = new BinaryTagWriter(new DataOutputStream(new OutputStream() {
+                @Override
+                public void write(int b) {
+                    NetworkBufferImpl.this.write(BYTE, (byte) b);
+                }
+            }));
+        }
```
TheMeinerLP, line 565: Looks dirty

src/test/java/net/minestom/server/adventure/TestCanCreateHoverEvent.java
```diff
@@ -0,0 +1,25 @@
...
+import net.kyori.adventure.text.event.HoverEvent;
+import net.kyori.adventure.text.format.NamedTextColor;
+import net.minestom.server.item.ItemComponent;
+import net.minestom.server.item.ItemStack;
+import net.minestom.server.item.Material;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class TestCanCreateHoverEvent {
```
TheMeinerLP, line 14: remove public

src/test/java/net/minestom/server/adventure/TestCanCreateHoverEvent.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test
+    public void createHoverEvent() {
```
TheMeinerLP, line 17: Also here remove public

TheMeinerLP (changes requested):

src/main/java/net/minestom/server/network/NetworkBufferImpl.java
GreatWyrm: Feels unnecessary (IntelliJ auto-groups imports into * if you have too many), but changed

src/main/java/net/minestom/server/component/DataComponentMapImpl.java
GreatWyrm: See https://github.com/Minestom/Minestom/pull/2501#discussion_r1895745862, this won't be merged without some solution being figured out.

src/main/java/net/minestom/server/network/NetworkBufferImpl.java
GreatWyrm: Non-actionable feedback, it's the best way to my knowledge of doing this.

src/main/java/net/minestom/server/network/NetworkBufferImpl.java
GreatWyrm: Non-actionable feedback

TheMeinerLP: So I could imagine that you make a package private input and outstream or subclass

src/main/java/net/minestom/server/component/DataComponentMapImpl.java
TheMeinerLP: Then we should consider it a “draft” until then

mworzala:
Superseded/expanded on by https://github.com/Minestom/Minestom/commit/514656a504b3e5bdecdb06c77f5c7b90b5b6b6f0 on 1.21.5

PR #2502, opened 2024-12-03
fix: client disconnect, server-side exceptions when changing entity type of a Player

Steanky:
Basically an extension of https://github.com/Minestom/Minestom/pull/2495, that also integrates https://github.com/Minestom/Minestom/pull/2421. Should fix crashes & some server-side errors when players are "disguised" as another entity.

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -12,6 +12,7 @@ import net.minestom.server.entity.damage.Damage;
 import net.minestom.server.entity.damage.DamageType;
 import net.minestom.server.entity.metadata.EntityMeta;
 import net.minestom.server.entity.metadata.LivingEntityMeta;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -344,14 +345,15 @@ public boolean damage(@NotNull Damage damage) {
...
+
                 if (additionalHearts > 0) {
                     if (remainingDamage > additionalHearts) {
                         remainingDamage -= additionalHearts;
-                        player.setAdditionalHearts(0);
+                        playerMeta.setAdditionalHearts(0);
                     } else {
-                        player.setAdditionalHearts(additionalHearts - remainingDamage);
+                        playerMeta.setAdditionalHearts(additionalHearts - remainingDamage);
                         remainingDamage = 0;
```
TheMeinerLP, line 357: Doesn't the code generally belong in the player?

TheMeinerLP (changes requested):

PR #2503, opened 2024-12-03
fix: "off-by-one" error in Suggestion

Steanky:
Previously, `Suggestion#getStart` would report an index that was off-by-one. For example, given a command completable to `/example command`, a client typing `/example com` would result in a Suggestion with an input string `example com`, and a starting index of 9 (instead of 8). 

This PR fixes this surprising behavior, while still ensuring the client gets the value it expects by offsetting the `index` field of the `TabCompletePacket` sent by the server. It also adds additional documentation and bounds checks to Suggestion.


TheMeinerLP (changes requested):
Can you rebase it ?

Issue #2505, opened 2024-12-03
UNSAFE_itemInsert in PlayerInventory syncs stale equipment data


GreatWyrm:
The tests `PlayerInventoryIntegrationTest#equipmentViewTest` and `PlayerInventoryIntegrationTest#heldItemViewTest` are currently failing.

When `playerArmored.setEquipment(EquipmentSlot.HELMET, MAGIC_STACK);` is called, the PlayerInventory attempts to update its internal itemstack array and send out the new equipment data.

`player.syncEquipment()` is called before `super.UNSAFE_itemInsert(slot, item, previous, sendPacket);`, and the syncEquipment method reads the stale data before the itemInsert method is called to update it.

Issue #2508, opened 2024-12-05
Incorrect Enum Order for EquipmentSlots


einfachBlu:
```
ItemStackImpl[material=minecraft:iron_helmet, amount=1, components=DataComponentMapImpl[components={13=>AttributeList[modifiers=[], showInTooltip=false], 15=>net.minestom.server.utils.Unit@30bbf458, 28=>Equippable[slot=CHESTPLATE, equipSound=minecraft:item.armor.equip_chain, model=minecraft:iron, cameraOverlay=null, allowedEntities=null, dispensable=false, swappable=false, damageOnHurt=false]}]]
```

To make the item display correctly, you have to make this iron_helmet at the EquipmentSlot "CHESTPLATE"

This is the Minestom Enum Order:
```
MAIN_HAND(false, -1, "mainhand"),
OFF_HAND(false, -1, "offhand"),
BOOTS(true, BOOTS_SLOT, "feet"),
LEGGINGS(true, LEGGINGS_SLOT, "legs"),
CHESTPLATE(true, CHESTPLATE_SLOT, "chest"),
HELMET(true, HELMET_SLOT, "head"),
BODY(false, -1, "body");
```

https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Pre-release_protocol
<img width="698" alt="image" src="https://github.com/user-attachments/assets/11d8ea62-396d-41b7-8afe-ee3b17ea3c2f">

mworzala:
Fixed in #2509

PR #2509, opened 2024-12-05
fix: updating equipment would send the client the wrong item

Steanky:
Fixes https://github.com/Minestom/Minestom/issues/2505, https://github.com/Minestom/Minestom/issues/2508.

The client expects `EntityEquipmentPacket` to have the old ordering of `EquipmentSlot`, but expects the *new* ordering in components like `Equippable`. This seems weird and Mojang might change it later, but for now this fixes the armor bug.


Steanky:
Also added a fix for https://github.com/Minestom/Minestom/issues/2508

mworzala:
I wanted to keep the enum ordering the same, technically it's a breaking change to change it. Also I changed the naming and explanation a little.

PR #2511, opened 2024-12-05
Add helpers for ITEM_NAME component

leon0017:


mworzala:
Item name is likely not what you should be using this (custom name is probably the correct one). I do not want to set a precedent of adding helpers for every component.

Closing for now, happy to discuss further however

PR #2512, opened 2024-12-05
Call InventoryCloseEvent when Player closeInventory is called

leon0017:


leon0017:
Fixed no check if the open inventory is null first

mworzala (approved):
Nice catch, thank you!

PR #2517, opened 2024-12-09
Change Additional Hearts default to 0 instead of 1

GreatWyrm:
Before the switch to use MetadataDef instead of Metadata (done in commit f4b32eddcf02174f842cb2ac5e1ccd6ed5815e5b), the default value for the ADDITIONAL_HEARTS field in PlayerMeta was 0 (indicated by [this line](https://github.com/Minestom/Minestom/commit/f4b32eddcf02174f842cb2ac5e1ccd6ed5815e5b#diff-3f119881a02c6face20cde30fb568adf048b1a503b38e208feab14754834def2L28), where the default value supplied is 0F).

Now it's 1f by default, which causes some confusion when applying damage to players, as the first time they are damaged it will be 1 less due to the additional heart field being set to 1.
This PR fixes that by changing the default to 0.

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataDef.java
+++ b/src/main/java/net/minestom/server/entity/MetadataDef.java
@@ -211,7 +211,7 @@ public sealed class MetadataDef {
     }
 
     public static final class Player extends LivingEntity {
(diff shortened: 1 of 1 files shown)


mworzala (approved):
Yeah the default should for sure be zero, thanks. We really do need a way to generate this file or validate it in some way against vanilla data...

PR #2518, opened 2024-12-09
Make MinecraftServer static

ItsFelix5:
MinecraftServer is fully static internally but you still need a variable to start the server, this is kind of annoying and would be better as a static function.
This PR also adds a start() that automatically chooses the right local address/port.

### DIFF
--- a/demo/src/main/java/net/minestom/demo/Main.java
+++ b/demo/src/main/java/net/minestom/demo/Main.java
@@ -34,7 +34,7 @@ public class Main {
 
         MinecraftServer.setCompressionThreshold(0);
 
-        MinecraftServer minecraftServer = MinecraftServer.init();
+        MinecraftServer.init();
 
         BlockManager blockManager = MinecraftServer.getBlockManager();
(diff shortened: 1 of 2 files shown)


mworzala:
There is not necessarily a "right" address/port for the server to listen on, I believe making people choose is correct.

While its true that MinecraftServer having an instance to call start is unnecessary, it is also a breaking change to every single Minestom project in existence, so does not seem worth changing at this time.

DasBabyPixel:
I just wanted to add to this, that while it is true that it is a breaking change, the longer the change (which to me seems has to happen at some point) is put aside, the worse it will be to deal with.
My opinion/suggestion is to make the change to static, but keep the previous behaviour (and signatures) intact, so it is not breaking. Then mark the old start for removal, and remove it at some point in the future

mworzala:
I do generally agree with that comment, but I also dont think this really has to change. I would prefer to not invalidate all existing content (in addition to codebases) unless there is a more significant improvement in addition.

PR #2520, opened 2024-12-10
Fix boat metadata, and prevent packet spam when riding a boat

### LINKED ISSUE #2519: Can't add passengers to boat entities
I can't add passengers to boats I'm on, commit `d760a60a5c` I made a minimal repo
```
public class Main {
    public static void main(final String[] args) {
        final MinecraftServer server = MinecraftServer.init();

        final InstanceContainer instance = MinecraftServer.getInstanceManager().createInstanceContainer();
        instance.setGenerator(unit -> unit.modifier().fillHeight(0, 60, Block.LODESTONE));
        instance.setTimeRate(0);
        instance.setTime(6000);

        final GlobalEventHandler handler = MinecraftServer.getGlobalEventHandler();
        handler.addListener(PlayerSpawnEvent.class, event -> {
            final Player player = event.getPlayer();
            player.setGameMode(GameMode.ADVENTURE);
            player.setFlying(true);
            player.setAllowFlying(true);

            final Entity entity = new Entity(EntityType.OAK_BOAT);
            entity.setInstance(instance, new Pos(0, 65, 10)).join();

            player.scheduler().scheduleTask(() -> {
                entity.addPassenger(player);
            }, TaskSchedule.seconds(5), TaskSchedule.stop());
        });
        handler.addListener(AsyncPlayerConfigurationEvent.class, event -> {
            event.getPlayer().setRespawnPoint(new Pos(0, 65, 0));
        });

    }
}
```

Crash report:
[disconnect-2024-12-09_17.32.43-client.txt](https://github.com/user-attachments/files/18066148/disconnect-2024-12-09_17.32.43-client.txt)

GreatWyrm:
Fixes #2519.
Boats no longer have a type field, causing a Network Protocol Error when trying to ride them (which causes their metadata to be sent)
Additionally, I found that boats sent their metadata packet every tick due to the client sending the ClientBoatSteerPacket constantly, so I have changed that method to only send the metadata when it changes.

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataDef.java
+++ b/src/main/java/net/minestom/server/entity/MetadataDef.java
@@ -141,10 +141,9 @@ public sealed class MetadataDef {
     }
 
     public static final class Boat extends AbstractVehicle {
-        public static final Entry<Integer> TYPE = index(0, Metadata::VarInt, 0);
-        public static final Entry<Boolean> IS_LEFT_PADDLE_TURNING = index(1, Metadata::Boolean, false);
-        public static final Entry<Boolean> IS_RIGHT_PADDLE_TURNING = index(2, Metadata::Boolean, false);
-        public static final Entry<Integer> SPLASH_TIMER = index(3, Metadata::VarInt, 0);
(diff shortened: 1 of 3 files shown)


mworzala (approved):
Nice fix, thanks

PR #2522, opened 2024-12-13
Fix re-registering vanilla registry objects not working properly

Aeltumn:
Encountered this issue while working on updating to 1.21.3 as my client was crashing from protocol errors constantly.

We edit the main plains biome to change its properties by re-registering it, and the old implementation would cause the biome to be included in the registry twice, causing a protocol error on modern clients.

This PR fixes it by updating the logic of the register method to properly replace the data instead of appending only.


mworzala (approved):
Thanks for the fix!

PR #2523, opened 2024-12-13
Fix shift clicking in last hotbar slot

GoldenStack:
Shift clicking into the last hotbar slot doesn't work because the ending slot is marked as 8, but the ending slot is exclusive. Increasing this from 8 to 9 fixes it.

This fixes blobfish's issue (mentioned on Discord).


mworzala (approved):
awkward

PR #2524, opened 2024-12-13
Remove legacy protocol IDs for equipment slots

### LINKED ISSUE #2515: Entities other than player have incorrect armor slots set
On the 1.21.4 branch

To reproduce just try set the entities boots

GoldenStack:
Removes legacy protocol IDs for equipment slots.
Their existence was causing issues, since apparently the client started using normal equipment slot IDs.

Fixes #2515.


mworzala:
These were just added so I am suspicious, more investigation needed

mworzala:
They were flipped, but both are still needed.

PR #2529, opened 2024-12-17
Fix EventNode bug where listeners aren't firing

DasBabyPixel:
My listeners weren't firing, so I investigated. Came up with the following test and fix for the bug. 
Didn't know what to name the HashMap, not happy with the name.

### DIFF
--- a/src/main/java/net/minestom/server/event/EventNodeImpl.java
+++ b/src/main/java/net/minestom/server/event/EventNodeImpl.java
@@ -22,6 +22,7 @@ non-sealed class EventNodeImpl<T extends Event> implements EventNode<T> {
     static final Object GLOBAL_CHILD_LOCK = new Object();
 
     private final Map<Class, Handle<T>> handleMap = new ConcurrentHashMap<>();
+    private final Map<Class<?>, CopyOnWriteArraySet<Class<?>>> recursiveSuperclasses = new ConcurrentHashMap<>();
     final Map<Class<? extends T>, ListenerEntry<T>> listenerMap = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


DasBabyPixel:
> Honestly I havent messed with the event node implementation a ton, but my understanding of the problem is the following (please correct me):
> 
> When listening to `Recursive1` we get a `Handle` for `Recursive1`, but when calling `Recursive2` we create a handle for `Recursive2` (which also encompasses the listeners for `Recursive1`. However when removing the original node we only walk up the inheritance chain from that nodes listeners, aka just `Recursive1`, meaning that the handle for `Recursive2` was not invalidated and remains calling new events.
> 
> Assuming that is all correct, it seems plausible to just check each listener to see if it inherits from the recursive type being removed and invalidate it in that case rather than tracking this all separately.

Your understanding is correct. 
Looping through all listeners is also an option, I decided to track the data to avoid looping through all the listeners.
I remember spigot's `HandlerList` unregistered by looping, and I saw somewhere in the forums that they said that was a mistake.
That being said, if I should implement this by looping, I'm happy to do so.

mworzala:
Hmm I can't immediately think of why it would be a mistake, if you can find the mention I am curious about it. Until then I like the idea of doing it with the existing structures better so we dont have another thing to keep track of. We can always change it later also this is all internal api.

DasBabyPixel:
image
Ehh can't find it, mentionend in papermc guide so maybe I'm wrong and confusing this with sth else. I'm gonna implement this with looping through all registered listeners later, can't find the time rn with family here

DasBabyPixel:
Kinda forgot this, should be ready for review

mworzala (approved):
Looks good, thanks as always :)

PR #2534, opened 2024-12-21
feat: Optional instance based threading

leon0017:
Added a boolean server flag `minestom.instance-based-dispatcher` to enable instance based threading.

I'm not really sure if this goes against how Minestom should be threaded, and if not I will keep a fork for this, but I was porting a minigame from Paper and keeping each instance single threaded was a much nicer option for me.

With the flag set, `minestom.dispatcher-threads` are still the total amount of dispatcher threads and are spread efficiently across instances (uses a new `leastOccupiedThread` ThreadProvider, which finds the thread with the least amount of partitions)

Partitions become instances instead of chunks being the partitions, and chunks and entities are the elements of the partitions.

Here's what I used to test it in the demo (there is instance unregistering testing in PlayerInit as well):

https://github.com/user-attachments/assets/ec28201e-387d-4bc3-9931-428eb4db24b7

```java
// Main.java
System.setProperty("minestom.instance-based-dispatcher", "true");
System.setProperty("minestom.dispatcher-threads", "3");
```

```java
// PlayerInit.java
private static InstanceContainer makeInstanceContainer() {
    InstanceContainer instanceContainer = MinecraftServer.getInstanceManager().createInstanceContainer();
    instanceContainer.setGenerator(unit -> {
        unit.modifier().fillHeight(0, 40, Block.STONE);
    
        if (unit.absoluteStart().blockY() < 40 && unit.absoluteEnd().blockY() > 40) {
            unit.modifier().setBlock(unit.absoluteStart().blockX(), 40, unit.absoluteStart().blockZ(), Block.TORCH);
        }
    });
    instanceContainer.setChunkSupplier(LightingChunk::new);
    instanceContainer.setTimeRate(0);
    instanceContainer.setTime(ThreadLocalRandom.current().nextInt(20000));
    return instanceContainer;
}

{
    InstanceManager instanceManager = MinecraftServer.getInstanceManager();
    
    for (int i = 0; i < 16; i++) {
        InstanceContainer instanceContainer = makeInstanceContainer();
        if (ThreadLocalRandom.current().nextDouble() < 0.75) // Unregister most instances
            instanceManager.unregisterInstance(instanceContainer);
    }
    
    for (int i = 0; i < 8; i++) {
        makeInstanceContainer();
    }
}
```

```java
// InstanceCommand.java
public class InstanceCommand extends Command {

    public InstanceCommand() {
        super("instance");

        ArgumentInteger argumentInteger = ArgumentType.Integer("index");

            Player player = (Player) sender;
            List<Instance> instances = MinecraftServer.getInstanceManager().getInstances().stream().toList();
            player.setInstance(instances.get(context.get(argumentInteger)), new Pos(0, 41, 0)).thenRun(()
                    -> player.scheduler().scheduleNextTick(() -> player.sendMessage("Your new ticking thread: " + player.acquirable().assignedThread().getName())));
        }, argumentInteger);
    }

}
```

### DIFF
--- a/src/main/java/net/minestom/server/ServerProcessImpl.java
+++ b/src/main/java/net/minestom/server/ServerProcessImpl.java
@@ -13,7 +13,6 @@ import net.minestom.server.event.GlobalEventHandler;
 import net.minestom.server.event.server.ServerTickMonitorEvent;
 import net.minestom.server.exception.ExceptionManager;
 import net.minestom.server.gamedata.tags.TagManager;
-import net.minestom.server.instance.Chunk;
 import net.minestom.server.instance.Instance;
 import net.minestom.server.instance.InstanceManager;
 import net.minestom.server.instance.block.BlockManager;
@@ -37,7 +36,6 @@ import net.minestom.server.scoreboard.TeamManager;
 import net.minestom.server.snapshot.*;
 import net.minestom.server.thread.Acquirable;
 import net.minestom.server.thread.ThreadDispatcher;
-import net.minestom.server.thread.ThreadProvider;
 import net.minestom.server.timer.SchedulerManager;
 import net.minestom.server.utils.PacketViewableUtils;
 import net.minestom.server.utils.collection.MappedCollection;
@@ -94,7 +92,7 @@ final class ServerProcessImpl implements ServerProcess {
 
     private final Server server;
 
-    private final ThreadDispatcher<Chunk> dispatcher;
+    private final ThreadDispatcher<? extends Tickable> dispatcher;
     private final Ticker ticker;
 
     private final AtomicBoolean started = new AtomicBoolean();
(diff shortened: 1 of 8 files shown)


leon0017:
Added `runChunkPartition` and `runInstancePartition` consumers, instead of doing ifs and `as(container)` conversions outside of ThreadDispatcher

src/main/java/net/minestom/server/ServerProcessImpl.java
```diff
@@ -139,7 +137,7 @@ public ServerProcessImpl() {
 
         this.server = new Server(packetParser);
 
-        this.dispatcher = ThreadDispatcher.of(ThreadProvider.counter(), ServerFlag.DISPATCHER_THREADS);
+        this.dispatcher = ThreadDispatcher.serverProcess();
```
mworzala, line 140: Not a fan of this name, probably the logic of choosing from inside #serverProcess should just be here instead

mworzala (commented):
While I do want to support instance grouping, I am not a big fan of this implementation of it. When improving ThreadDispatcher I would prefer generally supporting people slotting in their own dispatchers as opposed to (more or less) including both chunk and instance ones and checking at every relevant point

leon0017:
Does that mean that thread dispatchers would have handlers for things like entity instance setting and entity removal, instance registration and unregistration, and chunk loading and unloading?

mworzala:
I think yes, but I have not sat down to come up with an API. Happy to discuss further in Discord before you go further :)

PR #2536, opened 2024-12-25
refactor: add getter for `lastUpdate` in Cooldown

jejebecarte:


mworzala:
Cooldown is intended to be removed in the not-too-distant future because it uses wall clock time instead of tick time which is typically desirable for a Minecraft server. I'm not sure its great to add new api to something on the way out.

jejebecarte:
No worries then. Would you instead recommend using `TaskScheduler#submitTask` or `TaskShedular#scheduleTask` to achieve this then?

mworzala:
Kinda depends what you want to accomplish. If you do just need a cooldown then it could be as simple as recording the entity or world age (in ticks). A task is also fine!

PR #2541, opened 2024-12-27
fix: suggestion callback not sending

### LINKED ISSUE #2539: Command Parser is causing suggestion callback problems
If there is a command with a default executor and two args, and the 1st arg is an argument word with a suggestion callback, the suggestion callback isn't called. If the default executor is removed, the problem is fixed.

https://pastebin.com/RznvhzYf - player arg completions work
https://pastebin.com/FvsdseMy - player arg completions don't work

image
image

hapily04:
Fixes #2539


hapily04:
My fix seems clunky and I'm only 75% sure on how I got there, but it works so I'm marking this as ready to review. Feedback is appreciated and if possible I would love to get rid of the stream call related to preventing two different results for the same key.

TheMeinerLP (approved):
LGTM, proofed locally

PR #2571, opened 2024-12-30
fix: slight (item)entity desync

hapily04:
The `updateVelocity` method called in `PhysicsUtils` is using the old position instead of the new position, causing a slight desync.
`synchronizePosition` is removed since it's done just before the onGround check in `super.movementTick()`
`EntityType.ITEM` being in the `SYNCHRONIZE_ONLY_ENTITIES` set also contributes to the desync, so it was removed.

Before (the jump at the end once it touches the ground):
https://github.com/user-attachments/assets/1805ace7-4130-48cb-a194-03c1db064fa5

After:
https://github.com/user-attachments/assets/2a1b6053-be0d-4458-a8c7-f521eb692d9d

### DIFF
--- a/src/main/java/net/minestom/server/instance/DynamicChunk.java
+++ b/src/main/java/net/minestom/server/instance/DynamicChunk.java
@@ -211,7 +211,7 @@ public class DynamicChunk extends Chunk {
         final Section section = getSectionAt(y);
         final int blockStateId = section.blockPalette()
                 .get(globalToSectionRelative(x), globalToSectionRelative(y), globalToSectionRelative(z));
-        return Objects.requireNonNullElse(Block.fromStateId((short) blockStateId), Block.AIR);
+        return Objects.requireNonNullElse(Block.fromStateId(blockStateId), Block.AIR);
     }
 
     @Override
--- a/src/main/java/net/minestom/server/network/packet/server/play/PlayerInfoUpdatePacket.java
+++ b/src/main/java/net/minestom/server/network/packet/server/play/PlayerInfoUpdatePacket.java
@@ -10,16 +10,15 @@ import net.minestom.server.network.player.GameProfile;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.EnumSet;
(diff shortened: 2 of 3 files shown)


hapily04:
I thought this changed fix it, I think when I made the new branch some changes were lost. Converting to draft until I find those changes

hapily04:
Should be good to go now

src/main/java/net/minestom/server/collision/PhysicsUtils.java
```diff
@@ -39,7 +39,7 @@ public final class PhysicsUtils {
         Vec newVelocity = physicsResult.newVelocity();
 
         Pos positionWithinBorder = CollisionUtils.applyWorldBorder(worldBorder, entityPosition, newPosition);
-        newVelocity = updateVelocity(entityPosition, newVelocity, blockGetter, aerodynamics, !positionWithinBorder.samePoint(entityPosition), entityFlying, entityOnGround, entityNoGravity);
+        newVelocity = updateVelocity(newPosition, newVelocity, blockGetter, aerodynamics, !positionWithinBorder.samePoint(entityPosition), entityFlying, entityOnGround, entityNoGravity);
```
KermanIsPretty, line 42: Shouldnt this be positionWithinBorder instead of newPosition?

hapily04: Yes. Will update in the next commit

hapily04:
Update

The code I initially tested was dirty and I thought it would be an easy fix to swap out what was causing the problem with cleaner code. I struggled to find the difference between what was fixing it in the dirty code vs the clean broken code. I intend on finishing this pull request still, I've just been busy with other code.

TheMeinerLP:
I would first do a cleanup with “breaking” changes to 1.21.4 as an extra PR and develop the fix if necessary. If you are in a position to do so

hapily04:
> I would first do a cleanup with “breaking” changes to 1.21.4 as an extra PR and develop the fix if necessary. If you are in a position to do so

https://github.com/Minestom/Minestom/pull/2633

Issue #2353, opened 2024-08-24
BlockHandler#onTouch is firing inconsistently


Soupertonic:
The `onTouch` for `BlockHandler` implementations is fired inconsistently for many blocks. Some blocks fire `onTouch` for all block faces, some fire for all except the top block face and some only fire for west, north and bottom block faces.

### Minestom
Commit: a521c4e7cd
Repository: Maven Central
Extensions / Add-ons / Plugins: None

### Client
Mods: None
Resource Packs: None

### Details
Campfire Block: Triggers `onTouch` for all faces
Dragon Egg Block: Triggers `onTouch` for all faces except `up`
Gold Block: Triggers `onTouch` for faces `west`, `north` and `bottom`

### Demonstration
https://youtu.be/r4YtKXzS_Ko

### How to reproduce
```java
public class TestServer {
  public static class TestBlockHandler implements BlockHandler {
    public static final BlockHandler INSTANCE = new TestBlockHandler();

    @Override
    public void onTouch(@NotNull Touch touch) {
      touch.getInstance().sendMessage(
        Component.text("touched").color(
          TextColor.color(
            (int) (Math.random() * 255),
            (int) (Math.random() * 255),
            (int) (Math.random() * 255)
          )
        )
      );
    }

    @Override
    public @NotNull NamespaceID getNamespaceId() {
      return NamespaceID.from("test", "test");
    }
  }

  public static void main(String[] args) {
    MinecraftServer minecraftServer = MinecraftServer.init();

    InstanceContainer instanceContainer = MinecraftServer.getInstanceManager().createInstanceContainer(DimensionType.OVERWORLD);
    instanceContainer.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.SANDSTONE));
    instanceContainer.setChunkSupplier(LightingChunk::new);

    BlockManager blockManager = MinecraftServer.getBlockManager();
    blockManager.registerHandler(TestBlockHandler.INSTANCE.getNamespaceId(), () -> TestBlockHandler.INSTANCE);

    GlobalEventHandler globalEventHandler = MinecraftServer.getGlobalEventHandler();
    globalEventHandler.addListener(AsyncPlayerConfigurationEvent.class, event -> {
      event.setSpawningInstance(instanceContainer);
      event.getPlayer().setRespawnPoint(new Pos(0, 40, 0));
    });

    globalEventHandler.addListener(PlayerChatEvent.class, event -> {
[...]

PR #2575, opened 2024-12-31
Reworked BlockHandler's

KermanIsPretty:
Also fixes an issue with onTouch not working.

This PR is still accepting suggestions to make it better.

### DIFF
--- a/src/main/java/net/minestom/server/collision/CollisionUtils.java
+++ b/src/main/java/net/minestom/server/collision/CollisionUtils.java
@@ -33,7 +33,7 @@ public final class CollisionUtils {
      * @return the result of physics simulation
      */
     public static PhysicsResult handlePhysics(@NotNull Entity entity, @NotNull Vec entityVelocity,
-                                              @Nullable PhysicsResult lastPhysicsResult, boolean singleCollision) {
+                                                  @Nullable PhysicsResult lastPhysicsResult, boolean singleCollision) {
         final Instance instance = entity.getInstance();
         assert instance != null;
         return handlePhysics(instance, entity.getChunk(),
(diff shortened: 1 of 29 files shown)


KermanIsPretty:
Possibly Fixes #2353 At least on gold blocks. I dont like how the fix was implemented as we already do some physics in the movement tick but they dont return collisions on X and Z so its kind of hard to implement that here.

src/main/java/net/minestom/server/collision/CollisionUtils.java
```diff
@@ -118,9 +118,9 @@ public static PhysicsResult handlePhysics(@NotNull Instance instance, @Nullable
      */
     @ApiStatus.Internal
     public static PhysicsResult handlePhysics(@NotNull Block.Getter blockGetter,
-                                              @NotNull BoundingBox boundingBox,
-                                              @NotNull Pos position, @NotNull Vec velocity,
-                                              @Nullable PhysicsResult lastPhysicsResult, boolean singleCollision) {
+                                                  @NotNull BoundingBox boundingBox,
```
mworzala, line 121: random formatting changes should not be here

mworzala (commented):
Overall I think that the name changes and record specialization of this PR should not happen. The BlockHandler name changes are breaking and so at minimum need to be targeting a new version branch. But honestly I feel they are unnecessary at all.

Regarding the PhysicsResult and BlockHandler.* specializations, all of those classes are simple data holders, there is no need to have a split interfaces and implementations, and no need for multiple implementations of PhysicsResult it can simply be overloaded constructors.

I would like to re review the PR without the name/impl changes.

KermanIsPretty:
Thank you for you feedback, I am going to split the physics and block handler changes into separate pr's and remove any naming changes. I will go ahead and rebase on 1.12.4 branch as the way onTouch fires has changed and doesn't continually fire.

Issue #2470, opened 2024-11-06
Interaction distance limit does not account for hitbox size and eye position


CheaterTim:
When interacting with an entity, minestom measures the distance between the player's position and the entity's position to cancel any interactions more than 1 block outside of the player's range. This however, does not take into account the size of the entity that is being interacted with. You only need your range to suffice until you reach the hitbox of the entity.

![illustration](https://github.com/user-attachments/assets/85955b5e-c783-4908-b5c5-ef0fcea47415)

https://github.com/Minestom/Minestom/blob/f71ab6d851fc30a54bbdc9ef6864f6626dea3121/src/main/java/net/minestom/server/listener/UseEntityListener.java#L22-L27

## Other flaw
This measures the distance between the player's bottom center and the entity's bottom center. In reality, the raycast is shot from the height of the player's eyes.

## Reproduction
Spawns a slime of size 3 and positions the player so that it is just barely looking at it, with enforce-entity-interaction-range enabled, we will see that we are receiving interact packets but the interact event isn't being fired, with it disabled, we will see that the event is firing as it should.

```java
    public static void main(String[] args) {
        //System.setProperty("minestom.enforce-entity-interaction-range", "false");
        MinecraftServer minecraftServer = MinecraftServer.init();

        InstanceManager instanceManager = MinecraftServer.getInstanceManager();
        InstanceContainer instanceContainer = instanceManager.createInstanceContainer();
        instanceContainer.setChunkSupplier(LightingChunk::new);

        instanceContainer.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.GRASS_BLOCK));

        GlobalEventHandler globalEventHandler = MinecraftServer.getGlobalEventHandler();
        globalEventHandler.addListener(AsyncPlayerConfigurationEvent.class, event -> {
            final Player player = event.getPlayer();
            event.setSpawningInstance(instanceContainer);
            player.setRespawnPoint(new Pos(2, 40, 2.3, -49, 11));
        });

        var slime = new LivingEntity(EntityType.SLIME);
        var meta = (SlimeMeta)slime.getEntityMeta();
        meta.setSize(3);

        slime.setInstance(instanceContainer, new Pos(5, 40, 5));

        globalEventHandler.addListener(PlayerEntityInteractEvent.class, event -> {
            event.getPlayer().sendMessage("Interacted");
        });
        globalEventHandler.addListener(PlayerPacketEvent.class, event -> {
[...]

PR #2576, opened 2025-01-01
Fix to issue #2470

Nex0662:
Calculates the squared distance from the player's eye position to the closest point on the entity's bounding box, this also improves the accuracy of interaction range enforcement (preventing interactions w/ entities outside the allowed range regardless of bounding box size)

This PR attempts to fix https://github.com/Minestom/Minestom/issues/2470


mworzala:
This should have some tests with it, then I am good with it

mworzala (approved):
Looks great, thanks!

PR #2578, opened 2025-01-01
Add a helper for Properties on Blocks...

RealMangorage:
This makes it possible to have a nice Helper on Block.withProperty

```
Block.TNT.withProperty(FacingProperty.INSTANCE, Direction.WEST);

vs

Block.TNT.withProperty("facing", "west");
```

This also allows to ensure the value can be handled gracefully.


mworzala:
I do not want to add an API like this as a stub for people to implement themselves (constants can do something similar, with less type checking admittedly). I am not opposed to adding a typed property API however I do not want to take the maintenance cost of keeping it up to date so would be opposed to any change not based on a code generator. Thus far nobody has come up with a sane code generator for it.

Happy to discuss further here or on Discord.

PR #2580, opened 2025-01-04
Expose Physics Result Cache

KermanIsPretty:
Will be used in a upcoming BlockHandler pr targeted for the 1.21.4. Since this should not include any breaking changes this should be safe to be pushed to mainstream.


KermanIsPretty:
Can we get this merged?

Bloeckchengrafik (approved):

PR #2590, opened 2025-01-09
Remove reset of values from PaletteIndirect

MrMineO5:
Currently, PaletteIndirect resets the values at the end of the constructor. This is unexpected behaviour given the values are passed as a constructor parameter and results in incorrect reading of palettes from NetworkBuffer.

### DIFF
--- a/src/main/java/net/minestom/server/instance/palette/PaletteIndirect.java
+++ b/src/main/java/net/minestom/server/instance/palette/PaletteIndirect.java
@@ -49,8 +49,6 @@ final class PaletteIndirect implements SpecializedPalette, Cloneable {
             this.paletteToValueList.add(palette[i]);
             this.valueToPaletteMap.put(palette[i], i);
         }
-
-        this.values = new long[arrayLength(dimension(), bitsPerEntry)];
     }
 
     PaletteIndirect(int dimension, int maxBitsPerEntry, byte bitsPerEntry) {
--- a/src/test/java/net/minestom/server/instance/palette/PaletteIndirectTest.java
+++ b/src/test/java/net/minestom/server/instance/palette/PaletteIndirectTest.java
@@ -1,0 +1,17 @@
+package net.minestom.server.instance.palette;
+
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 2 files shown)


src/main/java/net/minestom/server/instance/palette/PaletteIndirect.java
```diff
@@ -49,8 +49,6 @@ final class PaletteIndirect implements SpecializedPalette, Cloneable {
             this.paletteToValueList.add(palette[i]);
             this.valueToPaletteMap.put(palette[i], i);
         }
-
-        this.values = new long[arrayLength(dimension(), bitsPerEntry)];
```
TheMeinerLP, line 53: Is there some reason why that happened or how we can reproduce it?

If it possible to add a test for that to prevent such behaviour in the future.

MrMineO5: Well a simple test would be writing a palette into a NetworkBuffer, reading it back out, and comparing that they're the same...

I'm not sure what you're trying to reproduce here, the state of the code is as such and it makes no sense to pass values a a parameter to the constructor and then immediately override the field

TheMeinerLP: I just don't have the proof, the test would ensure that other contributors don't make the same mistake

MrMineO5: I described a test that would prevent this specific mistake, the general way to prevent this type of mistake is code review.

TheMeinerLP: Implement a test instead of describing it

MrMineO5:
There's a test, not sure why exactly you find a test for this necessary, but enjoy!

src/main/java/net/minestom/server/instance/palette/PaletteIndirect.java
MrMineO5: Does that test satisfy your requirements?

mworzala (approved):

PR #2592, opened 2025-01-10
Skip size varint when reading single palette

MrMineO5:
The read method for palettes does not skip the unused (always 0) size varint of single palettes. This results in incorrect parsing of packets when data is present after the palette and constitutes and inconsistency between the read and write methods of the type.


mworzala (approved):
Good catch, thanks

PR #2594, opened 2025-01-12
fix: double sending of disconnect events

### LINKED ISSUE #2593: Double PlayerDisconnectEvent on Player#kick()
When kicking a player using the Player#kick() method, there are two resulting PlayerDisconnectEvents. This happens in any stage of connection (configuration, play, etc)

This is caused by net.minestom.server.network.socket.Server#playerReadLoop having an EOFException after the player is kicked, calling a secondary PlayerConnection#disconnect, and therefore calling a new event.

colton-boi:
Fixes #2593 

Ensure that the player isn't already disconnected before calling a new PlayerDisconnectEvent


hapily04 (approved):
Tested and this does fix it. LGTM

PR #2595, opened 2025-01-13
Add server flag for ChunkUpdateLimitChecker history size

socolio:


socolio:
https://github.com/Minestom/Minestom/pull/2596

PR #2596, opened 2025-01-13
Add server flag for ChunkUpdateLimitChecker history size

socolio:
ChunkUpdateLimitChecker may cause some visual glitches with invisible chunks e.t.c due to its ability to block some UpdateViewPosition packets.

I want to have a full control over what chunks player will see.

### DIFF
--- a/src/main/java/net/minestom/server/ServerFlag.java
+++ b/src/main/java/net/minestom/server/ServerFlag.java
@@ -27,6 +27,7 @@ public final class ServerFlag {
     public static final int PLAYER_PACKET_QUEUE_SIZE = intProperty("minestom.packet-queue-size", 1000);
     public static final long KEEP_ALIVE_DELAY = longProperty("minestom.keep-alive-delay", 10_000);
     public static final long KEEP_ALIVE_KICK = longProperty("minestom.keep-alive-kick", 15_000);
+    public static final int PLAYER_CHUNK_UPDATE_LIMITER_HISTORY_SIZE = intProperty("minestom.player.chunk-update-limiter-history-size", 6);
 
     // Network buffers
     public static final int MAX_PACKET_SIZE = intProperty("minestom.max-packet-size", 2_097_151); // 3 bytes var-int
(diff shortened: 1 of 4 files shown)


GreatWyrm:
Personally, I think it should throw an error if PLAYER_CHUNK_UPDATE_LIMITER_HISTORY_SIZE is negative, rather than setting to 0.
Also, minor nitpick, but you could change the comparisons to be < 0 in the constructor and == 0 in the addToHistory.

socolio:
> Personally, I think it should throw an error if PLAYER_CHUNK_UPDATE_LIMITER_HISTORY_SIZE is negative, rather than setting to 0. Also, minor nitpick, but you could change the comparisons to be < 0 in the constructor and == 0 in the addToHistory.

For some users who dont know how ChunkUpdateLimitChecker works, it may not be clear that history size = 1 will disable the limiter. So someone may try to set the limit to 0 or -1, as it is popular values for "disabling" something. That is why all non-positive values should disable the limiter.

About minor nitpick. In my experience, it is better to double the same check in constructor and addToHistory just for extra safety. Someone may change the constraint in constructor and forgot to change it in addToHistory.

GreatWyrm:
If it's unclear how it works, it should be javadoc'd. I'd recommend you add some javadocs to the class.
Throwing an exception on player creation is bad, and I'd prefer throwing it on server start. Changes to the intProperty method to add a check would be the better option in that case.

And for the minor nitpick, instead of having the check in two places, please refactor it out to a function named something like isValidBounds, so anyone changing this class in the future only has to change one place.

socolio:
> If it's unclear how it works, it should be javadoc'd. I'd recommend you add some javadocs to the class. Throwing an exception on player creation is bad, and I'd prefer throwing it on server start. Changes to the intProperty method to add a check would be the better option in that case.
> 
> And for the minor nitpick, instead of having the check in two places, please refactor it out to a function named something like isValidBounds, so anyone changing this class in the future only has to change one place.

Done. Also fixed up limiter logic (it did not check for the last incoming chunk) and added tests.

mworzala (approved):
I am down, but I think ChunkUpdateLimitChecker needs a bigger rework in the future.

Issue #2598, opened 2025-01-15
Block properties aren't sent to the client properly


lordtgm:
I'm replacing a redstone wire block with altered "power" property, which is properly registered on the server. however, the client does not see the new block state until I disconnect and reconnect to the server. this occurs in the block's handler's onPlace method,

lordtgm:
reproducer:
```java
MinecraftServer.getGlobalEventHandler().addListener(PlayerBlockPlaceEvent.class, playerBlockPlaceEvent -> {
            if (playerBlockPlaceEvent.getBlock().namespace() == Block.REDSTONE_WIRE.namespace()) {
                playerBlockPlaceEvent.setBlock(playerBlockPlaceEvent.getBlock().withHandler(new BlockHandler() {
                    @Override
                    public @NotNull NamespaceID getNamespaceId() {
                        return NamespaceID.from("minestom:test");
                    }

                    @Override
                    public void onPlace(@NotNull BlockHandler.Placement placement) {
                        placement.getInstance().setBlock(
                                placement.getBlockPosition(),
                                placement.getInstance().getBlock(placement.getBlockPosition())
                                        .withProperty("power","15")
                        );
                    }
                }));
            }
        });
```

PR #2599, opened 2025-01-15
Fix block replacement within onPlace desync

lordtgm:
solves https://github.com/Minestom/Minestom/issues/2598

InstanceContainer#setBlock calls InstanceContainer#UNSAFE_setBlock, which first calls Chunk#setBlock and then sends the updated block to the viewers. Ff the block is changed via InstanceContainer#setBlock inside the block's BlockHandler#onPlace, the inner setBlock (which happens after the outer setBlock) will send its update before the outer setBlock, which leads to the update being overwritten by its outer and former update.
With this PR only the innermost setBlock will send its update, so only the last update is sent to the viewers.

### DIFF
--- a/src/main/java/net/minestom/server/instance/InstanceContainer.java
+++ b/src/main/java/net/minestom/server/instance/InstanceContainer.java
@@ -188,6 +188,14 @@ public class InstanceContainer extends Instance {
             // Set the block
             chunk.setBlock(x, y, z, block, placement, destroy);
 
+            /* Check if the block was altered, cancel the rest of the operation.
+            Because the block might be changed by block handlers,
+            This is done so only the innermost setBlock handles the update.
+             */
+            if (chunk.getBlock(x, y, z) != block) {
+                return;
+            }
+
             // Refresh neighbors since a new block has been placed
             if (doBlockUpdates) {
                 executeNeighboursBlockPlacementRule(blockPosition, updateDistance);


src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -188,6 +188,14 @@ private synchronized void UNSAFE_setBlock(@NotNull Chunk chunk, int x, int y, in
             // Set the block
             chunk.setBlock(x, y, z, block, placement, destroy);
 
+            /* Check if the block was altered, cancel the rest of the operation.
+            Because the block might be changed by block handlers,
+            This is done so only the innermost setBlock handles the update.
+             */
+            if (chunk.getBlock(x, y, z) != block) {
```
TheMeinerLP, line 195: Would it be possible to write a test for this to avoid future errors?

TheMeinerLP (changes requested):

mworzala:
I'm not very sure on this solution it feels like a very unnecessary block read for quite hot code 🤔 However I don't immediately have a solution I like better.

TheMeinerLP:
> I'm not very sure on this solution it feels like a very unnecessary block read for quite hot code 🤔 However I don't immediately have a solution I like better.

We could take a closer look after or outside of this. There may be better solutions.

For the current case, I would like to simply merge them after the small test

src/main/java/net/minestom/server/instance/InstanceContainer.java
lordtgm: I'm not sure what test you mean here. to ensure that block changes within onPlace get sent properly?

TheMeinerLP: Simulate the condition and expected behaviour

KermanIsPretty:
If you want to move this out of the hot code block we could just append PlacementRules into BlockImpl; allowing you to dynamically setup the block; example `Block#withPlacementRule` and thats what it should be for mutating a onPlace. Unsure if thats the best implementation as it increases the memory footprint of BlockImpl but its a pallete anyways.

lordtgm:
I've rebased my branch upon master, but I'm unsure how I'm supposed to write such a test. any help is appreciated.

PR #2600, opened 2025-01-16
Event Rework

KermanIsPretty:
Sorry for the large commit. It was hard to split up the work as it would either compile or not.

- Delete Depercated functions inside events.
- Use record name for events unless inside the mutator. This is a heavily breaking change for an already major breaking change. 
- Allow split mutation, I dont like how this is implemented maybe possible to do something like the network buffers for mutability?

Most of the events need documentation. Any field I found that didnt have any and was getting renamed was deleted.

Seems to work on demo, fails unit tests need more testing. Probably my bad implementation of. Approx 15 tests are failing. (Most due to the updated tester implementation maybe?)

For now this a draft and possibly needs more refactoring. Anyways time for sleep.


hapily04:
I don't like the idea of "immutable" events in minestom since the ones that are mutable will just create more garbage when you mutate them. I don't see a benefit from this change and I feel like it adds unnecessary complexity. I don't think minestom cares a ton about backwards compatibility with people's code if at all, but with this, in my eyes, just adding unnecessary complexity, I don't think it's worth breaking people's code.

DasBabyPixel:
I think for this to ever get merged you need to outline the problem, issue or inconvenience you are fixing and why this is better than the old system. Otherwise, yeah, seems like added complexity without much benefit.

mworzala:
Yep I am in agreement with the previous takes, unless there is a very good problem this is solving it seems very fine for events to be mutated to me. Happy to discuss further here or on Discord (feel free to ping)

PR #2601, opened 2025-01-16
added Block.Setter parameter in the BlockPlacementRule

einfachBlu:
as talked about on discord, this pr adds the Block.Setter to the BlockPlacementRule so BlockPlacementRule's can set blocks. Useful for Doors to place the upper part or update nearby blocks if needed.

Additionally it also deprecates the instance() field because the naming does not make sense if both getter and setter fields are present there.

### DIFF
--- a/src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
+++ b/src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
@@ -55,7 +55,8 @@ public abstract class BlockPlacementRule {
     }
 
     public record PlacementState(
-            @NotNull Block.Getter instance,
+            @NotNull Block.Getter blockGetter,
+            @NotNull Block.Setter blockSetter,
             @NotNull Block block,
             @Nullable BlockFace blockFace,
             @NotNull Point placePosition,
@@ -64,6 +65,10 @@ public abstract class BlockPlacementRule {
             @Nullable ItemStack usedItemStack,
             boolean isPlayerShifting
     ) {
+        @Deprecated
+        public Block.Getter instance() {
+            return this.blockGetter;
+        }
     }
 
     public record UpdateState(@NotNull Block.Getter instance,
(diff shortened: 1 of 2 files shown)


Bloeckchengrafik (approved):
Nice work!

src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
```diff
@@ -64,12 +65,21 @@ public record PlacementState(
             @Nullable ItemStack usedItemStack,
             boolean isPlayerShifting
     ) {
+        @Deprecated
```
TheMeinerLP, line 68: Can you add some Javadoc with replacement ?

src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
```diff
@@ -64,12 +65,21 @@ public record PlacementState(
...
+        }
     }
 
-    public record UpdateState(@NotNull Block.Getter instance,
+    public record UpdateState(@NotNull Block.Getter blockGetter,
+                              @NotNull Block.Setter blockSetter,
                               @NotNull Point blockPosition,
                               @NotNull Block currentBlock,
                               @NotNull BlockFace fromFace) {
+        @Deprecated
```
TheMeinerLP, line 79: Also here

TheMeinerLP (changes requested):

src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
einfachBlu: yes, I have added that

src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
einfachBlu: yes, I have added that

PR #2607, opened 2025-01-19
Update ClientVehicleMovePacket to 1.21.4

BL19:
When riding a boat, I got the following message from the demo server on the 1.21.4 branch:
`23:28:56.872 [] WARN net.minestom.server.network.packet.PacketReading - WARNING: Packet (ClientVehicleMovePacket) 0x20 not fully read (NetworkBuffer{r34|w35->16383, registries=true, autoResize=true, readOnly=false})`

It seems a new field got added to the packet in 1.21.4, this PR adds that field to the packet record.
[Packet on the Protocol Wiki for 1.21.4](https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol?oldid=2821146#Move_Vehicle_2)

### DIFF
--- a/src/test/java/net/minestom/server/network/PacketWriteReadTest.java
+++ b/src/test/java/net/minestom/server/network/PacketWriteReadTest.java
@@ -5,6 +5,7 @@ import net.kyori.adventure.bossbar.BossBar;
 import net.kyori.adventure.nbt.CompoundBinaryTag;
 import net.kyori.adventure.text.Component;
 import net.minestom.server.MinecraftServer;
+import net.minestom.server.coordinate.Pos;
 import net.minestom.server.coordinate.Vec;
 import net.minestom.server.entity.EquipmentSlot;
 import net.minestom.server.entity.GameMode;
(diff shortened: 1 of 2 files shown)


TheMeinerLP (changes requested):
Can you please add a JUnit5 Test for prevent further issues in this topic ?

Is a small change and so far so good from my side.

BL19:
Could you please clarify what kind of test you would want for this change?

TheMeinerLP:
Proof the read and write process

src/test/java/net/minestom/server/network/PacketWriteReadTest.java
```diff
@@ -141,6 +143,8 @@ public static void setupServer() {
     @BeforeAll
     public static void setupClient() {
         CLIENT_PACKETS.add(new ClientHandshakePacket(755, "localhost", 25565, ClientHandshakePacket.Intent.LOGIN));
+        CLIENT_PACKETS.add(new ClientVehicleMovePacket(new Pos(5, 5, 5, 45f, 45f), true));
+        CLIENT_PACKETS.add(new ClientVehicleMovePacket(new Pos(6, 5, 6, 82f, 12.5f), false));
```
TheMeinerLP, line 147: Unfortunately, this does not test the read or write process but only whether the packet can be created from the Java view.

TheMeinerLP (changes requested):

src/test/java/net/minestom/server/network/PacketWriteReadTest.java
BL19: Correct me of I'm wrong, but doesn't that test class do exactly what you are asking for? It writes the packet to it's byte form and then reads it again from said bytes.

mworzala: It does, but it would also pass prior to this PR because it starts from a wrong definition anyway. We don't currently have any structure for tests in the requested way so its fine for now I think.

TheMeinerLP: It does not check whether bytes are correct, but only writes them. Otherwise the test is unnecessary and checks nothing

PR #2609, opened 2025-01-20
Block Sound Type Generation

GreatWyrm:
- ~~Switch code generation library since the [old one is archived](https://github.com/square/javapoet), [discussion](https://github.com/square/javapoet/discussions/866).~~
- Add Block Sound Type generation
- Add block sound types to the block registry entries


TheMeinerLP (changes requested):
We now have 2 options.
Option 1. we switch to FreeMarker from Apache to gain more flexibility for the generation
Option 2. we simply take another lib

Please separate feature request from rework or customization of the lib, this creates noise in the review or causes you to criticize things that may not be relevant at all :D

src/autogenerated/java/net/minestom/server/sound/BlockSoundType.java
```diff
@@ -0,0 +1,261 @@
+package net.minestom.server.sound;
+
+import java.util.HashMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * AUTOGENERATED by BlockSoundTypesGenerator
+ */
+public record BlockSoundType(float volume, float pitch, @NotNull SoundEvent breakSound,
```
TheMeinerLP, line 10: It doesn't quite fit into the scheme and I didn't see that in the first review. I would prefer to solve this with some kind of registry instead of custom code generators

TheMeinerLP (changes requested):

src/autogenerated/java/net/minestom/server/sound/BlockSoundType.java
GreatWyrm: It doesn't make sense in a registry, the data isn't stored in a registry in Minecraft and isn't sent in any network buffer.

GreatWyrm:
@TheMeinerLP 
I'm...not criticizing anything?
I went to add sound types into Minestom, worked on it for a bit, decided it would be best as a record.
Then went to code generate it as a record, saw that the current library 1. Did not support records, and 2. Was archived, and figured I could swap the library in this PR.
It's essentially a drop-in replacement since the code was forked from the original.

TheMeinerLP:
> I'm...not criticizing anything?

From my point of view, the critical aspect is to prevent duplication of effort

TheMeinerLP:
> Then went to code generate it as a record, saw that the current library 1. Did not support records, and 2. Was archived, and figured I could swap the library in this PR.

Record can be used, but from a general point of view it would be better to keep the standard the same for generation. If 90% use an interface with an inheritance structure, we should continue to do so in order to reduce the maintenance effort in the future

src/autogenerated/java/net/minestom/server/sound/BlockSoundType.java
mworzala, line 10: This should mimic the other generated resources:
* BlockSoundType interface inheriting from BlockSoundTypes generated with all constants.
* (package private) BlockSoundTypeImpl record containing the relevant fields.

(I think a registry could also work, but that seems to me like it needs a little more discussion. Either way I dont think we should introduce a new style of codegen for it)

code-generators/src/main/java/net/minestom/codegen/sounds/BlockSoundTypesGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private final String packageName;
+    private final String className;
+    private final InputStream entriesFile;
+    private final File outputFolder;
+
+    public BlockSoundTypesGenerator(
+            @NotNull String packageName, @NotNull String className,
+            @Nullable InputStream entriesFile, @NotNull File outputFolder
+    )
```
mworzala, line 31: weird formatting. Run IJ formatter on this file.

gradle/libs.versions.toml
```diff
@@ -65,7 +65,7 @@ jmh-annotationprocessor = { group = "org.openjdk.jmh", name = "jmh-generator-ann
 jcstress-core = { group = "org.openjdk.jcstress", name = "jcstress-core", version.ref = "jcstress" }
 
 # Code Generation
-javaPoet = { group = "com.squareup", name = "javapoet", version.ref = "javaPoet" }
+javaPoet = { group = "com.palantir.javapoet", name = "javapoet", version.ref = "javaPoet" }
```
mworzala, line 68: I would prefer to stick with javapoet for the moment personally, we only use its output (which is very simple in our case) so it's not a security concern. I am a little worried about palantir's pre-1.0 downgrade as well as their decision to change the package name.

mworzala (changes requested):

gradle/libs.versions.toml
GreatWyrm: To clarify (since both libraries are called javapoet), would you prefer I switch the library back to the original one from squareup?
If so, I'll have to change the generation slightly since the old library cannot generate records.

src/autogenerated/java/net/minestom/server/sound/BlockSoundType.java
GreatWyrm: Done
As said above, I'm not sure it would work in a registry since BlockSoundTypes aren't sent through the network and don't extend StaticProtocolObject.

src/autogenerated/java/net/minestom/server/sound/BlockSoundTypes.java
```diff
@@ -0,0 +1,251 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * AUTOGENERATED by BlockSoundTypesGenerator
+ */
+public interface BlockSoundTypes {
+    HashMap<String, BlockSoundType> SOUND_TYPE_INDEX = new HashMap<>();
+
+    BlockSoundType EMPTY = new BlockSoundTypeImpl("empty", 1.0f, 1.0f, BuiltinSoundEvent.get("minecraft:intentionally_empty"), BuiltinSoundEvent.get("minecraft:intentionally_empty"), BuiltinSoundEvent.get("minecraft:intentionally_empty"), BuiltinSoundEvent.get("minecraft:intentionally_empty"), BuiltinSoundEvent.get("minecraft:intentionally_empty"));
```
TheMeinerLP, line 13: Are the values listed in a JSON file?

GreatWyrm: Yes, they're part of the MinestomDataGenerator's output

GreatWyrm: Like with generator.generateKeys?

That could work, but I'd need to edit how the JSON is generated on the DataGen side a bit, and again I don't think BlockSounds would work with a registry.

They still aren't network data that needs to be synced to a client.

TheMeinerLP: I adjusted the Data Geneator and would propose something like that:
```java
interface BlockSoundTypes {
    BlockSoundType EMPTY = BlockSoundTypeImpl.get("minestom:empty");

    BlockSoundType WOOD = BlockSoundTypeImpl.get("minestom:wood");

    BlockSoundType GRAVEL = BlockSoundTypeImpl.get("minestom:gravel");

    BlockSoundType GRASS = BlockSoundTypeImpl.get("minestom:grass");
    // ... 
}
```

gradle/libs.versions.toml
mworzala: square

GreatWyrm: Reverted, though I'll open an issue regarding the archived library.

src/autogenerated/java/net/minestom/server/sound/BlockSoundTypes.java
```diff
@@ -0,0 +1,251 @@
...
+
+import java.util.HashMap;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * AUTOGENERATED by BlockSoundTypesGenerator
+ */
+public interface BlockSoundTypes {
+    HashMap<String, BlockSoundType> SOUND_TYPE_INDEX = new HashMap<>();
```
TheMeinerLP, line 11: Please put into implementation class with unmodified return value

TheMeinerLP (changes requested):

mworzala (changes requested):
This needs rebase + should be changed to look like the existing generated types (using Registry.Container, not DynamicRegistry)

GreatWyrm:
@mworzala 
Requires the merging of https://github.com/Minestom/MinestomDataGenerator/pull/41 to proceed, as the block sound type data is not in the format the Registry expects in it's load method.

code-generators/src/main/java/net/minestom/codegen/Generators.java
```diff
@@ -41,6 +42,8 @@ public static void main(String[] args) {
                 resource("villager_types.json"), outputFolder).generate();
         new WorldEventGenerator("net.minestom.server.worldevent", "WorldEvent",
                 resource("world_events.json"), outputFolder).generate();
+        new BlockSoundTypesGenerator("net.minestom.server.sound", "BlockSoundTypes", "BlockSoundTypeImpl",
+                resource("block_sound_types.json"), outputFolder).generate();
```
TheMeinerLP, line 46: I this still needed after rework ?

gradle/libs.versions.toml
```diff
@@ -3,7 +3,7 @@ metadata.format.version = "1.1"
 [versions]
 
 # Important dependencies
-data = "1.21.4-rv1"
+data = "1.21.4-dev"
```
TheMeinerLP, line 6: We need to double-check that, so far I know we don't rely on dev versions for production.

src/main/java/net/minestom/server/instance/block/BlockSoundImpl.java
```diff
@@ -0,0 +1,23 @@
+package net.minestom.server.instance.block;
+
+import net.minestom.server.registry.Registry;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+
+public record BlockSoundImpl(Registry.BlockSoundTypeEntry registry) implements BlockSoundType {
+    private static final Registry.Container<BlockSoundType> CONTAINER = Registry.createStaticContainer(Registry.Resource.BLOCK_SOUND_TYPES,
+            (namespace, properties) -> new BlockSoundImpl(Registry.blockSoundTypeEntry(namespace, properties)));
```
TheMeinerLP, line 10: Is there not the generated interfaces missing?

src/main/java/net/minestom/server/instance/block/BlockSoundType.java
```diff
@@ -0,0 +1,69 @@
...
+    default @NotNull SoundEvent placeSound() {
+        return registry().placeSound();
+    }
+
+    default @NotNull SoundEvent stepSound() {
+        return registry().stepSound();
+    }
+
+    default int id() {
+        return 0; // Not sent through packets in the protocol, also must be between 0 and [size of block sound type list] because id mappings are stored in an array
```
TheMeinerLP, line 67: Should be better to generated ID still for some equal checking

TheMeinerLP (changes requested):
Sorry for my late response

gradle/libs.versions.toml
GreatWyrm: Until a new release is published for the DataGen repo, I have to use the dev build.

src/main/java/net/minestom/server/instance/block/BlockSoundImpl.java
GreatWyrm: No? What are you expecting here?

src/main/java/net/minestom/server/instance/block/BlockSoundType.java
GreatWyrm: I disagree here. This is from StaticProtocolObject, and having an actual implementation for id would be wrong.
BlockSoundTypes aren't sent through the Protocol, so having a defined id per type wouldn't make much sense.

src/main/java/net/minestom/server/instance/block/BlockSoundImpl.java
GreatWyrm: I could do that, but the elements are attached to the Block Registry Entries and my intent is that you access them through the blocks.
I can have them standalone like the Material list if that's requested.

GreatWyrm: Talked with mattw, added the list.

TheMeinerLP (approved):
LGTM

mworzala (approved):

PR #2610, opened 2025-01-20
fix: Add simple boat meta check for velocity fix

### LINKED ISSUE #1880: Boat glitches when going off two block tall step on ice
When you ride a baot on ice and go down over a ledge two or more blocks tall, you lose your forward momentum.

Illustration video in attachments. (there isn't directly two blocks tall ledge, but the boat goes two blocks down in total when it happens)

Uploading Recording 2023-06-14 094124.mp4…

TheMeinerLP:
This pr fixes #1880 and add a additional check into a entity to prevent the velocity is send to the client if the type a boat.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,6 +80,8 @@ dependencies {
 
     // Testing
     testImplementation(libs.bundles.junit)
+    testImplementation(platform(libs.mockito.bom))
+    testImplementation(libs.bundles.mockito)
     testImplementation(project(":testing"))
 }
 
--- a/src/test/java/net/minestom/server/entity/BoatEntityTest.java
+++ b/src/test/java/net/minestom/server/entity/BoatEntityTest.java
@@ -45,7 +45,7 @@ class BoatEntityTest {
     @DisplayName("Test if velocity packet is not sent after 20 ticks if the entity a boat")
     @Issue("1880")
     void testIfVelocityPacketIsNotSent(EntityType entityType, Env env) {
-        final Entity entity = Mockito.spy(new Entity(EntityTypes.ACACIA_BOAT));
+        final Entity entity = Mockito.spy(new Entity(entityType));
         Instance flatInstance = env.createFlatInstance();
         entity.setInstance(flatInstance, Pos.ZERO).join();
         for (int i = 0; i < 21; i++) {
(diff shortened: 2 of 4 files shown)


KermanIsPretty:
Would the same issue happen if a client is on a pig or a horse?

TheMeinerLP:
I am not sure, needs to be tested. You can raise an issue with proof for that.

src/test/java/net/minestom/server/entity/BoatEntityTest.java
```diff
@@ -0,0 +1,75 @@
...
+            Arguments.of(EntityType.BAMBOO_CHEST_RAFT),
+            Arguments.of(EntityType.BAMBOO_RAFT)
+    );
+
+    @ParameterizedTest
+    @MethodSource("getTestTypes")
+    @DisplayName("Test if velocity packet is not sent after 20 ticks if the entity a boat")
+    @Issue("1880")
+    void testIfVelocityPacketIsNotSent(EntityType entityType, Env env) {
+        final Entity entity = Mockito.spy(new Entity(EntityTypes.ACACIA_BOAT));
```
Bloeckchengrafik, line 48:
```suggestion
        final Entity entity = Mockito.spy(new Entity(entityType));
```

Seems like the entity type parameter is not referenced anywhere. Should this be used here?

TheMeinerLP: Thanks for take a look

build.gradle.kts
```diff
@@ -80,6 +80,8 @@ dependencies {
 
     // Testing
     testImplementation(libs.bundles.junit)
+    testImplementation(platform(libs.mockito.bom))
```
mworzala, line 83: I do not like mockito tests or want to add it. They are unintuitive (to me at least) and unreliable across changes long term. In this case you can just use a test player to get the packets (or write a little utility to do that and ensure none received if you dont like the boilerplate)

mworzala (changes requested):
(just adding this because my I other comment was a single comment not review and GitHub still shows waiting for review)

build.gradle.kts
TheMeinerLP: From my point of view that would be a bit too much just “a bit” of quality for sure. What you suggest would be an integration test, during which I try to write an implementation test. Unfortunately, there is no better way in the current framework to catch such cases!

PR #2581, opened 2025-01-04
Update BlockHandlers

kermandev:
As it currently stands BlockHandlers have some issues with `BlockHandler#onTouch` as described by #2353 and fixes a global position issue (introduced in #2409). The major change between before this PR is  `BlockHandler#onTouch` as it would constantly fire like it was `onTouching` seems a bit misleading so instead its been condensed down to not fire if the entity isn't moving, mostly checked by `Entity#shouldSkipTouchTick`

This PR reworks some other portions of block handlers. Also adds new tests and splits players `Player#touchTick` implementation. The player implementation follows a bad behavior when inside blocks as there is a limitation with entities controllers.

Started using the cached physics (#2580) to determine the `Entity#touchTick` and check to see if the entity has been moving. The only downside I see with using the cached physics result for entities is that it only checks a single axis's so there is a bias to blocks. But the flip side is its much faster to check.

This also breaks the bad behavior of  `BlockHandler#onTouch` getting fired when an entity was inside a block. The player implementation follows the bad behavior as there is a limitation with entities get controlled in Minestom.

This PR will rebase on 1.21.4 branch when #2580 is merged into both branches.

---
## Breaking changes
- Converstion to records (Tick, Touch, Interaction, Dummy)
- `BlockHandler#onDestroy` uses absolute block position
- `BlockHandler#onPlace` uses the correct absolute block position
- The way that `Entity#touchTick` (now protected) works has changed completely and changes how `BlockHandler#onTouch` works
- `BlockHandler#onTouch` only fires when entity is moving configurable with `Entity#isTickingTouch` 
- `BlockHandler#onTouch` fires differently between players and entities, Configurable with `ServerFlag.USE_FAST_TOUCH` or `Entity#isFastTouch`  
- Added `Entity#touchBoundingBox` field that could break memory segment parsing usage. Unsure if anyone is doing this

---
Fixes #2353

PR #2613, opened 2025-01-21
BoundingBox updates

KermanIsPretty:
This will allow people to expand and create center (Feet aligned) bounding boxes easier. This was a bit easier to do when all boxes were considered centered but that has since been reverted. See https://github.com/Minestom/Minestom/commit/10d6d2bb2ed57414f18f1cf2f0b5b0441770d216

The new `BlockHandler#growSymmetrically` functions more like notchian AABB expand. Will be used to replace magic in #2581

### DIFF
--- a/src/main/java/net/minestom/server/collision/BoundingBox.java
+++ b/src/main/java/net/minestom/server/collision/BoundingBox.java
@@ -5,6 +5,7 @@ import net.minestom.server.coordinate.Pos;
 import net.minestom.server.coordinate.Vec;
 import net.minestom.server.entity.EntityPose;
 import net.minestom.server.instance.block.BlockFace;
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/collision/BoundingBox.java
```diff
@@ -94,6 +95,40 @@ public boolean boundingBoxRayIntersectionCheck(Vec start, Vec direction, Pos pos
...
+     *
+     * @param x the X offset, this will be applied on both sides
+     * @param y the Y offset, this will be applied on both sides
+     * @param z the Z offset, this will be applied on both sides
+     * @return a new {@link BoundingBox} expanded and centered from the original minY
+     */
+    @Contract(pure = true)
+    public @NotNull BoundingBox grow(double x, double y, double z) {
+        final Vec centerOffset = new Vec(-(width() + x) / 2, minY() - y / 2, -(depth() + z) / 2);
+        return new BoundingBox(width() + x, height() + y, depth() + z, centerOffset);
```
TheMeinerLP, line 112: Nitpick: Please remove magic values into Constants

src/main/java/net/minestom/server/collision/BoundingBox.java
```diff
@@ -94,6 +95,40 @@ public boolean boundingBoxRayIntersectionCheck(Vec start, Vec direction, Pos pos
...
+     *
+     * @param x the X offset, this will be applied on both sides
+     * @param y the Y offset, this will be applied on both sides
+     * @param z the Z offset, this will be applied on both sides
+     * @return a new {@link BoundingBox} expanded and centered from the original minY
+     */
+    @Contract(pure = true)
+    public @NotNull BoundingBox growSymmetrically(double x, double y, double z) {
+        final Vec centerOffset = new Vec(-width() / 2 - x, minY() - y, -depth() / 2 - z);
+        return new BoundingBox(width() + 2 * x, height() + 2 * y, depth() + 2 * z, centerOffset);
```
TheMeinerLP, line 129: DRY (Don't repeat yourself)

TheMeinerLP (changes requested):

src/main/java/net/minestom/server/collision/BoundingBox.java
KermanIsPretty: 2 is an acceptable magic value

mworzala (approved):

Issue #1622, opened 2022-12-26
ItemEntity goes out of sync with client


emortaldev:
An item entity likes to go out of sync with the client
- sometimes appearing in a location but not being able to be picked up (normally after velocity applied)
- after being alive for a while the item entity likes to go one block down into the ground (likely sync cooldown)

glowgrew:
Can confirm, I was able to fix this by setting custom syncronization cooldown to 1 day 💀

emortaldev:
> Can confirm, I was able to fix this by setting custom syncronization cooldown to 1 day skull

I assume you were only able to fix the bottom thing

glowgrew:
> I assume you were only able to fix the bottom thing

I fixed both things, but before I was able to pick up an item without actually seeing it

PR #2614, opened 2025-01-21
Post Breaking Bounding Box Updates

KermanIsPretty:
# Requires: #2613 
Thought it would be a good idea to seperate breaking changes; Changes on how items are picked up. Seemed to be an issue where picking up items below, as seen with the previous location subtraction (would make it so the upperbody cannot collect anything). This should fix that as the bounding box is expanded to include the Y component. Possibly fixes any pickup issue with https://github.com/Minestom/Minestom/issues/1622

Breaks any usage of LivingEntity#expandedBoundingBox as that old position doesn't need to be accounted for anymore.


KermanIsPretty:
Will rebase when upstream;

TheMeinerLP:
Marked into draft for waiting https://github.com/Minestom/Minestom/pull/2613

KermanIsPretty:
Waiting for upstream

KermanIsPretty:
Should be good to merge

PR #2618, opened 2025-01-23
Dont floor BlockVec's xyz again

KermanIsPretty:
Seems weird this hasn't come up in the coord conversation.

### DIFF
--- a/src/main/java/net/minestom/server/coordinate/BlockVec.java
+++ b/src/main/java/net/minestom/server/coordinate/BlockVec.java
@@ -29,6 +29,21 @@ public record BlockVec(double x, double y, double z) implements Point {
     }
 
     @Override
+    public int blockX() {
+        return (int) x;
+    }
+
+    @Override
+    public int blockY() {
+        return (int) y;
+    }
+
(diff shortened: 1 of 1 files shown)


GoldenStack:
Why? They're already floored:
```java
    public BlockVec {
        x = Math.floor(x);
        y = Math.floor(y);
        z = Math.floor(z);
    }
```

and the methods you're overriding already call this function on the values:
```java
    public static int globalToBlock(double xyz) {
        return (int) Math.floor(xyz);
    }
```

So what does this change?

Bloeckchengrafik (approved):

mworzala:
The constructor of BlockVec floors them, im unclear what the point of this is.

Bloeckchengrafik:
In the default impl of the point interface, 
```java
public static int globalToBlock(double xyz) {
        return (int) Math.floor(xyz);
}
```
is called. This floors it a second time for no reason at all

mworzala (approved):
I see, I just wanted to make sure it wasn't something I was missing besides a very micro optimization. Seems good otherwise

PR #2621, opened 2025-01-24
refactor: move from getUniqueId() to getUuid() and related

### LINKED ISSUE #1616: Make usage of uuid/uniqueId consistent
Both `uniqueId` and `uuid` naming is used, e.g `Instance#getUniqueId` vs `Entity#getUuid`. This can mislead people into thinking an instance doesn't have an ID since `getUuid` does not exist.

It would be useful to pick one name to be used across the project and deprecate the old methods.

Bloeckchengrafik:
This PR solves #1616. See that issue for more details.

closes #1616

CHANGE: Public API of Instance has changed! getUniqueId() -> getUuid()
Old version is deprecated

### DIFF
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -758,7 +758,7 @@ public abstract class Instance implements Block.Getter, Block.Setter,
      * Gets the instance unique id.
      *
      * @return the instance unique id
-     * @deprecated Use Instance#getUuid
+     * @deprecated Replace with Instance#getUuid
      */
     @Deprecated(forRemoval = true)
     public @NotNull UUID getUniqueId() {
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -755,6 +755,17 @@ public abstract class Instance implements Block.Getter, Block.Setter,
     }
 
     /**
+     * Gets the instance unique id.
+     *
+     * @return the instance unique id
+     * @deprecated Use Instance#getUuid
+     */
+    @Deprecated(forRemoval = true)
+    public @NotNull UUID getUniqueId() {
+        return uuid;
+    }
+
+    /**
      * Performs a single tick in the instance, including scheduled tasks from {@link #scheduleNextTick(Consumer)}.
      * <p>
      * Warning: this does not update chunks and entities.
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -758,7 +758,7 @@ public abstract class Instance implements Block.Getter, Block.Setter,
      * Gets the instance unique id.
      *
      * @return the instance unique id
-     * @deprecated Replace with Instance#getUuid
+     * @deprecated Replace with {@link Instance#getUuid()}
      */
     @Deprecated(forRemoval = true)
     public @NotNull UUID getUniqueId() {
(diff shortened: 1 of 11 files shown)


src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -750,8 +750,8 @@ public EntityTracker getEntityTracker() {
      *
      * @return the instance unique id
      */
-    public @NotNull UUID getUniqueId() {
-        return uniqueId;
+    public @NotNull UUID getUuid() {
```
mworzala, line 753: I'm down, but I would prefer to keep the old method deprecated in cases like this where it is trivial to maintain backwards compatibility for a grace period.

mworzala (changes requested):

src/main/java/net/minestom/server/instance/Instance.java
Bloeckchengrafik: I added it back but deprecated it

TheMeinerLP: Please add also forRemoval also into deprecation annotation and add Javadoc with replacement!

mworzala (approved):

src/main/java/net/minestom/server/utils/UUIDUtils.java
```diff
@@ -11,7 +11,7 @@
  * An utilities class for {@link UUID}.
  */
 @ApiStatus.Internal
-public final class UniqueIdUtils {
+public final class UUIDUtils {
```
TheMeinerLP, line 14: This is still a breaking change from my opinion

mworzala: I think using a marked internal class accepts a contract that it may change.

Bloeckchengrafik: IMO this does not need to be announced as such since it is internal api

src/main/java/net/minestom/server/utils/UUIDUtils.java
```diff
@@ -20,7 +20,7 @@ public final class UniqueIdUtils {
      * @param input The input string to be checked
      * @return {@code true} if the input an unique identifier, otherwise {@code false}
      */
-    public static boolean isUniqueId(String input) {
+    public static boolean isUuid(String input) {
         return input.matches(UNIQUE_ID_PATTERN.pattern());
```
theEvilReaper, line 24: Can you update the regex usage to utilize a precompiled pattern for better performance, like this:
`UNIQUE_ID_PATTERN.matcher(input).matches();`

This approach avoids recompiling the pattern at runtime. Optionally, it would be great to add a test case to verify whether a UUID is valid or not.

src/main/java/net/minestom/server/utils/UUIDUtils.java
TheMeinerLP: Some PPL use it because they don't understand what Internal means

mworzala: True, but they will learn. I think this is fairly obvious given that the IDE will warn of usages of internal apis, not just looking at the class itself. This is also how we have operated in the past.

src/main/java/net/minestom/server/utils/UUIDUtils.java
Bloeckchengrafik: I agree but I think this should belong in a separate PR. What do @mworzala or @TheMeinerLP think?

PR #2625, opened 2025-01-24
fix: don't recompile uuid pattern on every match

### LINKED ISSUE #2623: Avoid unnessary pattern compile in the UUIDUtils class
The [UUIDUtils](https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/utils/UUIDUtils.java) class contains a RegEx pattern that is used to check uuids. This is used in the `#isUuid` method. Unfortunately, the pattern is called via the `#matches` method of the string class, which means that the pattern is recreated internally for each call instead of using the existing one.

Depending on the pattern that has to be compiled, this can have an impact on the runtime.

The call that Java makes internally via `#matches` of String looks like this:

Image

This can be resolved if the check is performed directly via the pattern.

Bloeckchengrafik:
## Proposed changes

closes #2623 
This is a performance fix and also adds a corresponding test

## Types of changes

## Checklist

merging your code._

- [ ] I have added necessary documentation (if appropriate) - not required

## Further comments

_If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you
did and what alternatives you considered, etc..._

none


TheMeinerLP (approved):
LGTM

Issue #2644, opened 2025-02-02
Inventory swapping


Koboo:
### Discussed in https://github.com/Minestom/Minestom/discussions/2643

Hey folks,

I'm running into a annoying/weird behaviour. 

Here is my description to reproduce:

1. `Player` opens a new top inventory. (like chests, crafter, hopper etc)
2. `Player` clicks an item in the inventory
3. The click on the item creates a new inventory
4. The click calls `player.openInventory(newlyCreatedInventory)` (without closing the first one)
5. `Player` receives the open-request
6. `Player` kinda "overrides" the previous inventory and opens and new inventory (without resetting the cursor position in the open inventory)
7. Minestom doesn't call a `InventoryCloseEvent`

My problem is that I have some close logic in my listener, so that doesn't get executed and I've to check for open inventories, before I'm opening any new inventory.

To fix this I would like to create an `InventorySwapEvent` and check for open inventories on the `Player.openInventory(AbstractInventory)` call. If there is any open inventory and the client opened the inventory, I'll send a new `InventorySwapEvent` with the old and new `AbstractInventory`.

GoldenStack:
`InventoryCloseEvent` already supports getting and setting a new inventory, so personally I think `InventorySwapEvent` can be removed and all usages here can be replaced with `InventoryCloseEvent`, since this would just be adding a bunch of unnecessary complexity.

DasBabyPixel:
> `InventoryCloseEvent` already supports getting and setting a new inventory, so personally I think `InventorySwapEvent` can be removed and all usages here can be replaced with `InventoryCloseEvent`, since this would just be adding a bunch of unnecessary complexity.

oh didn't see that one, you are correct

Koboo:
The problem with the `InventoryCloseEvent` would be, that the `newInventory` field in the event needs to be of type `Inventory` and not `AbstractInventory`. 

To swap the `InventorySwapEvent` with `InventoryCloseEvent`, I would have to:

1. Make the `Inventory newInventory` field in `InventoryCloseEvent` an `AbstractInventory` (breaking change, because of the getter of the event)
2. Change the type `AbstractInventory openInventory` field in the `Player` to type `Inventory`

I would suggest the second solution, since that wouldn't be a breaking change because `Player.getOpenInventory()` only returns and `Player.openInventory(..)` only accepts `AbstractInventory`.

Koboo:
[...]

PR #2645, opened 2025-02-02
feat: Inventory swapping behaviour

Koboo:
## Proposed changes

Adds an InventorySwapEvent, which gets dispatched if the Player.openInventory gets called and there was an already opened inventory. The client will "swap" the new inventory with the previous one, which results in not resetting the cursor position within the inventory.

- Issue: https://github.com/Minestom/Minestom/issues/2644
- Discussion: https://github.com/Minestom/Minestom/discussions/2643

## Types of changes

## Checklist

merging your code._


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1745,7 +1750,12 @@ public boolean openInventory(@NotNull Inventory inventory) {
             newInventory.addViewer(this);
             this.openInventory = newInventory;
         });
-        return !inventoryOpenEvent.isCancelled();
+        boolean wasSuccessfulOpened = !inventoryOpenEvent.isCancelled();
+        if (wasSuccessfulOpened && previouslyOpenInventory != null) {
+            InventorySwapEvent inventorySwapEvent = new InventorySwapEvent(previouslyOpenInventory, this.openInventory, this);
```
DasBabyPixel, line 1755: event is cancellable but cancelled not handled

src/main/java/net/minestom/server/event/inventory/InventorySwapEvent.java
```diff
@@ -0,0 +1,69 @@
...
+    public @NotNull AbstractInventory getOldInventory() {
+        return oldInventory;
+    }
+
+    /**
+     * Gets the new inventory, which was opened by the client.
+     *
+     * @return the new inventory
+     */
+    public AbstractInventory getNewInventory() {
```
DasBabyPixel, line 56: NotNull

src/main/java/net/minestom/server/entity/Player.java
Koboo: Thank you ❤️

src/main/java/net/minestom/server/event/inventory/InventorySwapEvent.java
Koboo: ❤️

DasBabyPixel:
Could also add a test to ensure the InventoryCloseEvent is not called twice. Considering with this PR that would be intended behaviour and is (I think) not tested anywhere

Koboo:
I've added a test, to ensure no close event is fired during the `Player.openInventory(..)` call. 😄

DasBabyPixel (approved):

mworzala:
Currently opening a new inventory does not reset your cursor position which seems to be the major benefit of this PR. So I am assuming I am missing some context? Additionally, you can get the new and old inventory from `InventoryOpenEvent` so this doesn't seem to add anything.

(example included of opening new inventories without the cursor being reset)

https://github.com/user-attachments/assets/e7f4a393-0734-438c-b2aa-5d4b9a468b52

DasBabyPixel:
awkward
but agreed
it's a little unintuitive to get the old inventory via 
```
InventoryOpenEvent event = ...
event.getPlayer().getOpenInventory()
```
Thoughts about adding a method to the InventoryOpenEvent something like getOldInventory or getPreviousInventory which wraps that call?

mworzala:
Yeah I think a getOldInventory would be reasonable there

Koboo:
So if I understand correctly: 
1. Remove `InventorySwapEvent`
2. Add `InventoryOpenEvent.getOldInventory()` wrapper call for `Player.getOpenInventory()`
3. Document the behaviour on `InventoryOpenEvent` and `Player.openInventory(..)` javadocs

I can do that, but I'm in favor of the `InventorySwapEvent` since it's a pretty specific case. For me, a check if `InventoryOpenEvent` is a "swap" or a classic "open" would be pretty annoying and counter intuitive. (_Why would I check for client-side-closing inventories on an `InventoryOpenEvent`?_)

Maybe one more thing we could try:
1. `InventorySwapEvent extends InventoryOpenEvent`
2. Do the check directly in the `openInventory(..)` method 

This way we would still have a specific event, instead of relying on wrapper calls + only the initialization call for the current `InventoryOpenEvent` itself changes. It wouldn't cause any breaking- or behaviour-changes but clearly indicates, that Swapping differs from just Opening.

GoldenStack:
I'm still not convinced that InventorySwapEvent is necessary (although I don't really have any authority so I don't know how much this matters). I feel like at the minimum, a few types could be swapped between Inventory and AbstractInventory (like you mentioned) and that would support all of the functionality that you'd need.

Also, why would you have to check for client-side closing inventories in InventoryOpenEvent?

Koboo:
I think I wrote it a lilttle bit misleading. The question I mentioned is an example question a user could ask himself. 

If we use the InventoryOpenEvent wrapper method for the old inventory, you would have to check for the client-side swapping behaviour inside the open listener like:
```java
public class InventoryOpenListener extends Consumer<InventoryOpenEvent> {
  public void accept(InventoryOpenEvent event) {
    AbstractInventory inventory = event.getOldInventory();
    if(inventory != null) {
      // Inventory will be swapped client-side
    } 
  }
}
```

I think it's a bit weird to check that in an open event, but as you said, I'm not the decision maker either. I'm here to PR whatever is the cleanest way of doing it. 😅

mworzala:
I do see the point about checking on open seeming like an odd place, but im not sure an entirely new event is necessary for this. At least in my experience it's almost never relevant to know if an inventory is being 'swapped' vs close + reopen so delegating to the open event seems ok

Koboo:
I've implemented the delegation method in the `InventoryOpenEvent`, added javadocs to the `player.setInventory(Inventory inventory)` method and updated my Tests. :)

PR #2646, opened 2025-02-02
Rework Entity AI

GreatWyrm:
## Proposed changes
The goal of these changes are to rework how Entity AI is structured, and to provide better options for developers to have more complex AI.
To do this, I've changed a few things
- AIGoals now have their own defined target selectors, rather than target selectors being shared with all goals. Having multiple selectors to cover all AI goals is limiting and may confuse people when certain target selectors pick targets for their goals that they do not intend.
- Target Selectors can now pick locations as well as entities.
- Moved LivingEntity code out of EntityCreature. This may be moved into a separate LivingEntityCreature class in the future.
- Added code to support goals interrupting the current goal (say if a random stroll goal takes the creature in range to attack a nearby entity)
- Added a priority integer for AI Goals. Previously, it would use goals in the order they were added for priority.

## Types of changes

## Checklist

merging your code._

## Further comments

The Entity AI code has been out of date for several versions now, with somewhat confusing links between classes (see a whole private class just to make sure goal selectors had their entity creatures updated), and additional issues (you cannot use a non-living entity for entity creature or you disconnect players with a Network Protocol Error).
This PR aims to modernize our AI code by giving developers more control over how AI Goals are picked and to allow them to customize target selection per goal better.


mworzala (commented):
I havent looked at this in depth but one thing I would like to see in this PR is spitting the navigation logic from the entity subclasses. Entity subclasses have a protocol implication, but we should not enforce that implication on what you may or may not want to give AI to.

I think splitting off a common interface which is simple and delegates most of its logic to another class for processing would be an improvement.

GreatWyrm:
Sure, I can do that as well

Would you mind listing the classes that you'd like to make Entity agnostic? I assume it's the Navigator class and all the path generating/following, anything else specific?

PR #2650, opened 2025-02-04
Extend known packs timeout & make it configurable

mworzala:
## Proposed changes
It came up that when using a continuation in Velocity to wait for a resource pack the existing timeout window can be reached causing the player to be kicked.

This change extends the window to respond to 5m (OK since keep alive exists in config so we can wait). It is also configurable now as a system property (see ServerFlag).

## Types of changes


TheMeinerLP (approved):
LGTM

PR #2651, opened 2025-02-04
Use System.nanoTime() instead of System.currentTimeMillis()

DasBabyPixel:
## Proposed changes

Change use of elapsed time from `System.currentTimeMillis()` to `System.nanoTime()`

## Types of changes

- [x] Bugfix (change which fixes an issue)

## Checklist

merging your code._

## Further comments

- `Cooldown` starts by returning `true` for all #isReady calls until `refreshLastUpdate` is called
- `InstanceContainer#getLastBlockChangeTime()` is now nanos, so it's breaking
Also the `getLastBlockChangeTime()` is set at construction, so logic depending on it being 0 initially behaves differently
- `ItemEntity`'s documentation for `getSpawnTime()` said it returned a time since spawn (delta), which was not the case.
Existing logic using `getSpawnTime()` will not function identically, since I changed it to return a time delta. I also deprecated `getSpawnTime()` for removal in favor of `getTimeSinceSpawn()`. There may be an argument to have those two methods behave differently and getSpawnTime return the actual fixed time (in `System.nanoTime()`), but I'd say that is implementation specific so I'd rather just get rid of `getSpawnTime()` and just keep the delta method
- `Chunk#getLastChangeTime()` removed completely, had no usages and the usage according to docs is outdated anyway. Custom chunk implementations may break (if they don't extend DynamicChunk), otherwise I don't think there is any use of this method at all

### DIFF
--- a/.github/workflows/javadoc.yml
+++ b/.github/workflows/javadoc.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Build and deploy Javadoc
     env:
-      JAVA_VERSION: 23
+      JAVA_VERSION: 24
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK ${{ env.JAVA_VERSION }}
--- a/demo/src/main/java/net/minestom/demo/commands/RelightCommand.java
+++ b/demo/src/main/java/net/minestom/demo/commands/RelightCommand.java
@@ -4,15 +4,17 @@ import net.minestom.server.command.builder.Command;
 import net.minestom.server.entity.Player;
 import net.minestom.server.instance.LightingChunk;
 
+import java.util.concurrent.TimeUnit;
+
 public class RelightCommand extends Command {
     public RelightCommand() {
         super("relight");
         setDefaultExecutor((source, args) -> {
             if (source instanceof Player player) {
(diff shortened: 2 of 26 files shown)


mworzala:
What is the breaking change here? ItemEntity isn't edited

DasBabyPixel:
> What is the breaking change here? ItemEntity isn't edited

still a draft, I'm working locally on it and documenting all the changes here so I don't forget any

DasBabyPixel:
Should, I've gone into demo found no issues, ItemEntity (dropping items in demo) seems to work how it should.
Hope I didn't miss any breaking changes (I updated original PR comment)

DasBabyPixel:
nanoTime should always be used to measure elapsed time, because currentTimeMillis can jump unexpectedly when the OS decides to update the clock time because it is out of sync.
It's not about using millis or nanos (or that nanos are more precise).
I know I left currentTimeMillis in one or two places, where they were actually needed, because they were used to display some real-world time rather than measuring elapsed time.

KermanIsPretty (approved):
I think the idea is that it helps that minestom delta is usually less than a millisecond

mworzala (commented):
To be honest a bunch of this needs to become ticks instead of wall clock time. But this is OK in the mean time.

PR #2653, opened 2025-02-05
Support Entity copies

DasBabyPixel:
## Proposed changes

Add the ability for entities to be copied/cloned. 
The clone will have no assigned instance, but most relevant data of the original entity.
This data includes position, velocity, tags and metadata by default.

## Types of changes

## Checklist

merging your code._

## Further comments

This will be required for saving entities in chunks. When we save a chunk, we usually do that on another thread, so we don't block ticking. This makes any access to the existing Entity instance unsafe, because the instance will tick and the Entity will move, which can cause race conditions and memory visibility errors. To mitigate this, we need to introduce a way to copy/clone an Entity, so we can safely read all required data from the clone.


src/main/java/net/minestom/server/entity/MetadataHolder.java
```diff
@@ -157,6 +157,15 @@ public void setNotifyAboutChanges(boolean notifyAboutChanges) {
         return Map.copyOf(this.entries);
     }
 
+    @ApiStatus.Internal
+    public void copyFrom(MetadataHolder other) {
```
TheMeinerLP, line 161: make it package private

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2087,6 +2087,11 @@ public void addPacketToQueue(@NotNull ClientPacket packet) {
         }
     }
 
+    @Override
+    protected Entity createCopyInstance() {
+        throw new UnsupportedOperationException("Players can't be copied into a separate object");
+    }
```
TheMeinerLP, line 2093: I am not a fan of this. Is it possible to handle that with a additional interface for allow copy entities

src/test/java/net/minestom/server/entity/EntityCopyTest.java
```diff
@@ -0,0 +1,30 @@
...
+import net.minestom.server.coordinate.Pos;
+import net.minestom.server.coordinate.Vec;
+import net.minestom.testing.Env;
+import net.minestom.testing.EnvTest;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@EnvTest
+public class EntityCopyTest {
```
TheMeinerLP, line 13: nitpick: remove public

TheMeinerLP (changes requested):

src/main/java/net/minestom/server/entity/Player.java
DasBabyPixel: If we make that interface then the default Entity class should probably implement it, making the interface pointless?

DasBabyPixel:
Got any idea why the macos PlayerMovementIntegrationTest is failing? I didn't change any existing code lol

src/test/java/net/minestom/server/entity/EntityCopyTest.java
TheMeinerLP: Yes of course, because is Junit5 conform

DasBabyPixel:
now macos was successful 👀 very very interesting...

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -470,6 +470,56 @@ public final boolean removeViewer(@NotNull Player player) {
...
+     * <p>
+     * The position data of the copy will be set, but not the chunk, considering it is not in a chunk anymore.
+     *
+     * @return a copy with all entity data
+     * @throws UnsupportedOperationException if a copy can't be created, for example, because this entity is a player
+     * @see #fillCopyWithData(Entity)
+     * @see #createCopyInstance()
+     */
+    public final Entity createCopy() throws UnsupportedOperationException {
+        Entity copy = createCopyInstance();
```
mworzala, line 487: It should probably be an error if the type of Entity is not the same exact type as `this`

src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -470,6 +470,56 @@ public final boolean removeViewer(@NotNull Player player) {
...
+     *
+     * @param copy the copy to fill with the data of this entity
+     */
+    protected void fillCopyWithData(Entity copy) {
+        copy.tagHandler.updateContent(this.tagHandler.asCompound());
+        copy.setPositionInternal(this.position);
+        copy.onGround = this.onGround;
+        copy.metadata.copyFrom(this.metadata);
+        copy.velocity = this.velocity;
+        // maybe we want more data to be saved in the future, when that is the case we can always extend this
```
mworzala, line 520: Unnecessary comment

src/main/java/net/minestom/server/entity/Player.java
mworzala: Yeah I think requiring people to make a subclass of every other entity before use just so Player doesn't inherit it seems a bit much. Checking for unsupportedoperationexception is fine I suppose.

mworzala (commented):
Overall it seems _ok_ to me. It feels like having a `writeToTag` `readFromTag` method would be in many ways a better way to handle serialization. And without the serialization comment I am far less convinced on this api.

mworzala:
Yeah I think we do need a registry/factory somewhere for making entity instances. Though im not sure where (global? per instance? default global but can be replaced per instance?)

DasBabyPixel:
I believe it is important to have a factory at a global and at an instance level.
So global with instance level overrides is the best option I think

mworzala:
Could just have a global registry which can be replaced with a different one at the instance level (I dont think we need some kind of inheritance here necessarily)

DasBabyPixel:
> Could just have a global registry which can be replaced with a different one at the instance level (I dont think we need some kind of inheritance here necessarily)

Considering we need a getter and a setter on the instance level, I don't like this pattern.
If we create a new instance, call the getter, register a new entity factory, then we'd register that factory to the global registrar, wouldn't we?
Creating a copy of the global registrar at instance creation time would prevent this, but then any changes to the global registrar wouldn't be visible in the instance. Is that what we want?

DasBabyPixel:
@mworzala how would we go about serializing entity metadata? `MetadataHolder` only works with magic values, should we use those, even though they're bound to cause issues deserializing an entity when a new minecraft version comes out?
Personally I'd rather save them by metadata names, like `"NoGravity": true, ...` instead of magic number 5 for no gravity.

DasBabyPixel:
I think we'd need to add a kyori `Key` or `String` for each Entry in MetadataDef, which is a repetitive pain but doable

mworzala:
I think it only ever makes sense to serialize the Minecraft values for metadata (ie in the normal entity format). Making up our own incompatible thing doesn't make sense to me. That said, metadata doesn't map exactly 1:1 with metadata. so I think its not as simple as adding a key and serializer to MetadataDef.

PR #2654, opened 2025-02-05
ChunkSystem

DasBabyPixel:
## Proposed changes

Work in progress ChunkSystem. 
Mostly here right now so golden can stalk me and tell me how he doesn't like this because "this is really complicated i don't think it's justified".
(this is like the 4th iteration and I think it's not even that complex anymore. Was way worse in the beginning)

 - The chunk system will run in a fully synchronous manner in tests, unless async mode is explicitly requested. This is required for most tests to pass, because otherwise a test that requires a player object can already fail from a race condition.

View [Chunk System](https://discord.com/channels/706185253441634317/1193029605842485248) thread on discord for more information, I'm gonna update this description as things go on.

### Lighting
This is about how lighting is supposed to work.
Putting everything into this PR description was a bit too much, so I made a gist which I will update instead
https://gist.github.com/DasBabyPixel/d65dc63a2a213b79e592ff9c55ccd7cd

### TODO
 - [ ] Add back `Instance#generateChunk(int, int, Generator)`

## Notes
 - Logic in tests is executed synchronously. This includes chunk loading/unloading, etc. If we were to use async loading for tests, every single test would have to be updated to execute ticking logic and wait for async chunks to load. This is possible, but will change A LOT of test logic and make testing in general much more tedious.

## Types of changes

- [x] New feature (non-breaking change which adds functionality) Hella breaking but whatever

## Checklist

merging your code._

## Further comments
Fixes #2676
If you have thoughts/considerations please do share.


src/main/java/net/minestom/server/instance/ChunkSystemChunkAccessImpl.java
```diff
@@ -0,0 +1,20 @@
+package net.minestom.server.instance;
+
+import net.minestom.server.instance.chunksystem.impl.ChunkAccess;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * <b>For internal use only, may be changed at any time. Do not use this.</b>
+ */
+@ApiStatus.Internal
+public final class ChunkSystemChunkAccessImpl implements ChunkAccess {
```
mworzala, line 10: Why does this need to exist? This seems VERY gross

src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -218,6 +220,10 @@ public boolean breakBlock(@NotNull Player player, @NotNull Point blockPosition,
     @ApiStatus.Internal
     public abstract boolean breakBlock(@NotNull Player player, @NotNull Point blockPosition, @NotNull BlockFace blockFace, boolean doBlockUpdates);
 
+    public @NotNull ChunkManager getChunkManager() {
```
mworzala, line 223: Should just be abstract

src/main/java/net/minestom/server/instance/chunksystem/ChunkClaim.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public record ChunkClaim(int radius, int priority, @NotNull Shape shape) {
+    @ApiStatus.Internal
+    public ChunkClaim {
+    }
+
+    /**
+     * The shape of a claim. Only matters if radius > 0
+     */
+    public enum Shape {
```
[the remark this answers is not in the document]
mworzala: Also it seems like it doesnt need to be sealed? Is there a reason its a problem for users to implement their own if they have weird use cases

src/main/java/net/minestom/server/instance/chunksystem/ChunkManager.java
```diff
@@ -0,0 +1,221 @@
...
+    boolean isAutosaveEnabled();
+
+    /**
+     * Change whether the {@link ChunkManager} should auto-save unloaded chunks and instance data to storage.
+     * This removes the necessity of calling {@link #saveInstanceDataAndChunks()}.
+     * This setting should always be preferred over {@link #saveChunk(Chunk)} and similar methods.
+     *
+     * @param autosaveEnabled whether autosave is enabled
+     */
+    void setAutosaveEnabled(boolean autosaveEnabled);
```
mworzala, line 48: Im not sure I like coupling automatically unloading chunks to saving them. There are use cases to automatically unload but not save.

src/main/java/net/minestom/server/instance/chunksystem/ChunkClaim.java
```diff
@@ -0,0 +1,74 @@
...
+ * A chunk can have multiple identical {@link ChunkClaim}s, so that {@code claim1.equals(claim2)} equals {@code true}.
+ * In case of claim removal, only one {@link ChunkClaim} will be removed from the chunk (per call).
+ * <p>
+ * The {@link Chunk} reference is not in this class, but rather in {@link ChunkAndClaim}.
+ *
+ * @param radius   the radius of this {@link ChunkClaim}. Use 0 for a single chunk.
+ * @param priority the priority of this {@link ChunkClaim}. Higher priorities are loaded first.
+ * @param shape    the shape used for this {@link ChunkClaim}. {@link net.minestom.server.entity.Player Players} may want to use {@link Shape#CIRCLE}
+ */
+public record ChunkClaim(int radius, int priority, @NotNull Shape shape) {
```
mworzala, line 23: Probably should be an interface and hide its details. Nobody should be instantiating this themselves, and having to comment later about it being a record but instance matters is weird. People just shouldn't be able to see the impl that matters

src/main/java/net/minestom/server/instance/chunksystem/ChunkManager.java
```diff
@@ -0,0 +1,221 @@
...
+     * <p>
+     * Warning: only the global instance data will be saved, not chunks.
+     * You would need to call {@link #saveChunks()} too, or instead opt for {@link #saveInstanceDataAndChunks()}
+     *
+     * @return the future called once the instance data has been saved
+     * @deprecated try to use {@link #setAutosaveEnabled(boolean)} instead, if possible.
+     * If not, please state your use case and open an issue on GitHub
+     */
+    @Deprecated
+    @NotNull CompletableFuture<Void> saveInstanceData();
```
mworzala, line 154: Why are we adding a new api with deprecated methods 🙃

src/main/java/net/minestom/server/instance/chunksystem/impl/ChunkAccess.java
```diff
@@ -0,0 +1,14 @@
...
+import net.minestom.server.instance.Chunk;
+import org.jetbrains.annotations.ApiStatus;
+
+/**
+ * <b>For internal use only, may be changed at any time. Do not use this.</b>
+ * <p>
+ * An interface for the chunk system to access protected methods of some classes like {@link Chunk}
+ */
+@ApiStatus.Internal
+public interface ChunkAccess {
```
mworzala, line 12: Follow up to the above comment - This needs to go. this does nothing but scream terrible api.

mworzala (commented):
I havent looked at the impl details very much yet ill come back for those

DasBabyPixel:
@mworzala relax this is still a WIP and doesn't even function yet. Regarding the ChunkAccess I fully intend to remove that later I just wanted to change as little as possible in existing code for now to make rebasing/dropping commits easier.
Nowhere near ready for review as of now 😂

src/main/java/net/minestom/server/instance/chunksystem/ChunkClaim.java
DasBabyPixel: 👍 gonna do later

src/main/java/net/minestom/server/instance/chunksystem/ChunkManager.java
DasBabyPixel: wait wat?
You can tell the system to only unload chunks, but not save them, and you can tell it to save chunks when unloading.
I think I don't get your point

src/main/java/net/minestom/server/instance/chunksystem/ChunkManager.java
DasBabyPixel: I needed some way to indicate that enabling autosave is the better option if people just want stuff to work. Could be marked differently, but I think this should be marked in some way.

src/test/java/net/minestom/server/tag/TagStructureTest.java
```diff
@@ -12,7 +12,7 @@
 
 public class TagStructureTest {
 
-    private static final Tag<Entry> STRUCTURE_TAG = Tag.Structure("entry", new TagSerializer<>() {
+    private static final Tag<Entry> STRUCTURE_TAG = Tag.Structure("node", new TagSerializer<>() {
```
DasBabyPixel, line 15: Note to self: when did this happen lol probably some IDEA rename stuff going on gonna have to undo this

DasBabyPixel:
> Just trying to help 🥺

sowwy... Glad you wanna help 👍 
Just wanted to say that your time as of now is spent better reviewing other PRs, this is still gonna change in many aspects.
I'm gonna comment when this is closer to completion (when I have the first working version probably)

src/main/java/net/minestom/server/ServerFlag.java
```diff
@@ -71,6 +71,17 @@ public final class ServerFlag {
     public static final boolean REGISTRY_LATE_REGISTER = booleanProperty("minestom.registry.late-register");
     public static final boolean REGISTRY_UNSAFE_OPS = booleanProperty("minestom.registry.unsafe-ops");
     public static final boolean EVENT_NODE_ALLOW_MULTIPLE_PARENTS = booleanProperty("minestom.event.multiple-parents");
+    /**
+     * Valid values:
+     * <ul>
+     *     <li>simple (fastest, inaccurate)</li>
```
GoldenStack, line 77: Maybe you could specify that this is [taxicab distance](https://en.wikipedia.org/wiki/Taxicab_geometry), just for clarity

src/main/java/net/minestom/server/instance/chunksystem/ChunkClaim.java
```diff
@@ -0,0 +1,74 @@
...
+        private final boolean wantSquared;
+
+        Shape(boolean wantSquared) {
+            this.wantSquared = wantSquared;
+        }
+
+        @ApiStatus.Internal
+        public boolean doesWantSquared() {
+            return wantSquared;
+        }
```
GoldenStack, line 69: is there a reason this exists instead of just squaring the value inside the function? if removed I feel like this API would be a lot cleaner

DasBabyPixel: bad habits and premature optimization gonna change Shape to an interface later and remove this

kermandev (commented):
I have reviewed some of the pr; I know its a draft so do whatever seems best. But here is some main ideas.

Also, I dont like this whole tick0 thing. I feel that should be the otherway around for a private to internal method.

src/main/java/net/minestom/server/instance/chunksystem/ChunkClaimImpl.java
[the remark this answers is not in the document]
DasBabyPixel: Yeah I know. The point of this was that record are usually viewed as not having an identity, they are meant to be serialized and used with equals/hashCode methods. 
Considering people could store their claims in a HashMap or a similar data structure, I think it is best to keep this the way it is

src/main/java/net/minestom/server/instance/chunksystem/ChunkManager.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Removes a claim from a chunk.
+     * <p>
+     * The passed {@link ChunkClaim} must be the exact same object instance returned when the claim was created.
+     * Even though {@link ChunkClaim} is a record, the object instance matters to identify the exact claim.
+     *
+     * @param claim the {@link ChunkClaim} that should be removed
+     * @return a future for when the claim was removed.
+     */
+    @NotNull CompletableFuture<Void> removeClaim(@NotNull ChunkClaim claim);
```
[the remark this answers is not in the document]
DasBabyPixel: How would you manage the control over the claim though?
If the user is allowed to create a claim object, then the user can add the same claim to multiple chunks.
The user could even add the same claim to the same chunk multiple times.
If we allow that, we make removing claims and validating correct usage infinitely more difficult

src/main/java/net/minestom/server/instance/chunksystem/ChunkAndClaim.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Holder data structure containing a claim and a chunk.
+ *
+ * @param chunkFuture the chunk future which will be completed when the main chunk has finished loading
+ * @param chunkClaim  the claim
+ */
+public record ChunkAndClaim(@NotNull CompletableFuture<@NotNull Chunk> chunkFuture, @NotNull ChunkClaim chunkClaim) {
```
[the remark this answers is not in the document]
DasBabyPixel: We need this information to access the created claim and the loaded chunk for the center chunk of that claim. So we need both

src/main/java/net/minestom/server/instance/chunksystem/PriorityDrop.java
```diff
@@ -0,0 +1,49 @@
+package net.minestom.server.instance.chunksystem;
+
+/**
+ * Used to calculate how much the priority should drop for a chunk with a given distance from the center of the claim.
+ * This is used to make closer chunks load first.
+ */
+sealed interface PriorityDrop {
```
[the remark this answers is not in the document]
DasBabyPixel: I can do this, but custom drop-offs are not the easiest to implement correctly. Because we need to propagate updates down the line, currently every chunk (except the claim origin) must have a neighbor with a higher priority. Otherwise chunks are not gonna be updated correctly.

Maybe gonna change and document this behavior later, but not rn.

src/main/java/net/minestom/server/instance/chunksystem/TaskSchedulerThread.java
```diff
@@ -0,0 +1,507 @@
...
+        final CompletableFuture<Void> future;
+        final CompletableFuture<Void> partitionDeleted = new CompletableFuture<>();
+
+        UnloadTask(@NotNull Chunk chunk, CompletableFuture<Void> future) {
+            this.chunk = chunk;
+            this.future = future;
+        }
+    }
+
+    record SaveTask(@NotNull Chunk chunk, @NotNull CompletableFuture<Void> future) {
```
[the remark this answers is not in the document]
DasBabyPixel: cause it's not a content of Task, required for the SingleThreadedManager. Gonna move this, nice catch

src/main/java/net/minestom/server/instance/chunksystem/TaskSchedulerThread.java
```diff
@@ -0,0 +1,507 @@
...
+     *
+     * @return the future for when shutdown has completed
+     */
+    private CompletableFuture<?> shutdown() {
+        this.exit = true;
+        this.signaling.signal();
+        return this.shutdownFuture;
+    }
+
+    static final class LoadTask {
```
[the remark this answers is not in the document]
DasBabyPixel: same thing

src/main/java/net/minestom/server/instance/chunksystem/TaskSchedulerThread.java
```diff
@@ -0,0 +1,507 @@
...
+        });
+    }
+
+    private void start() {
+        // We finally start ourselves.
+        // Now we start accepting tasks.
+        // Virtual thread is appropriate here, we
+        // will spend most time waiting for signals/workers.
+        // Also, 1 platform thread per instance can become quite expensive
+        Thread.ofPlatform().name("ChunkTaskScheduler-" + id).start(this);
```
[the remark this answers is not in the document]
DasBabyPixel: Nice catch. I gotta change this back to virtual threads. I was dealing with tons of deadlocks so I changed this temporarily to platform threads to get some stack traces in the debugger

src/main/java/net/minestom/server/instance/chunksystem/ChunkWorker.java
```diff
@@ -0,0 +1,238 @@
...
+     */
+    private static ExecutorService WORKER_EXECUTOR;
+    private static ExecutorService SAVE_EXECUTOR;
+    /**
+     * We allow twice the number of available processors to be submitted before waiting.
+     * This is so we don't waste time.
+     * As soon as a task finishes, the thread will be able to take the next task,
+     * which has already been submitted.
+     */
+    private static final Semaphore AVAILABLE_TASKS = new Semaphore(Runtime.getRuntime().availableProcessors() * 2);
```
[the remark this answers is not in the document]
DasBabyPixel: This can't be done using virtual threads, because we propagate chunk updates lazily, that is when it makes sense to do so, because there is a worker free (or soon to be free).
If we were to use virtual threads and wait for a worker in there, we'd update all chunks at the same time, removing the laziness of the system, making it almost identical to what we had before

src/main/java/net/minestom/server/instance/chunksystem/TaskSchedulerThread.java
```diff
@@ -0,0 +1,507 @@
...
+        case null, default -> new PriorityDrop.HypotenuseSquared();
+    };
+
+    // Needed to make this reusable.
+    private final ReentrantLock mainLock = new ReentrantLock();
+    private CompletableFuture<Void> shutdownFuture = new CompletableFuture<>();
+
+    private final ReentrantLock singleThreadedManagerLock = new ReentrantLock();
+    private final SingleThreadedManager singleThreadedManager;
+    private final int id = ThreadLocalRandom.current().nextInt();
```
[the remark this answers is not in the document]
DasBabyPixel: true, was just for debugging and this was faster than adding another field. I'm gonna remove it

src/main/java/net/minestom/server/instance/chunksystem/TaskSchedulerThread.java
```diff
@@ -0,0 +1,507 @@
...
+            this.loader = loader;
+            this.chunkSupplier = chunkSupplier;
+            this.generator = generator;
+            this.x = x;
+            this.z = z;
+            this.lastUpdatePriority = lastUpdatePriority;
+        }
+    }
+
+    static final class UnloadTask {
```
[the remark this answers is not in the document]
DasBabyPixel: same thing as above again

src/main/java/net/minestom/server/instance/chunksystem/ChunkClaim.java
```diff
@@ -0,0 +1,101 @@
...
+     * {@link net.minestom.server.entity.Player Players} would use {@link Shape#CIRCLE}
+     *
+     * @return the shape of this claim
+     */
+    @NotNull Shape shape();
+
+    /**
+     * The shape of a claim. Only matters if radius > 0
+     */
+    interface Shape {
```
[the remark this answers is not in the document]
DasBabyPixel: eh you got a point there

DasBabyPixel:
> I have reviewed some of the pr; I know its a draft so do whatever seems best. But here is some main ideas.
> 
> Also, I dont like this whole tick0 thing. I feel that should be the otherway around for a private to internal method.

I'm also not a huge fan of tick0, but tick is already defined in the Tickable interface as the method that is called to tick something.
I thought it is better to introduce that tick0 instead of tick not ticking correctly, because it doesn't process scheduler tasks

DasBabyPixel:
@mworzala please don't rebase this, I'll do it myself. Thx ❤️

src/main/java/net/minestom/server/instance/chunksystem/ChunkManager.java
DasBabyPixel: Re-reading this now makes me think this is resolved. 
Chunks are always unloaded automatically, unless a claim is added which keeps them loaded. 
Auto-Saving is optional. It can be turned on, so all chunks will be saved when unloaded. It can be turned off so now chunks will be saved when unloaded.

kermandev:

PR #2655, opened 2025-02-06
Entity Query API

### LINKED ISSUE #2649: [Bug]: EntityTracker leaking memory
### Affected Version

https://github.com/Minestom/Minestom/commit/87f6524aeb6ea4f9ce8eabad6c32170a300b894b (master)
https://github.com/Minestom/Minestom/commit/e13a03a8313b6386d8811d289d845652fd6aac51 (1_21_4)

### Describe the bug

The EntityTracker is currently leaking memory by keeping references by chunk index.
With infinitely large world generation this will add up over time

Image

I propose a fix by using a ReadWriteLock and removing the list from the hashmap once it is empty

Put a breakpoint after the unregister method (code sample)

### Code sample

```
@Test
public void clearing() {
    var ent1 = new Entity(EntityType.ZOMBIE);

    EntityTracker tracker = EntityTracker.newTracker();
    tracker.register(ent1, Vec.ZERO, EntityTracker.Target.ENTITIES, null);
    tracker.unregister(ent1, EntityTracker.Target.ENTITIES, null);
}
```

TheMode:
## Proposed changes

Add a new entity querying API, replacing the specific retrieval methods and encouraging batching.

Breaking change includes the inability to get view collections. I believe this is acceptable, may add new methods taking an array as parameter for the tracker to append entities inside instead of the `Stream` as a higher performance alternative.
Should fix #2649

Should probably delete the existing `EntityFinder`. Although I did not figure out yet how to handle the `SELF` target selector.
Hesitating to remove `Instance#getEntities/Players` as they force a copy, but I guess they could stay as slow convenience.

Eventually, querying could include `Tag` checks.

`EntityTracker` could at some point take indexing hints, for example to disable spatial partitioning entirely (useful on small maps with few entities) or to more properly distinguish entity types.

Welcome any `Query` API change, as I do not feel too strongly about it, and I can agree that it sometimes look a bit bloaty.

## Types of changes

## Checklist

merging your code._

## Further comments

Did not profile it yet.


src/main/java/net/minestom/server/entity/EntityQuery.java
```diff
@@ -0,0 +1,169 @@
...
+    Property<Double> POS_Y = new Index<>();
+    Property<Double> POS_Z = new Index<>();
+    Property<Float> YAW = new Index<>();
+    Property<Float> PITCH = new Index<>();
+    Property<Integer> CHUNK_X = new Index<>();
+    Property<Integer> CHUNK_Z = new Index<>();
+    Property<Double> DISTANCE = new Index<>();
+    Property<EntityType> TYPE = new Index<>();
+    Property<GameMode> GAME_MODE = new Index<>();
+    Property<Integer> EXPERIENCE = new Index<>();
```
mworzala, line 37: Users own properties would be helpful here

src/main/java/net/minestom/server/entity/EntityQuery.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        /**
+         * Allows any predicate which MUST be pure.
+         * No guarantee as to when called, and may not take advantage of future optimizations.
+         * <p>
+         * Use at your own risk.
+         */
+        @ApiStatus.Internal
+        static <T> Condition<T> __condition(Property<T> property, BiPredicate<Point, T> predicate) {
```
mworzala, line 121: should be named normally (`condition`, `userCondition`, `customCondition`?) + Experimental is probably better than Internal

src/main/java/net/minestom/server/entity/EntityView.java
```diff
@@ -232,8 +235,13 @@ private Collection<T> references() {
             final Point point = trackedLocation.point();
 
             Int2ObjectOpenHashMap<T> entityMap = new Int2ObjectOpenHashMap<>(lastSize);
-            instance.getEntityTracker().nearbyEntitiesByChunkRange(point, RANGE, target,
-                    (entity) -> entityMap.putIfAbsent(entity.getEntityId(), entity));
+            final EntityQuery nearbyQuery = players ? entityQuery(
```
mworzala, line 238: It's just a syntactical change but I wonder if it's worth ignoring null conditions in `entityQuery`? It would allow writing

```java
          final EntityQuery nearbyQuery = entityQuery(
                    players ? equalsCondition(EntityQuery.PLAYER, true) : null,
                    chunkRangeCondition(RANGE)
            );
```

which to me seems a little easier to read

src/main/java/net/minestom/server/entity/EntityQuery.java
mworzala: Can leave it up to users to update probably (agreed on the safety issue but it's just user error at that point)

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
```diff
@@ -1,136 +1,127 @@
...
+    // Spatial partitioning
+    private final Long2ObjectMap<Set<Entity>> chunksEntities = new Long2ObjectOpenHashMap<>();
 
     @Override
-    public <T extends Entity> void register(@NotNull Entity entity, @NotNull Point point,
-                                            @NotNull Target<T> target, @Nullable Update<T> update) {
-        EntityTrackerEntry newEntry = new EntityTrackerEntry(entity, point);
-
-        EntityTrackerEntry prevEntryWithId = entriesByEntityId.putIfAbsent(entity.getEntityId(), newEntry);
+    public synchronized void register(@NotNull Entity entity, @NotNull Point point, @Nullable Update update) {
```
DasBabyPixel, line 40: a consideration: replace synchronized with locks to avoid pinning. JDK 24 fixes this, but until minestom forces java 24 I think this should be a lock

KermanIsPretty: That only matters if you call with a move with a virtual thread, which is out of scope for entities.

DasBabyPixel: true

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
```diff
@@ -139,217 +130,103 @@ public void remove(@NotNull T removed) {
...
-            // Single chunk
-            final var chunkEntities = (List<T>) entities.get(CoordConversion.chunkIndex(point));
-            if (chunkEntities != null && !chunkEntities.isEmpty()) {
-                chunkEntities.forEach(entity -> {
-                    final Point position = entriesByEntityId.get(entity.getEntityId()).getLastPosition();
-                    if (point.distanceSquared(position) <= squaredRange) query.accept(entity);
-                });
+            case NEAREST_PLAYER -> {
+                final TrackedEntity nearest = findNearest(origin, true);
+                yield nearest != null ? Stream.of(nearest) : Stream.empty();
```
ThatGravyBoat, line 143: the nearest check should be done after the filter shouldnt it? Atleast in the command context, as thats how Mojang does it with their vanilla commands as you can filter out the closest entity and you then need to get the second closest or else it will always be empty.

DasBabyPixel:
Is there something holding this up or is this PR scrapped? I kinda need to know this for #2654, especially the `Chunk implements Viewable` is interesting to me.
Since the player object itself tracks whether it can see a chunk or not, so we can't just use an arbitrary range/radius collector.

PR #2666, opened 2025-02-23
Add new BlockEvent Handling

DersWasTaken:
## Proposed changes

Add instance events to expose the native setBlock(x,y,z,block) method to be changed by users through an event.

Add BlockEvent.Source to better communicate the source of block events through a sealed interface union allowing for more expansion in the future (i.e BlockEvent.Source.Explosion or similar). Rename

Change how BlockPlace and BlockBreak events are called by having a PreEvent and PostEvent for each & rename BlockPlace to BlockSet for clarity. This allows the block to be modified before it is changed (reducing calls to Instance#setBlock, avoiding infinite loops) and have a separate event intended for monitoring the event after pre-place changes are made. PostEvent generally should not be making changes to the block as that is what PreEvent is intended for.

Deprecate old PlayerBlockBreakEvent & PlayerBlockPlaceEvent in favor of new events while keeping their functionality intact to prevent breaking changes.

## Types of changes

## Checklist

merging your code._

## Further comments

This solution aims to expose non player block changes and create a more unified system for handling changes. Additionally it is necessary for a future PR I am working on that will combine BlockManager & BlockPlacementRules into a single unified event driven system for clarity and better code structure, as the current system is outdated, clunky, inconsistent, and difficult to work with at a large scale.

P.S Sorry for the auto closed PR's, I am bad at GitHub.


DersWasTaken:
Unsure how to fix Merging is Blocked issue. If someone can let me know I will do so.

DersWasTaken:
Everything seems to work fine on my end. I am unsure what tests would be good to add for this if any though.

Let me know if any changes are needed

utfunderscore:
Is there any performance impact when large amounts of blocks are changed? e.g block batches

DersWasTaken:
I can check it in a bit but I would be very surprised if it wasn't on the order of literal nanoseconds at most. Minestom already handles tick events which run *alot*. Especially since only BlockSetPreEvent and BlockSetPostEvent would be triggered once (no player sources) because block batches wouldn't use player placement, which is only two more events

DersWasTaken:
Here's the results I've got:

Parameters:

@Warmup(iterations = 50, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 100, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Fork(3)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)

Benchmark                                                      Mode  Cnt   Score   Error  Units
SetBlockBenchmark.playerPlaceBlockNew  avgt  300  27.931 ± 0.452  ns/op (1 Event Sent)
SetBlockBenchmark.playerPlaceBlockOld    avgt  300   6.161 ± 0.266  ns/op (5 Events Sent)

It seems to scale about as you would expect, with every additional event sent adding about 6ns, which I would argue is very little. 

Admittedly I haven't benchmarked things before, so I am posting my code I used to benchmark this below:

https://pastebin.com/zS8cDsr6

DersWasTaken:
Setting blocks for an entire chunk per operation (16 * 16 * 320), with block updates on

Benchmark                                        Mode  Cnt    Score          Error      Units
SetBlockBenchmark.setBlockNew  avgt     60      195.866    ± 0.693   ms/op
SetBlockBenchmark.setBlockOld    avgt     60      194.791     ± 0.676   ms/op

Dividing that's about 1.99ms per block for new and 1.98 for old, or a difference of about 0.01ms in total from old to new.

Code for benchmark:

`java
@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 20, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Fork(3)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
public class SetBlockBenchmark {

    InstanceContainer container;

    @Setup
    public void setup() {
        MinecraftServer server = MinecraftServer.init();

        GlobalEventHandler globalEventHandler = new GlobalEventHandler();

        container = MinecraftServer.getInstanceManager().createInstanceContainer();
        container.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.GRASS_BLOCK));

        globalEventHandler.addListener(PlayerBlockPlaceEvent.class, playerBlockPlaceEvent -> {});
        globalEventHandler.addListener(PreSetBlockEvent.class, playerBlockPlaceEvent -> {});
        globalEventHandler.addListener(PostSetBlockEvent.class, playerBlockPlaceEvent -> {});

    }

    @Benchmark
    public void setBlockOld() {
        for (int x = 0; x < 16; x++) {
            for (int y = -64; y < 320; y++) {
                for (int z = 0; z < 16; z++) {
                    container.setBlock(x, y, z, Block.DIAMOND_BLOCK, true);
                }
            }
        }
    }

    @TearDown
    public void tearDown(Blackhole bh) {
        bh.consume(container);
    }
}`

DersWasTaken:
Should note that for just purely calling the events, it stacks linearly (i.e about 4x slower calling 4x the events)

Benchmark                                                     Mode Cnt Score    Error      Units
SetBlockBenchmark.playerPlaceBlockNew avgt  300 27.931   ± 0.452 ns/op (5 Events Sent)
SetBlockBenchmark.playerPlaceBlockOld   avgt  300 6.161     ± 0.266 ns/op (1 Event Sent)

src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -150,6 +152,26 @@ private synchronized void UNSAFE_setBlock(@NotNull Chunk chunk, int x, int y, in
             return;
         }
 
+        BlockEvent.Source source = new BlockEvent.Source.Instance(this);
+
+        if(block.isAir()) {
```
mworzala, line 157: formatting issues, run ij formatter on this,

src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -150,6 +152,26 @@ private synchronized void UNSAFE_setBlock(@NotNull Chunk chunk, int x, int y, in
...
+            PreBreakBlockEvent preBreakBlockEvent = new PreBreakBlockEvent(
+                getBlock(x,y,z), srcInstance, null, new BlockVec(x,y,z), source
+            );
+
+            EventDispatcher.call(preBreakBlockEvent);
```
mworzala, line 162: event should not be cancellable if cancelling it does nothing  - Same comment to the other 3 events

mworzala (commented):
Some minor comments above, but more generally im not sure what exactly this PR is aiming to do. Please give the more concrete use case for what the existing API doesnt solve but this does.

src/main/java/net/minestom/server/instance/InstanceContainer.java
DersWasTaken: Ran, although unsure exactly what configuration options I was supposed to use. Hopefully this is okay.

src/main/java/net/minestom/server/instance/InstanceContainer.java
DersWasTaken: Fixed except for Post-Events, see conversation below.

DersWasTaken:
@mworzala 

The purpose of the PR is to expose the ability to modify all changes to an instances blocks using Instance#setBlock, as previously only changes made by a player would call block events (i.e PlayerPlaceBlockEvent from player placing a block).

This allows developers to have access to any changes made to an instances blocks, which is useful for normal developers in multiple scenarios, for example when adding blockHandlers to blocks placed, as you need to manually add the handler to them when they are changed. Before you couldn't do this when the Instance#setBlock was called unless you had control over the Instance#setBlock call.

On reflection though, I do notice the pointlessness of PostSetBlock & PostDestroyBlock events, and I would be fine with removing them as in hindsight the purpose they would serve would be limited and available in other ways.

DersWasTaken:
Would it be a good idea to rebase this onto the 1.21.4 branch?

cody-quinn:
I think `BreakBlockEvent` should be renamed to `BlockBreakEvent` to better match other event names in Minestom. Same with `SetBlockEvent`.

DersWasTaken:
Also if this is merged to the 1.21.4 instead I think we should consider having this as a breaking change with full removal of the currently deprecated events as the actual change the user would need to make is minor (convert it to new events, use Event.getSource() instanceof BlockEvent.Source.Player

DersWasTaken:
Closing to make a better implementation as I'm not happy with this one.

PR #2674, opened 2025-03-06
Send AcknowledgeBlockChangePacket in all cases

DasBabyPixel:
## Proposed changes

Use try-finally to send AcknowledgeBlockChangePacket in all cases.

## Types of changes

## Checklist

merging your code._

## Further comments

I spent the entire day searching for the reason my blocks weren't updating.

Following code pattern is (was, after this PR) illegal:
```
.addListener(PlayerBlockPlaceEvent.class, event -> {
                event.setCancelled(true);
                event.getInstance().setBlock(event.getBlockPosition(), Block.STONE.withHandler(h));
            });
```

The `setBlock` call can also be at a later point in time, it still doesn't update on the client, because the client thinks it has the correct state, because we never said "hey, we got your packet".

To make sure this doesn't happen again (when one of these methods gets changed) I used try-finally. Also makes things easier with return statements.

### DIFF
--- a/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
+++ b/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
@@ -41,147 +41,147 @@ public class BlockPlacementListener {
         final BlockFace blockFace = packet.blockFace();
         Point blockPosition = packet.blockPosition();
 
-        final Instance instance = player.getInstance();
-        if (instance == null)
-            return;
+        try {
+            final Instance instance = player.getInstance();
+            if (instance == null)
+                return;
 
-        // Prevent outdated/modified client data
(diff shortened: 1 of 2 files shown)


DersWasTaken:
Is this not a duplicate of https://github.com/Minestom/Minestom/pull/2599

DasBabyPixel:
> Is this not a duplicate of #2599

Nope this is regarding packets. No (server) logic is changed at all, just acknowledge packets are sent correctly to the client

DersWasTaken:
Hmm, okay. Also I'm having trouble understanding what exactly the point of cancelling the block placement only to then set it a few seconds later is, could you explain? I get it's a demonstration but I'd imagine you would either want to use a block handler or just change the block directly instead of cancelling the event?

PR #2675, opened 2025-03-08
WeakRef all Entity targets in EntityMeta

KermanIsPretty:
## Proposed changes

It should allow entities to be garbage collected if they are no longer in the entity tracker but are a target/shooter/addition.

### Reason
For example, if you shoot an arrow as a player, you can always get a player object for as long as the arrow exists. This is a bad idea if the player left and is deregistered, and now you would have to check this edge case.

## Types of changes

## Checklist

merging your code._

## Further comments
It is a side effect of adding some comments and abstracting projectiles a bit more. Trident projectiles now have a shooter, which is not part of the protocol.

It's probably a breaking change if people usually get targets with deregistered entities, but that's not the case you should be doing.

### DIFF
--- a/src/main/java/net/minestom/server/entity/EntityProjectile.java
+++ b/src/main/java/net/minestom/server/entity/EntityProjectile.java
@@ -14,6 +14,7 @@ import net.minestom.server.instance.Chunk;
 import net.minestom.server.instance.Instance;
 import net.minestom.server.instance.block.Block;
 import net.minestom.server.thread.Acquirable;
+import net.minestom.server.utils.validate.Check;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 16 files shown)


DasBabyPixel:
I feel like the reason you listed is not valid. Regardless, you still have to check the edge case. The garbage collector won't run immediately, so the reference will in many cases still be a player object, with the associated player being offline.

KermanIsPretty:
You still have to check if they are online, but assuming your arrow persists forever (Maybe you don't want them to despawn after a minute when stuck in a block), you will have entity objects that are never garbage collected after the entity has been removed from the entity tracker because it is still referenced in that entity object.  

The entity tracker should source all hard references for entities or a data entity for multi-entity structures. This PR could be extended to remove the weak reference to the entity object of the constructor of EntityMeta, as it is stored in the entity class itself.

DasBabyPixel:
I think a better way to handle this is to make shooter non-final and have a setter. Then the user can decide if he wants to have persistent arrows. I feel like this is a very rare and specific use case, and doesn't justify the inconsistent behavior that comes with weak references

DasBabyPixel:
also I just realized you missed the `shooter` field in `EntityProjectile`

KermanIsPretty:
Good catch it seems EntityProjectile shooter should probably not exist since its entity meta holds it.

DasBabyPixel:
I mean Idk I still don't like things being weak references. It's bound to introduce undetectable/unreproducable/undebuggable problems

mworzala:
I think im with DasBabyPixel on this, it seems like taking a way some possibly unexpected (?) behavior and replacing it with new possibly unexpected behavior. its not obvious that its a weak reference and people may not be expecting the shooter to 'randomly' disappear.

Overall I'm not convinced its Minestoms fault to prevent people from leaking entity references.

PR #2680, opened 2025-03-11
rework/blockhandler

DersWasTaken:
## Proposed changes

The purpose of this PR would be to move BlockHandler to a event driven system & to entirely remove BlockPlacementRules.
Additionally, the functionality of BlockPlacementRules would be merged into BlockHandler, allowing users to continue to user Block#withHandler when they want to set a specific block handler to take effect, and use BlockManager#registerHandler to assign a BlockHandler to all instances of a block.

This is done by building off #2666 and implementing further changes and improvements. 

Firstly, PlayerBlockEvent's are completely removed in this PR, instead being merged into BlockChangeEvent which represents all changes to blocks, both done by the player and done by the server via Instance#setBlock. This aims to reduce repetitive calls to events that happened in #2666 by having a singular event represent all changes, where BlockEvent.Source represents the "source" of the changes (i.e Instance or Player) where each type of source contains additional data to the event.

Since this is a breaking change, I tried to make the transition as smooth as possible:

```java
MinecraftServer.getGlobalEventHandler().addListener(BlockChangeEvent.class, event -> {
    BlockEvent.Source source = event.getSource(); // new Source type
    if(source instance BlockEvent.Source.Instance) {
       // The source comes from the instance via Instance#setBlock.
    }
    if(source instanceof BlockEvent.Source.Player) {
      // The source comes from a player action (i.e Breaking Block, Placing Block) comes with the following data:
      source.getPlayer() //Never Null
      source.getBlockFace() //May be null
      source.getCursorPosition() //May be null
      source.getHand() //May be null
    }
});
```

And since BlockChangeEvent represents both breaking and placing, you need to check which one it is via:

```java
MinecraftServer.getGlobalEventHandler().addListener(BlockChangeEvent.class, event -> {
      if(event.getBlock().isAir() && event.getSource() instanceof BlockEvent.Source.Player) {
        // This would represent a event where the block is being broken by the player.
      }
    }
});
```

As for the main purpose of this change, BlockHandlers, the change is a little less simple. Since BlockPlacementRules have been removed, BlockHandlers has been modified to accept all of the methods that BlockPlacementRules did, along with the ones which were previously inaccessible to BlockPlacementRules.

Additionally, some unnecessary classes were removed, that being:

* PlacementState
* UpdateState
* Replacement
* Placement
* PlayerPlacement
* Destroy
* PlayerDestroy
* Interaction
* Touch
* Tick

The reason for the removal is pretty obvious, these all share a lot of the same data and have very similar names. Additionally, the benefit of constructing an entire object to pass around is dubious, especially when no modification to the actual data is being done. Instead of this, all methods now take in the parameters they need directly, or have switched to taking in an event directly. For these, you are able to fix any broken code by simply just using the fields you used to construct the object. 

In the end, the changes here are breaking, and can take a bit of time depending on the speed of the developer, however the actual code inside the new BlockHandlers will be nearly a drop in replacement, requiring only some minor name changes when copy pasting in from your old BlockHandlers & BlockPlacementRules.

## Types of changes

## Checklist

merging your code._

## Further comments

Minor Changes:

* BlockManager now takes in GlobalEventHandler in its constructor to ensure safe construction of a EventNode that relays changes to BlockHandlers. This node has a default priority of 50, which allows the developer to control when changes they may make to a BlockChangeEvent take place, with the default being before the BlockHandlers changes are made.

* Instance#setBlock, Instance#placeBlock, Instance#breakBlock now relay a unmodifiable object called BlockChangeEvent.Result, which contains data about if the method failed, and data about the finalized version of the block that was changed. Additionally, these methods no longer require Placement/Destroy objects, as they were removed. 

 * AnvilLoader now saves the handler.getBlock().id() for block entities. This change is made to align with the new system as handlers no longer have their own namespaces, and are instead attached to the blocks id. This may be changed as I am unsure how this would affect upgrading anvil worlds.
 
 * BlockHandler#onInteract now uses the full Interaction Event.
 
 * Other block events such as PlayerStartDiggingEvent, PlayerFinishDiggingEvent, PlayerCancelDiggingEvent have had BlockEvent.Source added to them, although their type is restricted to BlockEvent.Source.Player.
 
 Still WIP:
 
 * ~~Need to remove redundant fields from BlockEvents where the data is contained in BlockEvent.Source~~
 * ~~Need to fix building/github workflows~~
 * Need to cleanup and format code
 * Need to test how AnvilLoader changes work and possibly revert to namespaceId's
 * Further Notes and guides on migration of breaking changes.
 * Better description of why these changes are needed.


DersWasTaken:
I am unsure why these checks fail here, if anyone has any idea let me know.

DersWasTaken:
Closed because of #2692 on 1.21.5 branch

PR #2683, opened 2025-03-12
Add state name method and log it when a packet isn't found.

GreatWyrm:
## Proposed changes

Display the state name when the server cannot find a registered packet.
This should hopefully cut down on instances where developers are confused when they try do to something in an invalid client/server state (e.g. set gamemode when in config state).

## Types of changes

## Checklist

merging your code._

## Further comments
I would have liked to do it through `getClass().getSimpleName()` instead of having to manually type it out, but those were empty strings due to java shenanigans.

### DIFF
--- a/src/main/java/net/minestom/server/network/packet/PacketRegistry.java
+++ b/src/main/java/net/minestom/server/network/packet/PacketRegistry.java
@@ -1,5 +1,6 @@
 package net.minestom.server.network.packet;
 
+import net.minestom.server.network.ConnectionState;
 import net.minestom.server.network.NetworkBuffer;
 import net.minestom.server.network.packet.client.ClientPacket;
 import net.minestom.server.network.packet.client.common.*;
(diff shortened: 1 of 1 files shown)


KermanIsPretty:
Probably use the client state enum.

src/main/java/net/minestom/server/network/packet/PacketRegistry.java
```diff
@@ -364,7 +430,7 @@ protected PacketInfo<T> computeValue(@NotNull Class<?> type) {
                         return (PacketInfo<T>) info;
                     }
                 }
-                throw new IllegalStateException("Packet type " + type + " isn't registered!");
+                throw new IllegalStateException("Packet type " + type + " isn't registered for state " + side().name() + "_" + state().name() + "!");
```
LooFifteen, line 433: This should probably be rephrased. `packet type {type} cannot be sent in state {side} {state}`

LooFifteen (approved):

mworzala (approved):

PR #2688, opened 2025-03-15
remove unnecessary distance check

reosfire:
Removed distance check which is always false because EntityTracker::nearbyEntities method already checks that for us and will not invoke our lambda with such entities


GreatWyrm (approved):
Can confirm distance is checked in nearbyEntities method, the extra check is not needed.

mworzala (approved):

PR #2692, opened 2025-03-19
rework/block-handling

DersWasTaken:
## Proposed changes

Firstly, PlayerBlockEvent's are completely removed in this PR, instead being merged into BlockChangeEvent which represents all changes to blocks, both done by the player and done by the server via Instance#setBlock. This aims to reduce repetitive calls to events that happened in #2666 by having a singular event represent all changes, where BlockEvent.Source represents the "source" of the changes (i.e Instance or Player) where each type of source contains additional data to the event.

Since this is a breaking change, I tried to make the transition as smooth as possible:

```java
MinecraftServer.getGlobalEventHandler().addListener(BlockChangeEvent.class, event -> {
    BlockEvent.Source source = event.getSource(); // new Source type
    if(source instance BlockEvent.Source.Instance) {
       // The source comes from the instance via Instance#setBlock.
    }
    if(source instanceof BlockEvent.Source.Player) {
      // The source comes from a player action (i.e Breaking Block, Placing Block) comes with the following data:
      source.getPlayer() //Never Null
      source.getBlockFace() //May be null
      source.getCursorPosition() //Never Null
      source.getHand() //May be null
    }
});
```

And since BlockChangeEvent represents both breaking and placing, you need to check which one it is via:

```java
MinecraftServer.getGlobalEventHandler().addListener(BlockChangeEvent.class, event -> {
      if(event.fromPlayer() && event.isBlockBreak()) {
        // This would represent an event where the block is being broken by the player.
      }
    }
});
```

Additionally, some unnecessary classes were removed, that being:

* PlacementState
* UpdateState
* Replacement
* Placement
* PlayerPlacement
* Destroy
* PlayerDestroy
* Interaction
* Touch
* Tick

The reason for the removal is pretty obvious, these all share a lot of the same data and have very similar names. Additionally, the benefit of constructing an entire object to pass around is dubious, especially when no modification to the actual data is being done. Instead of this, all methods now take in the parameters they need directly, or have switched to taking in an event directly. For these, you are able to fix any broken code by simply just using the fields you used.

## Types of changes

## Checklist

merging your code._

## Further comments

I still need to add some comments and documentation, but would like a review before I do so. Additionally I want to ask about the possibility of making it so that BlockHandler is only used for storing data that makes block types unable to be flattened, and instead BlockPlacementRules handles block changes (i.e breaking, placing) and interaction/entity touch. 

My reasoning for this is that a lot of blocks need these, especially interactions (i.e right clicking) but do not need to actually store data. This makes it so that blocks that DO need to store data, and thus cannot be flattened, rely on handler to store that data, which can be easily modified/recieved from BlockPlacementRules by just calling Block#handler and setting the handler when the block is placed. 

However, when blocks that DONT need to store data, but still need access to interactions, they can do so without the unnecessary overhead of BlockHandler.


src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
```diff
@@ -22,23 +22,34 @@ protected BlockPlacementRule(@NotNull Block block) {
...
 
     /**
      * Called when the block is placed.
      * It is recommended that you only set up basic properties on the block for this placement, such as determining facing, etc
      *
-     * @param placementState The current parameters to the block placement
      * @return the block to place, {@code null} to cancel
      */
-    public abstract @Nullable Block blockPlace(@NotNull PlacementState placementState);
+    public abstract @Nullable Block blockPlace(@NotNull Block.Getter instance,
```
DersWasTaken, line 40: Should consider removing some of the data here as it is unnecessary. I'm not sure I see a reason to pass in the Players Position, ItemStack, Cursor Position, and if they are shifting and not just pass in the player itself and have the user handle it.

PR #2696, opened 2025-03-21
Removal of BlockPlacementRule

DersWasTaken:
## Proposed changes

BlockPlacementRules realistically should be events, and the current implementation is not very configurable for the developer. Currently, there is no way to modify the shape of block updates, which is needed as specific blocks have directional updates, or do not receive updates from specific directions. Examples of this are:

* Fences: they do not receive block updates from above or below.
* Repeaters/Comparators: See image below. 
Screenshot 2025-03-21 at 4 50 46 PM

*Note: these examples are not exhaustive, are aim to demonstrate the point*

Additionally, BlockPlacementRules are inconsistently applied in code, where sometimes they are applied before an event is called, and sometimes after. This is bad because:

* The user may make changes to a block that then has its events cancelled, meaning the potentially expensive computation was unnecessary.
* The user may want to make a change to a block before an event is fired, which could result in a change in how the event operates. Instead, the event is fired first and cancels it, meaning that the block change that should have been done, isn't. 

Removal of block placement rules solves all of the above problems.

Additionally, leaving the developer to implement these systems is more ideal, as they can adapt it to their specific use-case and removes unnecessary overhead if the user does not want to implement any block placement rules like behavior.

## Types of changes

## Checklist

merging your code._

## Further comments


src/main/java/net/minestom/server/listener/BlockPlacementListener.java
```diff
@@ -91,32 +92,45 @@ public static void listener(ClientPlayerBlockPlacementPacket packet, Player play
...
             return;
         }
 
         // Ensure that the final placement position is inside the world border.
-        if (!instance.getWorldBorder().inBounds(blockPosition)) {
+        if (!instance.getWorldBorder().inBounds(placementPosition)) {
             canPlaceBlock = false;
         }
 
         if (!canPlaceBlock) {
```
DersWasTaken, line 122: Potential issue here where user may want to determine themselves if the block is replaceable. A possible solution may be to remove this check entirely and pass it onto the event, which then can be changed if the user desires.

DersWasTaken: Fixed in newer commits by addition of ReplacementRule

DersWasTaken:
Now i am aware adding BlockReplacementRule's seems stupid considering the PR is to get rid of BlockPlacementRules, but I dont really think there is another good way to do it since creating an event for this seems pretty dumb, and passing it through to the event makes it unnecessarily complicated.

Removal of BlockPlacementRules still is optimal for everything else, but replacement behavior is kind of baked in and since placement rejection for not being able to place blocks should be before an event is called, I think this is a reasonable compromise. 

The difference between this and BlockPlacementRules is that this is the only practical way to do this, while everything else can and should be done via events.

PR #2707, opened 2025-04-04
Fix disabling damage animation

TogAr2:
## Proposed changes

Fixed the damage animation still being sent, even if `EntityDamageEvent#shouldAnimate()` was set to false.
I also removed sending the `EntityAnimationPacket` since it has been removed in vanilla and does nothing.

## Types of changes

## Checklist

merging your code._


TogAr2:
Just noticed #2063, however I am fairly sure the `DamageEventPacket` does not dictate the animation direction (`HitAnimationPacket` does). So I think adding an animation type does not make sense here.

mworzala (approved):

mworzala:
#2063 might be from before HitAnimationPacket existed, though it does seem like we should send a HitAnimationPacket in this case. I am going to merge this since it's unrelated and close the other one since it's out of date.

Will need to look more into where we need to send a HitAnimationPacket (currently its never sent)

PR #2715, opened 2025-04-14
feat: ThreadDispatcher handlers

leon0017:
## Proposed changes

Based of my older PR and would close https://github.com/Minestom/Minestom/pull/2534

These changes allow users of Minestom to slot in their own handlers for different events that would require action from the ThreadDispatcher. It also allows customization of the ThreadProvider.

Before, the ThreadDispatcher always had a partition type of Chunk, but now it can be anything that extends Tickable.

There are 2 ThreadDispatcherHandler classes that I provided in by default, they are:
 - `ChunkBasedThreadDispatcherHandler` the default, and which has the exact same behaviour of threading as before, where chunks can have different ticking threads
 - `InstanceBasedThreadDispatcherHandler` where the partition is `Instance`, meaning an instance can only have 1 ticking thread managing it, and every chunk will tick under that 1 ticking thread

Users can slot in their own ThreadDispatcherHandler by using `MinecraftServer#init(ThreadDispatcherHandler<P> dispatcherHandler)` The `MinecraftServer#init()` function without any parameters has not been removed of course.

I am sure there are most likely improvements to be made and I am not sure how to go about writing tests for this feature.

## Types of changes

## Checklist

merging your code._

## Further comments

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -80,18 +80,18 @@ public final class MinecraftServer implements MinecraftConstants {
         return init(createDefaultDispatcherHandler());
     }
 
-    @ApiStatus.Internal
-    public static ServerProcess updateProcess() {
-        return updateProcess(createDefaultDispatcherHandler());
-    }
-
-    public static <P extends Tickable> MinecraftServer init(ThreadDispatcherHandler<P> dispatcherHandler) {
+    public static <P extends Tickable> MinecraftServer init(@NotNull ThreadDispatcherHandler<P> dispatcherHandler) {
         updateProcess(dispatcherHandler);
         return new MinecraftServer();
     }
 
     @ApiStatus.Internal
-    public static <P extends Tickable> ServerProcess updateProcess(ThreadDispatcherHandler<P> dispatcherHandler) {
+    public static ServerProcess updateProcess() {
+        return updateProcess(createDefaultDispatcherHandler());
+    }
+
+    @ApiStatus.Internal
+    public static <P extends Tickable> ServerProcess updateProcess(@NotNull ThreadDispatcherHandler<P> dispatcherHandler) {
         ServerProcess process = new ServerProcessImpl<>(dispatcherHandler);
         serverProcess = process;
         return process;


src/main/java/net/minestom/server/instance/InstanceManager.java
```diff
@@ -122,11 +122,10 @@ public void unregisterInstance(@NotNull Instance instance) {
             EventDispatcher.call(event);
 
             // Unload all chunks
-            if (instance instanceof InstanceContainer) {
+            if (instance instanceof InstanceContainer)
                 instance.getChunks().forEach(instance::unloadChunk);
```
leon0017, line 126: I removed the deletePartition from here as `Instance#unloadChunk` already deletes the partitions

src/main/java/net/minestom/server/thread/ThreadDispatcherHandler.java
kermandev, line 1: Any reason this can't be an interface?

src/main/java/net/minestom/server/thread/ChunkBasedThreadDispatcherHandler.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void onChunkCached(@NotNull Chunk chunk, @NotNull Instance instance) {
+        dispatcher.createPartition(chunk);
+    }
+
+    @Override
+    public void onInstanceUnregister(@NotNull Instance instance) {
+        // Nothing required to be done as all chunks will be unloaded when instance is unregistered, deleting all the partitions from onChunkUnload
```
kermandev, line 46: Could you add an assertion for this?

src/main/java/net/minestom/server/thread/ThreadDispatcherHandler.java
leon0017: I thought having a `ThreadDispatcher<P> dispatcher` variable would be better than passing it into each function every time but i'm unsure

src/main/java/net/minestom/server/thread/ChunkBasedThreadDispatcherHandler.java
leon0017: I'm unsure what the assertion would look like

kermandev: Ensure there are no partitions for any chunks after this is called

leon0017: Because the thread dispatcher call of deleting the chunk partitions would put it on a message queue, I think the assertion could run before all the partitions are removed and then wrongly fail

src/main/java/net/minestom/server/thread/ThreadDispatcherHandler.java
kermandev: I think it would be more flexible if it was a function that returned a dispatcher instead. That way the field could be final and the default interface could have a dispatcher field

src/main/java/net/minestom/server/thread/ChunkBasedThreadDispatcherHandler.java
kermandev: I guess that is right, I think the of chunks should get de partitioned here though

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -75,17 +77,30 @@ public final class MinecraftServer implements MinecraftConstants {
...
     public static MinecraftServer init() {
-        updateProcess();
-        return new MinecraftServer();
+        return init(createDefaultDispatcherHandler());
     }
 
     @ApiStatus.Internal
     public static ServerProcess updateProcess() {
-        ServerProcess process = new ServerProcessImpl();
+        return updateProcess(createDefaultDispatcherHandler());
```
mworzala, line 85: weird method ordering. should group the updateProcess methods together

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -75,17 +77,30 @@ public final class MinecraftServer implements MinecraftConstants {
...
+    }
+
+    public static <P extends Tickable> MinecraftServer init(ThreadDispatcherHandler<P> dispatcherHandler) {
```
mworzala, line 88: add `@NotNull` annotation

src/main/java/net/minestom/server/ServerProcess.java
```diff
@@ -114,7 +114,12 @@ public interface ServerProcess extends Registries, Snapshotable {
     /**
      * Dispatcher for tickable game objects.
      */
-    @NotNull ThreadDispatcher<Chunk> dispatcher();
+    @NotNull ThreadDispatcher<? extends Tickable> dispatcher();
+
+    /**
+     * Handler for dispatcher events.
+     */
+    @NotNull ThreadDispatcherHandler<? extends Tickable> dispatcherHandler();
```
mworzala, line 122: The generic here seems to be largely unused, and I am not a big fan of it invading the `ServerProcess` definition.

I do wonder if it would make more sense to generalize the definition of `ThreadDispatcherHandler` to just `addEntry(Tickable)` and `removeEntry(Tickable)`. and leave it up to the implementation to figure out what to do.

It would be simple enough to check the types you care about and handle them and would remove the entity/chunk/instance special cases. There is some ambiguity about what to do if someone adds some random Tickable, but I believe it would be OK to throw an error and expect them to implement their own `ThreadDispatcherHandler` to support their case.

As for entities moving chunks, either could allow re-adding the entry or an `updateEntry(Tickable)` method.

Thoughts?

src/main/java/net/minestom/server/thread/ThreadDispatcher.java
```diff
@@ -236,7 +239,9 @@ private void processLoadedPartition(P partition) {
         this.partitions.put(partition, partitionEntry);
         this.partitionUpdateQueue.add(partition);
         if (partition instanceof Tickable tickable) {
-            processUpdatedElement(tickable, partition);
+            if (!(tickable instanceof Instance)) { // Instances are ticked separately by server process
```
mworzala, line 242: This is a bit tricky - I generally think that the behavior of ticking instances separately at the start of the tick doesn't make sense for a per-instance scheme. Though at the same time I would expect instances to be ticked before chunks and entities so some kind of ordering would be required here. Not sure the best solution.

src/main/java/net/minestom/server/ServerProcess.java
leon0017: I have an alternative instead of invading `ServerProcess`

And instead of passing in the `ThreadDispatcherHandler` to the `MinecraftServer#init` function, you could use this alternative:

```java
MinecraftServer.setThreadDispatcherHandler(new InstanceBasedThreadDispatcherHandler());
```

And then there would be a `MinecraftServer#getThreadDispatcherHandler` function.

I think this is a better approach instead of invading the init function.

The only downside I can see is that because of generics, you would need to cast the `ThreadDispatcher<P>` like this when setting it in the `ThreadDispatcherHandler` for use:

```java
public abstract class ThreadDispatcherHandler<P extends Tickable> {

    protected ThreadDispatcher<P> dispatcher;

    @SuppressWarnings("unchecked")
    @ApiStatus.Internal
    public void setupDispatcher(@NotNull ThreadDispatcher<? extends Tickable> threadDispatcher) {
        this.dispatcher = (ThreadDispatcher<P>) threadDispatcher;
    }
    // ...
```
(this is called in `ServerProcessImpl` like this)
```java
this.dispatcher = ThreadDispatcher.of(MinecraftServer.getThreadDispatcherHandler().provider(), ServerFlag.DISPATCHER_THREADS);
MinecraftServer.getThreadDispatcherHandler().setupDispatcher(this.dispatcher);
```

It should never fail as long as it is only called here.

Do you think this is a better option?

leon0017: And as for the `addEntry` and `removeEntry` function ideas, I think that might be a better solution and I will play around with that

PR #2718, opened 2025-04-20
feat: add basic bed and sleep functionality

FrederikHeinrich:
## Proposed changes

This PR introduces a basic implementation of bed interaction in Minestom. It allows players to interact with bed blocks and simulates entering the sleeping state. This lays the groundwork for future support of Minecraft’s sleep mechanics and potential multiplayer sleep synchronization.

## Types of changes

## Checklist

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -607,6 +607,13 @@ public class LivingEntity extends Entity implements EquipmentHandler {
         }
     }
 
+    public void leaveBed() {
+        LivingEntityMeta meta = getLivingEntityMeta();
+        if (meta != null) {
+            meta.setBedInWhichSleepingPosition(null);
+        }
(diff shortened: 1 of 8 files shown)


src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -607,6 +607,13 @@ public void refreshActiveHand(boolean isHandActive, boolean offHand, boolean rip
         }
     }
 
+    public void leaveBed() {
```
mworzala, line 610: perhaps should add `enterBed` or `sleep` taking a position to mimic this method.

mworzala (commented):
Small comment but otherwise looks good, thanks for this

src/main/java/net/minestom/server/entity/LivingEntity.java
FrederikHeinrich: done

PR #2727, opened 2025-05-01
Feat: Faster Socket Writing wakeup

kermandev:
## Proposed changes

Adds a faster non-tick-based socket write locking, which is disabled by default behind a server flag until it can be proven stable; no issues with assertions enabled on demo were found.

This should remove the 25ms delay observed by the player's ping, but might also increase the possibility of work not being shared adequately across players evenly, as the sleep unmounts the virtual thread.

## Types of changes

## Checklist


mworzala (approved):

DasBabyPixel:
Hmm we ignoring the spurious wakeup with an assertion for new content after it? 🤔

mworzala:
what is the spurious wakeup

PR #2699, opened 2025-03-25
1.21.5 Support

mworzala:
Still has a couple remaining TODOs, but should be functional for the vast majority of people.

1.21.4 will be merged first.

PR #2732, opened 2025-05-06
Implement new data component predicates in block predicates

FluxCapacitor2:
## Proposed changes

Implements data component predicates, which are now included in block predicates. Resolves the issue described in [this comment](https://github.com/Minestom/Minestom/pull/2699#issuecomment-2848516382) and this TODO comment:
https://github.com/Minestom/Minestom/blob/2398778b46cd1718780a62b74c85c9ad9a041363/src/main/java/net/minestom/server/instance/block/predicate/BlockPredicate.java#L56

## Types of changes

## Checklist

- [x] Make sure the client can read all the new predicates correctly (test manually)
- [x] Make sure that the `container` predicate works with an item predicate
- [x] Add support for block tags
- [x] Clean up unused network types (IntBounds/DoubleBounds are only ever serialized to NBT, so the network type is unnecessary)
- [ ] Consider renaming some of the classes to be less confusing. The predicates are named the same as some of their data component counterparts, which might cause incorrect IDE suggestions. Should I add a `*Predicate` suffix to all of them?
- [x] Add unit tests to make sure my custom network types can encode a BlockPredicate with data component predicates and then decode it again to the same object
- [x] Add some Javadoc comments

## Remaining questions

1. In `DataComponentPredicates`, I keep a list of the different predicate types in the order that the client registers them in a registry. Is there a way to autogenerate that, and if so, is it worth the effort?
2. Are there any other places that these predicates can/should be used, like advancements or entity predicates in commands?

I'm leaving this PR in draft mode until more of the checklist above is completed. Feel free to leave feedback in the meantime!

## Breaking Changes

- The `minecraft:lock` data component is now an `ItemPredicate` instead of `CustomData`
- The 3-argument `BlockPredicate` constructor which previously accepted a `CompoundBinaryTag` now accepts an `NbtPredicate`. The 1-argument constructor has a wrapper that creates an NbtPredicate from the given CompoundBinaryTag, but the 3-argument constructor doesn't.


src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,97 @@
...
+        return StructCodec.struct(
+                "contains", Contains.createCodec(codec).optional(), CollectionPredicate::contains,
+                "count", Count.createCodec(codec).optional(), CollectionPredicate::counts,
+                "size", DataComponentPredicates.INT_RANGE_CODEC.optional(), CollectionPredicate::size,
+                CollectionPredicate::new
+        );
+    }
+
+    @Override
+    public boolean test(Iterable<T> collection) {
```
GoldenStack, line 24: From what I can tell, the data components that use this structure (`attribute_modifiers`, `bundle_contents`, `container`, `fireworks`, `writable_book_content`, `written_book_content`) all store the items as lists, so this should be able to be converted to a `List<T>` or `Collection<T>` which means you should be able to work over one of those types and remove the ugly `sizeOf` method and some other associated stuff. LMK if I missed something that makes this impossible, though.

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,97 @@
...
+
+import net.minestom.server.codec.Codec;
+import net.minestom.server.codec.StructCodec;
+import net.minestom.server.utils.Range;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+
+public record CollectionPredicate<T, P extends Predicate<T>>(Contains<T, P> contains, Count<T, P> counts,
```
GoldenStack, line 11: Should probably at least put `@Nullable` on the nullable fields (you omit the `@Nullable` in other field definitions too but I'm just mentioning it here), and preferably also `@NotNull` on the non-nullable ones.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,376 @@
...
+            }
+            for (DynamicRegistry.Key<net.minestom.server.item.enchant.Enchantment> e : enchantments) {
+                if (holderEnchants.has(e) && (levels == null || levels.inRange(holderEnchants.level(e)))) {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+    }
```
GoldenStack, line 125: You should be able to deduplicate this by having `Enchantments`/`StoredEnchantments` pass in the relevant `EnchantmentList`. There don't need to be two implementations.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,376 @@
...
+
+    record Potions(List<PotionType> potionTypes) implements DataComponentPredicate {
+        public static Codec<Potions> CODEC = PotionType.CODEC.listOrSingle().transform(Potions::new, Potions::potionTypes);
```
GoldenStack, line 128: This doesn't handle the case of there being a tag of potion types. Essentially, per [the wiki](https://minecraft.wiki/w/Data_component_predicate#potion_contents) (and I did confirm otherwise that this is how it actually works), you need to be able to support a tag of something, e.g. for items, if the string `#minecraft:logs` is passed in it needs to refer to all logs. You also need to handle this in many other cases across the `DataComponent` predicates but I'm just mentioning this here so I don't have to spam it everywhere.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,376 @@
...
+
+        @Override
+        public boolean test(DataComponent.Holder holder) {
+            var potion = holder.get(DataComponents.POTION_CONTENTS);
+            if (potion == null) return false;
+            return potionTypes.contains(potion.potion());
+        }
+    }
+
+    record CustomData(BinaryTag nbt) implements DataComponentPredicate {
```
GoldenStack, line 138: Just like I mentioned for `BlockPredicate#nbt` this does also need to be a NBT predicate.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,376 @@
...
+    record Container(CollectionPredicate<ItemStack, ItemPredicate> items) implements DataComponentPredicate {
+        public static Codec<Container> CODEC = StructCodec.struct(
+                "items", CollectionPredicate.createCodec(ItemPredicate.CODEC), Container::items,
+                Container::new
+        );
+
+        @Override
+        public boolean test(DataComponent.Holder holder) {
+            Iterable<ItemStack> itemStacks = holder.get(DataComponents.CONTAINER);
+            return items().test(itemStacks);
```
GoldenStack, line 157: This needs to exclude air (or otherwise air needs to always pass/not affect the results).

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
GoldenStack: Okay I can see it used in `AttributeModifiers#test` but I think this isn't at all a hard dependency.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,376 @@
...
+        public boolean test(DataComponent.Holder holder) {
+            if (modifiers == null) return true;
+            AttributeList attributes = holder.get(DataComponents.ATTRIBUTE_MODIFIERS);
+            if (attributes == null) return false;
+            return modifiers.test(attributes.modifiers().stream().map(modifier -> new Entry(
+                    modifier.attribute(), modifier.modifier().id(), modifier.modifier().amount(), modifier.modifier().operation(), modifier.slot()
+            ))::iterator);
+        }
+
+        record Entry(Attribute attribute, Key id, Double amount, AttributeOperation operation,
```
GoldenStack, line 312: Is there a reason why there's this intermediate `Entry` class?

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,376 @@
...
+                "material", TrimMaterial.CODEC.transform(holder -> holder.resolve(MinecraftServer.getTrimMaterialRegistry()), Holder.Direct::new).listOrSingle().optional(), ArmorTrim::material,
+                "pattern", TrimPattern.CODEC.transform(holder -> holder.resolve(MinecraftServer.getTrimPatternRegistry()), Holder.Direct::new).listOrSingle().optional(), ArmorTrim::pattern,
+                ArmorTrim::new
+        );
+
+        @Override
+        public boolean test(DataComponent.Holder holder) {
+            var trim = holder.get(DataComponents.TRIM);
+            if (trim == null) return false;
+            if (material != null && material.contains(trim.material().resolve(MinecraftServer.getTrimMaterialRegistry())))
```
GoldenStack, line 356: This should be `!material.contains`

GoldenStack (commented):
There are a couple implementation issues but it's definitely mostly correct. LMK if I can help implement anything (since I was going to anyway before you told me about this PR).

src/main/java/net/minestom/server/item/predicate/ItemPredicate.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public boolean test(ItemStack itemStack) {
+        if (items != null && !items.contains(itemStack.material()))
+            return false;
+        if (count != null && !count.inRange(itemStack.amount()))
+            return false;
+        if (predicates == null)
+            return true;
+
+        // Construct a DataComponentMap with ALL the item's components, including the default ones originating from the item's material
```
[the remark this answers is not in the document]
FluxCapacitor2: Good idea, I forgot that ItemStack implemented DataComponent.Holder!

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
FluxCapacitor2: It looks like this isn't possible for the `jukebox_playable` and `trim` predicates until the tag rework PR #2709 is merged, but I added tag support for `enchantments`, `stored_enchantments`, and `potion_contents`.

FluxCapacitor2:
@GoldenStack Thanks for the very thorough review! I have addressed everything you've mentioned and I'll keep working on javadocs, unit tests, and manually verifying that the client reads the all the predicates properly.

I wasn't able to easily add tag support for the jukebox_playable and trim predicates ([thread](https://github.com/Minestom/Minestom/pull/2732#discussion_r2096619807)) because they don't exist in vanilla, so Minestom doesn't load them. It looks like that'll be possible when #2709 lands.

FluxCapacitor2:
Working on some unit tests now and then I'll be done

src/main/java/net/minestom/server/instance/block/predicate/BlockPredicate.java
```diff
@@ -29,58 +34,85 @@
...
+ * @param nbt                 The block nbt to match.
+ * @param componentPredicates Predicates to match a block entity's stored data components
  */
 public record BlockPredicate(
         @Nullable RegistryTag<Block> blocks,
         @Nullable PropertiesPredicate state,
-        @Nullable CompoundBinaryTag nbt,
-        @NotNull DataComponentPredicates components
+        @Nullable NbtPredicate nbt,
+        @NotNull DataComponentPredicates componentPredicates
```
mworzala: should be called `components` inline with the other names here

src/main/java/net/minestom/server/instance/block/predicate/BlockPredicate.java
```diff
@@ -29,58 +34,85 @@
...
+        this(null, null, null, new DataComponentPredicates(null, predicates));
+    }
+
+    public BlockPredicate(@NotNull DataComponentPredicates predicates) {
+        this(null, null, null, predicates);
     }
 
-    public BlockPredicate(@Nullable RegistryTag<Block> blocks, @Nullable PropertiesPredicate state, @Nullable CompoundBinaryTag nbt) {
-        this(blocks, state, nbt, DataComponentPredicates.EMPTY);
+    public BlockPredicate(RegistryTag<Block> blocks, PropertiesPredicate state, NbtPredicate nbt) {
```
mworzala: Missing nullability annotations

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,156 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * A predicate that counts the number of matching sub-predicates
+     * and tests whether it's in the <code>count</code> range.
+     */
+    public record Count<T, P extends Predicate<T>>(
+            @NotNull List<Entry<T, P>> entries) implements Predicate<Collection<T>> {
```
mworzala: Weird formatting, newline for `) implements ...` or no newline before entries field

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * A predicate that requires that all of its sub-predicates match at least once.
+     */
+    public record Contains<T, P extends Predicate<T>>(@NotNull List<P> predicates) implements Predicate<Collection<T>> {
+
+        public Contains {
+            predicates = List.copyOf(predicates);
+        }
+
+        public static <T, P extends Predicate<T>> @NotNull Codec<Contains<T, P>> createCodec(Codec<P> codec) {
```
mworzala: canonically just called `codec`, also missing nullability (same comment for other occurrences)

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,156 @@
...
+            return codec.listOrSingle().transform(Contains::new, Contains::predicates);
+        }
+
+        @Override
+        public boolean test(@NotNull Collection<T> collection) {
+            List<Predicate<T>> predicates = new ArrayList<>(this.predicates);
```
mworzala: The copy here seems unnecessary, can just count the matches

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,156 @@
...
+
+        /**
+         * Specifies that the collection's size must be inside the <code>size</code> range.
+         */
+        public Builder<T, P> matchSize(Range.Int size) {
+            this.size = size;
+            return this;
+        }
+
+        public CollectionPredicate<T, P> build() {
```
mworzala, line 162: Nullability annotations everywhere (im going to stop commenting this but please go through and check cases)

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -1,35 +1,208 @@
...
+
+    public static final Codec<Range.Double> DOUBLE_RANGE_CODEC = StructCodec.struct(
+            "min", Codec.DOUBLE, Range.Double::min,
+            "max", Codec.DOUBLE, Range.Double::max,
+            Range.Double::new
+    ).orElse(Codec.DOUBLE.transform(Range.Double::new, Range.Double::min));
+
+    public enum ComponentPredicateType {
+
+        DAMAGE("damage", DataComponentPredicate.Damage.CODEC),
```
mworzala: This should be done with an internal registry (see enchantment components) or just tagged union codec (see recipe types eg SlotDisplay), happy to help more with this ping me on discord if needed

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -1,35 +1,208 @@
...
+            if (id < 0 || id >= values().length) {
+                throw new IllegalArgumentException("Invalid ComponentPredicateType ID: " + id);
+            }
+            return values()[id];
+        }
     }
 
-    private List<Unit> partial() {
-        return List.of();
+    private static final Codec<Map<ComponentPredicateType, DataComponentPredicate>> predicateCodec = Codec.NBT_COMPOUND.transform(
```
mworzala: I might be missing something but my initial guess is that this should not be based on NBT_COMPOUND but should itself be a structcodec. please lmk what I am missing if there is something.

src/main/java/net/minestom/server/instance/block/predicate/NbtPredicate.java
```diff
@@ -0,0 +1,83 @@
...
+import net.kyori.adventure.nbt.ListBinaryTag;
+import net.kyori.adventure.nbt.TagStringIOExt;
+import net.minestom.server.codec.Codec;
+import net.minestom.server.network.NetworkBuffer;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.function.Predicate;
+
+public record NbtPredicate(@Nullable CompoundBinaryTag nbt) implements Predicate<@Nullable BinaryTag> {
+    private static final Codec<CompoundBinaryTag> NBT_STRING_CODEC = Codec.STRING.transform(
```
mworzala: Should add to Codec as `NBT_COMPOUND_COERCED` (similar to UUID_COERCED)

src/main/java/net/minestom/server/item/predicate/ItemPredicate.java
```diff
@@ -0,0 +1,72 @@
...
+) implements Predicate<ItemStack> {
+
+    public static final Codec<ItemPredicate> CODEC = StructCodec.struct(
+            "items", Material.CODEC.listOrSingle().optional(), ItemPredicate::items,
+            "count", DataComponentPredicates.INT_RANGE_CODEC.optional(), ItemPredicate::count,
+            StructCodec.INLINE, DataComponentPredicates.CODEC.optional(), ItemPredicate::predicates,
+            ItemPredicate::new
+    );
+
+    public static final NetworkBuffer.Type<ItemPredicate> NETWORK_TYPE = NBT_COMPOUND.transform(
```
mworzala: This is `NetworkBuffer.TypedNBT(CODEC)`

src/main/java/net/minestom/server/instance/block/predicate/NbtPredicate.java
```diff
@@ -0,0 +1,83 @@
...
+     * have extra fields that are not contained in the standard.
+     *
+     * @param standard   the standard that the comparison must have all elements of
+     * @param comparison the comparison, that is being compared against the standard. NBT compounds in this parameter,
+     *                   whether deeper in the tree or not, are allowed to have keys that the standard does not - it's
+     *                   basically compared against a standard.
+     * @return true if the comparison fits the standard, otherwise false
+     */
+    // Modified from GoldenStack's trove library under MIT license: https://github.com/GoldenStack/trove/blob/d8f329644c2025cd71802450c7c36a4a6a0c0410/src/main/java/net/goldenstack/loot/util/nbt/NBTUtils.java#L13-L64
+    public static boolean compareNBT(@Nullable BinaryTag standard, @Nullable BinaryTag comparison) {
```
mworzala, line 38: This needs some tests

mworzala (commented):
I added some comments but this is overall looking extremely good, thanks for taking this work!

src/main/java/net/minestom/server/MinecraftServer.java
```diff
@@ -309,6 +311,10 @@ public static AdvancementManager getAdvancementManager() {
         return process().enchantmentLocationEffects();
     }
 
+    public static @NotNull DynamicRegistry<Codec<? extends DataComponentPredicate>> componentPredicateTypes() {
+        return process().componentPredicateTypes();
+    }
+
```
FluxCapacitor2: Should I expose the registry here and in `Registries` if it's only ever going to be used internally? Minestom users will probably never have to access it.

src/main/java/net/minestom/server/codec/Codec.java
```diff
@@ -150,11 +156,19 @@ sealed interface RawValue permits CodecImpl.RawValueImpl {
...
     }
 
     default <V> @NotNull Codec<Map<T, V>> mapValue(@NotNull Codec<V> valueCodec) {
-        return mapValue(valueCodec, Integer.MAX_VALUE);
+        return mapValue((ignored) -> valueCodec, Integer.MAX_VALUE);
+    }
+
+    default <V> @NotNull Codec<Map<T, V>> mapValue(@NotNull Function<T, @NotNull Codec<V>> valueCodecGetter, int maxSize) {
+        return new CodecImpl.MapImpl<>(Codec.this, valueCodecGetter, maxSize);
+    }
```
FluxCapacitor2: Are you ok with this and the corresponding CodecImpl change? It makes my codecs in `ComponentPredicateSet` much easier to write.

src/main/java/net/minestom/server/instance/block/predicate/BlockPredicate.java
```diff
@@ -29,58 +34,85 @@
...
+    public BlockPredicate(@NotNull DataComponentMap components) {
+        this(null, null, null, new DataComponentPredicates(components, null));
+    }
+
+    public BlockPredicate(@NotNull ComponentPredicateSet predicates) {
+        this(null, null, null, new DataComponentPredicates(null, predicates));
+    }
+
+    public BlockPredicate(@NotNull DataComponentPredicates predicates) {
+        this(null, null, null, predicates);
```
FluxCapacitor2: There are so many constructors for this class now. Should we add a builder, and if so, should we remove the partial constructors?

GoldenStack:
From testing with loot tables, I've discovered that your EnchantmentListPredicate codec is wrong; it needs to have a default value for the `max` [here](https://github.com/FluxCapacitor2/Minestom/blob/feat/component-predicates/src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java#L104), since right now 144 (of the 1244 total) vanilla loot tables fail to deserialize because of this. I guess a suitable maximum is probably Byte.MAX_VALUE or Short.MAX_VALUE, but I'm not sure which.

Also when you add the maximum I guess you'd have to make a copy of the int range codec, since it doesn't make sense to have a max value like this outside of enchantments (I'm just saying this because I feel like I might forget if I were doing this).

src/main/java/net/minestom/server/codec/Codec.java
FluxCapacitor2: Same question for the related NetworkBuffer.Type mapValue change

FluxCapacitor2:
Vanilla MC uses Optionals for their number ranges so that the output of deserializing and then serializing the same range will equal the input. In Minestom, though, we use primitive types, so there's no way to properly represent an unbounded end of a range. (e.g. if we decoded `{"min":null,"max":10}` and encoded it again, it would become `{"min":-2147483648,"max":10}`)

If we wanted to be completely correct, we could either:
- Switch `Range` subclasses to nullable boxed versions instead of primitive types
- Use `Optional` just like MC does

Or we could use MIN_VALUE and MAX_VALUE for each type of number and just serialize them as `null`, which would be the smallest diff but I don't really like adding a special case like that.

This Range codec is used in other parts of the game as well, so it would be worth making the change there instead of making a new one specifically for data component predicate codecs.

What do you think @GoldenStack?

GoldenStack:
I think the Minestom codebase tends to use `@Nullable T` instead of `Optional<T>` so that'd probably be better. And honestly what you said makes a lot of sense, so I'm convinced that's better than what I originally suggested. So I guess I'd just recommend adding`@Nullable T`. Thanks for the detailed response! Hopefully this PR can get merged pretty soon.

FluxCapacitor2:
Agreed, and seems like matt agrees as well: https://discord.com/channels/706185253441634317/706186227493109860/1373233767753187411

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -0,0 +1,80 @@
...
+import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.registry.Registries;
+import net.minestom.server.registry.RegistryKey;
+import net.minestom.server.utils.validate.Check;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public record ComponentPredicateSet(Int2ObjectMap<DataComponentPredicate> predicates) {
```
FluxCapacitor2: Do you like that this is immutable? I was looking at DataComponentMap while implementing this

FluxCapacitor2 (commented):
@mworzala I switched ComponentPredicateType to a registry and replaced the type of `DataComponentPredicates#predicates` with a new class, `ComponentPredicateSet`. I like it better than a `Map<ComponentPredicateType, DataComponentPredicate>` because it's typesafe (no risk of key not matching the type of its value) and easier to construct than a Map.

It also gives me a nice place to put the necessary codec and network type. The way I implemented it isn't as optimized as before since it reads them into a Map and then copies it into a `ComponentPredicateSet`, but it's way more readable IMO.

I have a few questions about style below:

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
GoldenStack: Looks like this is no longer the case since d09a537ddcf5733d5650d2cba9167856faba9574

FluxCapacitor2: I definitely could write it that way, but I wanted to avoid duplicating the names of all the possible keys. I think when I wrote my own enum it was a bit questionable, but now that I'm using a registry and the built-in codecs and network types, I think it's the best solution.

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
kermandev: A defensive copy should be good here.

src/main/java/net/minestom/server/MinecraftServer.java
mworzala: So far I have exposed registries like this (ie the enchantment ones are the same). There is an open pr to convert Registries into a Registry of Registry values, in which case they will still be present.

tl;dr: yes

src/main/java/net/minestom/server/codec/Codec.java
mworzala: Having the value based on the key screams tagged union (`unionType` or a registry one) to me, not map. I can try to fiddle with it myself soon and see if I understand why its not that

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
mworzala: I would prefer to avoid publicly exposing fastutil here (ie hide behind an interface)

src/main/java/net/minestom/server/instance/block/predicate/BlockPredicate.java
```diff
@@ -88,6 +120,21 @@ public boolean test(Block block) {
...
+            return false;
+        if ((components.exact() == null || components.exact().isEmpty()) && (components.predicates() == null || components.predicates().isEmpty()))
+            return true;
+        if (block.nbt() == null)
+            return false; // If a block has no NBT (it's not a block entity), any component predicates must return false
+
+        CompoundBinaryTag componentsTag = block.nbt().getCompound("components");
+        final Transcoder<BinaryTag> coder = new RegistryTranscoder<>(Transcoder.NBT, MinecraftServer.process());
+        Result<DataComponentMap> result = DataComponent.MAP_NBT_TYPE.decode(coder, componentsTag);
+        switch (result) {
```
kermandev: orElseThrow

src/main/java/net/minestom/server/codec/Codec.java
```diff
@@ -357,7 +363,15 @@ default <V> Codec<Map<T, V>> mapValue(Codec<V> valueCodec, int maxSize) {
      */
     @Contract(pure = true)
     default <V> Codec<Map<T, V>> mapValue(Codec<V> valueCodec) {
-        return mapValue(valueCodec, Integer.MAX_VALUE);
+        return mapValue((ignored) -> valueCodec, Integer.MAX_VALUE);
+    }
+
+    default <V> Codec<Map<T, V>> mapValue(Function<T, Codec<V>> valueCodecGetter, int maxSize) {
```
kermandev: Please document these and add their contracts.

src/main/java/net/minestom/server/codec/CodecImpl.java
```diff
@@ -199,6 +199,12 @@ public <D> Result<Map<K, V>> decode(Transcoder<D> coder, D value) {
                 if (!(keyResult instanceof Result.Ok(K decodedKey)))
                     return keyResult.cast();
                 // The throwing decode here is fine since we are already iterating over known keys.
+                Codec<V> valueCodec;
```
kermandev: This could be final as it cant fall through

src/main/java/net/minestom/server/codec/CodecImpl.java
```diff
@@ -219,6 +225,12 @@ public <D> Result<D> encode(Transcoder<D> coder, @Nullable Map<K, V> value) {
                 final Result<D> keyResult = keyCodec.encode(coder, entry.getKey());
                 if (!(keyResult instanceof Result.Ok(D encodedKey)))
                     return keyResult.cast();
+                Codec<V> valueCodec;
```
kermandev: This could be final as it cant fall through

src/main/java/net/minestom/server/codec/CodecImpl.java
```diff
@@ -181,7 +181,7 @@ public <D> Result<D> encode(Transcoder<D> coder, @Nullable Set<T> value) {
         }
     }
 
-    record MapImpl<K, V>(Codec<K> keyCodec, Codec<V> valueCodec,
+    record MapImpl<K, V>(Codec<K> keyCodec, Function<K, Codec<V>> valueCodecGetter,
```
kermandev: I feel like this is getting awfully close to the RegistryUnion types here; maybe the registry type should use this under the hood? What do you think @mworzala ?

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -0,0 +1,91 @@
...
+import net.minestom.server.registry.RegistryKey;
+import net.minestom.server.utils.validate.Check;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public record ComponentPredicateSet(Map<Integer, DataComponentPredicate> predicates) {
+
+    public static final ComponentPredicateSet EMPTY = new ComponentPredicateSet();
```
kermandev: Pass the Int2ObjectMaps.emptyMap() to the Empty

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -0,0 +1,91 @@
...
+import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.registry.Registries;
+import net.minestom.server.registry.RegistryKey;
+import net.minestom.server.utils.validate.Check;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public record ComponentPredicateSet(Map<Integer, DataComponentPredicate> predicates) {
```
kermandev: Please add Contract(pure=true) to pure methods

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -2,33 +2,74 @@
...
+
+        private static final NetworkBuffer.Type<DataComponentPredicates> delegate = NetworkBufferTemplate.template(
+                DataComponent.MAP_NETWORK_TYPE.transform(Function.identity(), map -> Objects.requireNonNullElse(map, DataComponentMap.EMPTY)), DataComponentPredicates::exact,
+                ComponentPredicateSet.NETWORK_TYPE.transform(Function.identity(), map -> Objects.requireNonNullElse(map, ComponentPredicateSet.EMPTY)), DataComponentPredicates::predicates,
+                DataComponentPredicates::new
+        );
+
+        @Override
+        public void write(NetworkBuffer buffer, DataComponentPredicates value) {
+            delegate.write(buffer, Objects.requireNonNullElseGet(value, () -> new DataComponentPredicates(null, null)));
```
kermandev: Shouldnt you use EMPTY here without the supplier?

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -2,33 +2,74 @@
...
-    }
+    public static final Codec<DataComponentPredicates> CODEC = StructCodec.struct(
+            "components", DataComponent.PATCH_CODEC.optional(), DataComponentPredicates::exact,
+            "predicates", ComponentPredicateSet.CODEC.optional(), DataComponentPredicates::predicates,
+            DataComponentPredicates::new
+    );
+
+    public static final NetworkBuffer.Type<DataComponentPredicates> NETWORK_TYPE = new NetworkBuffer.Type<>() {
+
+        private static final NetworkBuffer.Type<DataComponentPredicates> delegate = NetworkBufferTemplate.template(
```
kermandev: I don't get this weird delegate when exact and predicates empty value could just be EMPTY.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -2,33 +2,74 @@
...
 
-import java.util.List;
+import java.util.*;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
-// TODO: Pending pr #2732
-public class DataComponentPredicates {
-    public static final DataComponentPredicates EMPTY = new DataComponentPredicates();
+public record DataComponentPredicates(@Nullable DataComponentMap exact,
```
kermandev: Seems weird to have these be null able when empty exists; also having these both be null is worthless meaning and should never be a attached predicate as it provides no value.

src/main/java/net/minestom/server/item/predicate/ItemPredicate.java
```diff
@@ -0,0 +1,51 @@
...
+import net.minestom.server.item.Material;
+import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.utils.Range;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public record ItemPredicate(
+        @Nullable List<Material> items,
```
kermandev: List#copyOf when its not null is required

src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
```diff
@@ -786,7 +786,7 @@ public Map<K, V> read(NetworkBuffer buffer) {
             V[] values = (V[]) new Object[size];
             for (int i = 0; i < size; i++) {
                 keys[i] = buffer.read(parent);
-                values[i] = buffer.read(valueType);
+                values[i] = buffer.read(valueTypeGetter.apply(keys[i]));
```
kermandev: Read the object into a final variable and then do set it to the key and apply

src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
```diff
@@ -766,14 +766,14 @@ public S read(NetworkBuffer buffer) {
...
-    record MapType<K, V>(Type<K> parent, NetworkBuffer.Type<V> valueType,
+    record MapType<K, V>(Type<K> parent, Function<K, NetworkBuffer.Type<V>> valueTypeGetter,
                          int maxSize) implements NetworkBufferTypeImpl<Map<K, V>> {
         @Override
         public void write(NetworkBuffer buffer, Map<K, V> map) {
             buffer.write(VAR_INT, map.size());
             for (Map.Entry<K, V> entry : map.entrySet()) {
                 buffer.write(parent, entry.getKey());
-                buffer.write(valueType, entry.getValue());
+                buffer.write(valueTypeGetter.apply(entry.getKey()), entry.getValue());
```
kermandev: Read into final like below

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -2,33 +2,74 @@
...
+                                      @Nullable ComponentPredicateSet predicates) implements Predicate<DataComponent.Holder> {
 
-    public static final NetworkBuffer.Type<DataComponentPredicates> NETWORK_TYPE = NetworkBufferTemplate.template(
-            NetworkBuffer.UNIT.list(), DataComponentPredicates::exact,
-            NetworkBuffer.UNIT.list(), DataComponentPredicates::partial,
-            DataComponentPredicates::new);
-    public static final Codec<DataComponentPredicates> CODEC = StructCodec.struct(new DataComponentPredicates());
+    public static final DataComponentPredicates EMPTY = new DataComponentPredicates(null, null);
```
kermandev: See above, but this could just be `DataComponentMap.EMPTY, ComponentPredicateSet.EMPTY` unless there some reason im not understanding on this.

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicates.java
```diff
@@ -2,33 +2,74 @@
...
+            }
+            return new DataComponentPredicates(exact, predicates);
+        }
+    };
 
-    private List<Unit> partial() {
-        return List.of();
+    @Override
+    public boolean test(DataComponent.Holder holder) {
+        if (exact != null && !exact.entrySet().stream().allMatch(entry -> Objects.equals(holder.get(entry.component()), entry.value()))) {
```
kermandev: No streams in test please

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,574 @@
...
+    record Container(@Nullable CollectionPredicate<ItemStack, ItemPredicate> items) implements DataComponentPredicate {
+        public static final Codec<Container> CODEC = StructCodec.struct(
+                "items", CollectionPredicate.codec(ItemPredicate.CODEC).optional(), Container::items,
+                Container::new
+        );
+
+        @Override
+        public boolean test(DataComponent.Holder holder) {
+            List<ItemStack> itemStacks = holder.get(DataComponents.CONTAINER);
+            return items == null || items.test(itemStacks != null ? itemStacks.stream().filter(item -> !item.isAir()).toList() : List.of());
```
kermandev: no streams in tests please

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,158 @@
...
+                return StructCodec.struct(
+                        "test", itemCodec, Entry::predicate,
+                        "count", DataComponentPredicates.INT_RANGE_CODEC, Entry::count,
+                        Entry::new
+                );
+            }
+
+            @Override
+            public boolean test(Collection<T> collection) {
+                return this.count.inRange((int) collection.stream().filter(this.predicate).count());
```
kermandev: No streams in test

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,574 @@
...
+            if (enchantments != null) {
+                for (RegistryKey<Enchantment> key : enchantments) {
+                    if (enchantmentList.has(key) && (levels == null || levels.inRange(enchantmentList.level(key)))) {
+                        return true;
+                    }
+                }
+                return false;
+            } else if (levels != null) {
+                // If `enchantments` is not specified, the predicate returns true when any enchantment matches the specified `level`
+                return enchantmentList.enchantments().entrySet().stream().anyMatch(entry -> levels.inRange(entry.getValue()));
```
kermandev: No stream in tests

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,574 @@
...
+
+        public Enchantments(@Nullable RegistryTag<Enchantment> enchantments,
+                            @Nullable Range.Int levels) {
+            this(List.of(new EnchantmentListPredicate(enchantments, levels)));
+        }
+
+        @Override
+        public boolean test(DataComponent.Holder holder) {
+            EnchantmentList enchantments = holder.get(DataComponents.ENCHANTMENTS);
+            return enchantments != null && children.stream().allMatch(enchantment -> enchantment.test(enchantments));
```
kermandev: no streams in tests

src/main/java/net/minestom/server/instance/block/predicate/DataComponentPredicate.java
```diff
@@ -0,0 +1,574 @@
...
+
+        public StoredEnchantments(@Nullable RegistryTag<Enchantment> enchantments,
+                                  @Nullable Range.Int levels) {
+            this(List.of(new EnchantmentListPredicate(enchantments, levels)));
+        }
+
+        @Override
+        public boolean test(DataComponent.Holder holder) {
+            EnchantmentList enchantments = holder.get(DataComponents.STORED_ENCHANTMENTS);
+            return enchantments != null && children.stream().allMatch(enchantment -> enchantment.test(enchantments));
```
kermandev: no streams in tests

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,158 @@
...
+            for (Entry<T, P> entry : entries) {
+                if (!entry.test(collection)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }
+
+    public static <T, P extends Predicate<T>> CollectionPredicate.Builder<T, P> builder() {
```
kermandev, line 125: Contract(pure=true)

src/main/java/net/minestom/server/instance/block/predicate/CollectionPredicate.java
```diff
@@ -0,0 +1,158 @@
...
+        return new CollectionPredicate.Builder<>();
+    }
+
+    public static class Builder<T, P extends Predicate<T>> {
```
kermandev: final

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -0,0 +1,91 @@
...
+        Map<RegistryKey<Codec<? extends DataComponentPredicate>>, DataComponentPredicate> map = new HashMap<>(this.predicates.size());
+        for (Map.Entry<Integer, DataComponentPredicate> entry : predicates.entrySet()) {
+            var key = MinecraftServer.componentPredicateTypes().getKey(entry.getKey());
+            map.put(key, entry.getValue());
+        }
+        return map;
+    }
+
+    private static ComponentPredicateSet fromMap(Map<RegistryKey<Codec<? extends DataComponentPredicate>>, DataComponentPredicate> input) {
+        ComponentPredicateSet map = new ComponentPredicateSet(new Int2ObjectArrayMap<>(input.size()));
```
kermandev: Sadly this doesnt work the way you want, as it will just clear the input size after the first add

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -0,0 +1,91 @@
...
+        for (DataComponentPredicate predicate : input.values()) {
+            map = map.add(predicate);
+        }
+        return map;
+    }
+
+    public ComponentPredicateSet add(DataComponentPredicate predicate) {
+        var newMap = new Int2ObjectArrayMap<>(predicates);
```
kermandev: this will probably cause a rehashing (after doubling) each time as predicates might be to size already

kermandev (changes requested):
I was mainly focusing on doing a short review on changes involving codecs, but Ive gone ahead and looked through most of it and shared some comments below.

Most of them stem from avoid usage of streams inside test predicates; as they are kind of unnecessary and you seem to already use the alternative in a few places.

For immutable objects you may want to put withX as it could be useful to users, and make sure to add pure contracts for ide warnings.

Also for predicates it seems really bad if everything was null or empty as the predicate would be worthless at that point. Maybe you should throw in the constructor if that happens.

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -10,19 +11,40 @@
...
 
-public record ComponentPredicateSet(Map<Integer, DataComponentPredicate> predicates) {
+/**
+ * A list of {@link DataComponentPredicate}s.
+ * <p>
+ * Note: instances of this class are immutable. Calling {@link #add} or {@link #remove}
+ * will return a new instance of this class with the element added or removed.
+ *
+ */
+public class ComponentPredicateSet implements Predicate<DataComponent.Holder> {
```
kermandev: Missing final if this continues to not be a record (see below)

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -10,19 +11,40 @@
...
     public static final ComponentPredicateSet EMPTY = new ComponentPredicateSet();
 
+    /**
+     * A map of DataComponentPredicate registry entry IDs to their values
+     */
+    private final Map<Integer, DataComponentPredicate> predicates;
+
     public ComponentPredicateSet() {
-        this(new Int2ObjectArrayMap<>());
+        this(Int2ObjectMaps.emptyMap());
```
kermandev: This makes no sense to exist as its equivalent to .EMPTY, if anything it should just be a vararg on the delegate of Collections version.

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -10,19 +11,40 @@
...
+    public ComponentPredicateSet(Collection<DataComponentPredicate> predicates) {
+        this.predicates = new Int2ObjectArrayMap<>(predicates.size());
+        for (DataComponentPredicate predicate : predicates) {
+            var key = MinecraftServer.componentPredicateTypes().getKey(predicate.codec());
+            Check.notNull(key, "Unknown DataComponentPredicate type");
+            this.predicates.put(MinecraftServer.componentPredicateTypes().getId(key), predicate);
+        }
+    }
+
+    private ComponentPredicateSet(Map<Integer, DataComponentPredicate> predicates) {
```
kermandev: I think its fine to keep this a record, just ensure a `predicates = Int2ObjectMaps.unmodifiable(new Int2ObjectArrayMap<>(predicates))` is performed.

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
```diff
@@ -10,19 +11,40 @@
...
+
     public ComponentPredicateSet() {
-        this(new Int2ObjectArrayMap<>());
+        this(Int2ObjectMaps.emptyMap());
     }
 
-    public ComponentPredicateSet(Map<Integer, DataComponentPredicate> predicates) {
-        this.predicates = new Int2ObjectArrayMap<>(predicates);
+    public ComponentPredicateSet(Collection<DataComponentPredicate> predicates) {
+        this.predicates = new Int2ObjectArrayMap<>(predicates.size());
```
kermandev: If your moving this to a record, you can make this implementation into a private static to return a new Int2ObjectArrayMap

kermandev (changes requested):

src/main/java/net/minestom/server/instance/block/predicate/ComponentPredicateSet.java
FluxCapacitor2: I switched it to a regular class because it makes it possible for me to have a private constructor that doesn't copy the input map. This allows the `add` and `remove` methods to only copy the map once instead of twice.

A regular class also makes it possible to completely hide the Map from the API client. This class really should behave more like a Set, and the Map should be considered an implementation detail that's only used during serialization.

Let me know what you think.

InfiniteCoder01:
What's the current state of this PR? I checked it out with Trove (I am implementing entity predicates for sheep wool drop), seems to work (of course, this isn't a good test, but it shows that this PR is useful).

GoldenStack:
I reviewed this a few times but I honestly forgot what the blocker is. It might have been waiting on Matt having time and wanting to review it.

InfiniteCoder01:
Just checked it on 26.1, works wonderfully.
Is there anything I can do to help this get merged? It seems like this didn't go anywhere since September...

kermandev (approved):

PR #2735, opened 2025-05-08
Chore: Misc Changes and Improvements

kermandev:
Multiple changes for reasons described in each commit message, these are smaller changes that don't belong in their own PR as they would be too small. 

## Types of changes

## Checklist

## Further comments

All reasoning is provided in commit messages; their likeness to break ill behavior is labeled in the commit title.
Note: Breaking-ish is when things are marked as `@ApiStatus.Internal` have changed.


kermandev:
Additional small pr's that could fall under this are #2711 #2723 #2727

src/main/java/net/minestom/server/entity/MetadataImpl.java
```diff
@@ -16,52 +16,50 @@
...
-        EMPTY_VALUES.set(TYPE_NBT, NBT(EndBinaryTag.endBinaryTag()));
-        EMPTY_VALUES.set(TYPE_PARTICLE, Particle(Particle.DUST));
-        EMPTY_VALUES.set(TYPE_PARTICLE_LIST, ParticleList(List.of()));
-        EMPTY_VALUES.set(TYPE_VILLAGERDATA, VillagerData(VillagerMeta.VillagerData.DEFAULT));
-        EMPTY_VALUES.set(TYPE_OPT_VARINT, OptVarInt(null));
-        EMPTY_VALUES.set(TYPE_POSE, Pose(EntityPose.STANDING));
-        EMPTY_VALUES.set(TYPE_CAT_VARIANT, CatVariant(CatMeta.Variant.TABBY));
-        EMPTY_VALUES.set(TYPE_WOLF_VARIANT, WolfVariant(WolfVariant.PALE));
-        EMPTY_VALUES.set(TYPE_FROG_VARIANT, FrogVariant(FrogMeta.Variant.TEMPERATE));
+    private static final Map<Byte, Metadata.Entry<?>> EMPTY_VALUES = Map.ofEntries(
```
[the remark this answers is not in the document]
kermandev: Yes, Map is slower in testing, changed to List instead. Maybe the ObjectArray class should have a toList method added if grown from a zero index.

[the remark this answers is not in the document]
kermandev: I still might revert the change back if I do that as I don't like how the ordering is the only thing that is keeping the index the same on that.

src/main/java/net/minestom/server/network/packet/server/play/data/WorldPos.java
```diff
@@ -12,6 +12,7 @@ public record WorldPos(@NotNull String dimension, @NotNull Point blockPosition)
             NetworkBuffer.STRING, WorldPos::dimension,
             NetworkBuffer.BLOCK_POSITION, WorldPos::blockPosition,
             WorldPos::new);
+    public static final NetworkBuffer.Type<WorldPos> OPT_NETWORK_TYPE = NETWORK_TYPE.optional();
```
mworzala, line 15: Not a fan, I dont want to start bloating all classes network types or codecs which may be used in an optional context. Generally id only like to define a second constant if its a fundamentally different codec, otherwise the single allocation of a second optional wrapper is acceptable.

src/main/java/net/minestom/server/entity/MetadataImpl.java
```diff
@@ -26,42 +25,39 @@
...
-        EMPTY_VALUES.set(TYPE_FROG_VARIANT, FrogVariant(FrogMeta.Variant.TEMPERATE));
-        // OptGlobalPos
-        EMPTY_VALUES.set(TYPE_PAINTING_VARIANT, PaintingVariant(new Holder.Reference<>(PaintingVariant.KEBAB)));
-        EMPTY_VALUES.set(TYPE_SNIFFER_STATE, SnifferState(SnifferMeta.State.IDLING));
-        EMPTY_VALUES.set(TYPE_ARMADILLO_STATE, ArmadilloState(ArmadilloMeta.State.IDLE));
-        EMPTY_VALUES.set(TYPE_VECTOR3, Vector3(Vec.ZERO));
-        EMPTY_VALUES.set(TYPE_QUATERNION, Quaternion(new float[]{0, 0, 0, 0}));
-        EMPTY_VALUES.trim();
-    }
+    private static final List<Metadata.Entry<?>> EMPTY_VALUES = List.of(
```
mworzala, line 28: I much prefer the older explicit mapping of type to ID. Debugging an id offset by 1 is more difficulty vs an entirely missing value.

src/main/java/net/minestom/server/entity/Metadata.java
```diff
@@ -163,6 +163,10 @@ public static Entry<DynamicRegistry.Key<ChickenVariant>> ChickenVariant(@NotNull
         return new MetadataImpl.EntryImpl<>(TYPE_CHICKEN_VARIANT, value, ChickenVariant.NETWORK_TYPE);
     }
 
+    public static Entry<WorldPos> OptionalWorldPosition(@Nullable WorldPos value) {
```
mworzala, line 166: Unused does not need a definition, also should be called OptWorldPos inline with other optionals and the type name

src/main/java/net/minestom/server/entity/Metadatas.java
```diff
@@ -0,0 +1,51 @@
+package net.minestom.server.entity;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+// TODO(autogenerate)
+sealed interface Metadatas permits Metadata {
```
mworzala, line 6: MetadataTypes or something, Metadatas is very general. Also, seems like a generally unnecessary change these aren't intended to be used externally and its not particularly long.

mworzala (commented):
Generally I don't like this kind of PR. The changes are generally irrelevant and having to sort through some agreement and some not is not super productive. Please refrain from this type of small (essentially) formatting/opinions pr in the future.

src/main/java/net/minestom/server/entity/Metadatas.java
kermandev: Reverted this change

src/main/java/net/minestom/server/entity/Metadata.java
kermandev: Changed to OptWorldPos

src/main/java/net/minestom/server/entity/MetadataImpl.java
kermandev: Changed to use an ObjectArray, then froze the results in a List

kermandev:
Dropped all registry changes in favor of #2754

PR #2739, opened 2025-05-11
Add slot in ItemDropEvent

Andrei965:
## Proposed changes

This PR adds the slot from which the item was dropped to the ItemDropEvent

## Types of changes

## Checklist

merging your code._

## Further comments

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -1275,9 +1275,9 @@ public class Player extends LivingEntity implements CommandSender, HoverEventSou
      * @param item the item to drop
      * @return true if player can drop the item (event not cancelled), false otherwise
      */
-    public boolean dropItem(@NotNull ItemStack item) {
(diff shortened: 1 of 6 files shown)


Andrei965:
Hi! I do agree with you

> makes slot information mandatory for all drops, even when it doesn't make sense to have it

But connecting InventoryPreClickEvent with the drop seems like a work around. I do still think that the slot is important for the ItemDropEvent. Maybe there is another way?

GoldenStack:
> But connecting InventoryPreClickEvent with the drop seems like a work around. I do still think that the slot is important for the ItemDropEvent. Maybe there is another way?

It's not a workaround; if you need to know when the player drops the item manually, and you need more information about it, you just listen to the more relevant event, and then check if it's the type you need.

Andrei965:
You're right! I'll leave this PR open tho in case a maintainer wants to merge it

mworzala:
I am generally in agreement with golden that pre click event is more relevant if you need the specifics about inventory drops. Ill close for now but happy to discuss more if anyone thinks there is more here.

Issue #2740, opened 2025-05-14
[Feature]: Change EntityPotionAddEvent to be cancelable


EpicBirb:
### Is your feature request related to a problem?

So I am making a game based of another game because I'm bored as heck and I've come to realize it would probably be a pain to implement a check everywhere in the code to see whether the player is eligible for a specific effect.

### Describe the solution you'd like

To keep it short, make EntityPotionAddEvent cancelable.

### Describe alternatives you've considered

I mean the alternative solution was to put the same code over and over again to check.
I did do an implementation of the feature request on a fork: https://github.com/EpicBirb/Minestom

mudkipdev:
If you open a PR, make sure it is not from a `master` branch or else it will automatically close. Also, [this](https://javadoc.minestom.net/net/minestom/server/event/EventDispatcher.html#callCancellable\(net.minestom.server.event.trait.CancellableEvent,java.lang.Runnable\)) might be helpful.

EpicBirb:
> If you open a PR, make sure it is not from a `master` branch or else it will automatically close. Also, [this](https://javadoc.minestom.net/net/minestom/server/event/EventDispatcher.html#callCancellable(net.minestom.server.event.trait.CancellableEvent,java.lang.Runnable)) might be helpful.

ok thanks

EpicBirb:
Closed #2742

PR #2741, opened 2025-05-14
Cancelable EntityPotionAddEvent

EpicBirb:
## Proposed changes

This is a feature request linked to #2740

## Types of changes

## Checklist

## Further comments

I lack the knowledge to actually put the code into a testing environment but I believe the implementation is correct since it was just some code rearranging and copying code from other events to serve as a foundation for me to format the code properly.

### DIFF
--- a/.github/workflows/javadoc.yml
+++ b/.github/workflows/javadoc.yml
@@ -8,22 +8,25 @@ jobs:
   build:
     runs-on: ubuntu-latest
     name: Build and deploy Javadoc
+    env:
+      JAVA_VERSION: 23
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up JDK 21
-        uses: actions/setup-java@v1
+      - uses: actions/checkout@v4
+      - name: Set up JDK ${{ env.JAVA_VERSION }}
+        uses: actions/setup-java@v4
         with:
-          java-version: 21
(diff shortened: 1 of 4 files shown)


cody-quinn:
Should be noted this is a breaking change. The event is now called before the effect is applied, as opposed to after.

Perhaps instead of changing this event a new pre event, `PreEntityPotionAddEvent`, could be added.

EpicBirb:
> Should be noted this is a breaking change. The event is now called before the effect is applied, as opposed to after.
> 
> Perhaps instead of changing this event a new pre event, `PreEntityPotionAddEvent`, could be added.

Ill make a change tomorrow. I'm tired for the day

EpicBirb:
I felt like I made a mistake, so I dropped some commits and readded the changes, my bad

mworzala (changes requested):
I think I would generally like to move away from the `Pre` label on events. We aren't very consistent about it (ie most cancellable events are not `Pre` labelled, and when we have both the non-cancellable one is often mostly irrelevant). (cc @cody-quinn)

Please target to the 1.21.5 branch as this is a breaking change (I would like to get it in before that branch). Additionally, please write some tests for this (there are many examples in the tests directory).

EpicBirb:
moved to #2742

PR #2754, opened 2025-06-02
Misc Registry Changes/Improvements

kermandev:
## Proposed changes

See the commit description; Some of the code is similar to my misc changes and improvements branch

## Types of changes

## Checklist

merging your code._

## Further comments

Tests might be required for some changes, like ObjectArray
Check's change could be rolled back, it is kind of extraneous as it could just throw IllegalState...


src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -99,18 +127,12 @@ public int getId(@NotNull RegistryKey<T> key) {
...
-//                "Registering an object to a dynamic registry ({0}) after the server is started can lead to " +
-//                        "registry desync between the client and server. This is usually unwanted behavior. If you " +
-//                        "know what you're doing and would like this behavior, set the `minestom.registry.late-register` " +
-//                        "system property.", id);
-
-        lock.lock();
-        try {
-            final RegistryKey<T> registryKey = new RegistryKeyImpl<>(key);
-            int id = idToKey.indexOf(registryKey);
+        if (canCompress() && MinecraftServer.process() != null && !MinecraftServer.isStarted())
```
mworzala: isn't the last case inverted?

src/main/java/net/minestom/server/registry/RegistryData.java
```diff
@@ -43,6 +45,7 @@
  */
 public final class RegistryData {
```
mworzala, line 54: honestly didnt read much into the changes to this file, but:
* This should all remain mutable is used exclusively for parsing the big static registry json files so speed is kind of a relevant factor
* seems like kinda just reordering thats unnecessary for pure internals

mworzala (commented):
I need to read through the registry change a bit more but some comments on the other bits

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
kermandev: Oops, yes that an issue thanks

src/main/java/net/minestom/server/registry/RegistryData.java
kermandev: From my testing, the only slowdown was using the merge map as it would take two immutable maps and combine them, causing it to recalculate hash keys for each of them. I've gone ahead and just replaced this. Maybe a nicer solution would be to expose a withX API for mutating the static registry data. For example, want to make the run speed of mobs faster on a custom block, just by changing the speedFactor? Minestom could support that.

PR #2759, opened 2025-06-07
1.21.7 Support

mworzala:
Tracking branch for 1.21.6 support. Will be marked ready after #2699 is merged.


davidmayr:
In MetadataDef the Pig definition is wrong. Saddle is no longer present in vanilla, and that causes the offsets to be wrong. This probably applies to all other saddleable entities as well (since it's now an item slot), but I have not checked that in vanilla yet.

Edit: Just checked: Strider doesn't have the field anymore either, and in abstract horse it is no longer in the flags. The flag indices have not changed, but the one for saddle is just unused now

PR #2765, opened 2025-06-17
Block#state string

TheMode:
Added `Block#state`/`Block#fromState` string.

Currently a draft because I am not sure what to do about the `ArgumentSyntaxException` thrown on parsing which currently create a weird dependency. The alternative would be to duplicate the code for `Block`/`ArgumentBlockState`


src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -204,6 +218,10 @@ default boolean compare(@NotNull Block block) {
         return BlockImpl.REGISTRY.get(key);
     }
 
+    static @Nullable Block fromState(String state) {
```
mworzala: NotNull on state

PR #2768, opened 2025-06-17
Prepare for the removal of TagStringIOExt

kermandev:
## Proposed changes

Adventure 4.22.0 now has TagStringIOExt upstreamed; they might be more verbose, but they are accessible now.

## Types of changes

- [x] Future Breaking change (fix or future feature that would cause existing functionality to not work as expected)

## Checklist

merging your code._

## Further comments

Future breaking change in 1.21.6, will drop the class after this is finalized and merged.


mworzala (changes requested):
We should define a constant TagStringIO which has heterogeneous lists enabled. We are on a mc version which supports them so it should be assumed/the default.

kermandev:
Delayed until we get full heterogeneous list support in Adventure, some methods are missing configuration.

src/main/java/net/minestom/server/command/builder/arguments/minecraft/ArgumentItemStack.java
```diff
@@ -36,26 +38,42 @@ public class ArgumentItemStack extends Argument<ItemStack> {
     public static final int INVALID_MATERIAL = 3;
     public static final int INVALID_COMPONENT = 4;
 
+    private final TagStringIO tagStringIO;
+
     public ArgumentItemStack(String id) {
         super(id, true);
+        this.tagStringIO = MinestomAdventure.tagStringIO();
```
mworzala: Is there any reason not to just use the global here? I dont think _not_ wanting heterogeneous lists is a super relevant use case

src/main/java/net/minestom/server/command/builder/arguments/minecraft/ArgumentItemStack.java
```diff
@@ -149,13 +167,14 @@ public void consume(char c) {
             return Key.key(input.substring(start, index));
         }
 
-        public @NotNull BinaryTag readTag() {
+        public @NotNull BinaryTag readTag(TagStringIO tagStringIO) {
```
mworzala: Missing annotation

kermandev: Private class, but I'll add it anyway.

src/main/java/net/minestom/server/command/builder/arguments/minecraft/ArgumentItemStack.java
kermandev: I'd wanted to give as much flexibility on options you would want for parsing these; It should probably follow the client (for clientside error detection), so I can remove this if you want.

mworzala: Yeah lets follow the client

PR #2777, opened 2025-06-19
Tag<Boolean> setter shortcut

TheMode:
Add convenient functions for setting `Tag<Boolean>` to true.

Perhaps confusing as the getter version (without tag default value) can return null and not only true/false.


mworzala:
I do think the mismatch between set and then future get is a little confusing, it feels like if anything we should add a `setBool`/`getBool`, but then that begs to add all other primitives which I don't really want to do...

mworzala:
Closing, feel free to reopen with more discussion.

Issue #2757, opened 2025-06-05
[Bug]: Network Protocol Error when setting biomes on a per-block basis


Bloeckchengrafik:
### Affected Version

1_21_5-4d91778331

### Describe the bug

When using a custom generator to set biomes at a fine-grained level (e.g., for every 4x4x4 block volume), clients attempting to load the chunk are disconnected with a "Connection Lost: Network Protocol Error".

Additional info by TheMode:
> happening when the list has >=8 biomes

### Steps to reproduce the bug

- Create a new InstanceContainer.
- Set a generator for the instance.
- Inside the generator, loop through the chunk's coordinates with a step of 4 for X, Y, and Z.
- In the inner loop, call unit.modifier().setBiome(x, y, z, randomBiome) to assign a biome to that 4x4x4 section.
- Start the server and attempt to join.

### Code sample

```java
import net.minestom.server.MinecraftServer;
import net.minestom.server.instance.InstanceContainer;
import net.minestom.server.instance.InstanceManager;
import net.minestom.server.instance.block.Block;
import net.minestom.server.utils.math.Point;
import net.minestom.server.world.biome.Biome;
import net.minestom.server.world.biome.DynamicRegistry;

import java.util.List;
import java.util.Random;

public class BiomeGeneratorExample {

    public Instance createInstanceWithCustomBiomes() {
        final InstanceManager instanceManager = MinecraftServer.getInstanceManager();
        final InstanceContainer instance = instanceManager.createInstanceContainer();
        final Random random = new Random();

        final List<DynamicRegistry.Key<Biome>> biomes = List.of(
                Biome.PLAINS,
                Biome.DESERT,
                Biome.FOREST,
                Biome.TAIGA,
                Biome.SWAMP,
                Biome.MUSHROOM_FIELDS,
                Biome.JUNGLE,
                Biome.DARK_FOREST,
                Biome.BIRCH_FOREST,
                Biome.RIVER,
                Biome.BEACH,
                Biome.LUKEWARM_OCEAN
        );

        instance.setGenerator(unit -> {
            final var mod = unit.modifier();
            mod.fillHeight(0, 20, Block.STONE);

            final Point start = unit.absoluteStart();
            final int blockX = start.blockX();
            final int blockZ = start.blockZ();
            final int minY = start.blockY();
            final int maxY = unit.absoluteEnd().blockY();

            for (int y = minY + 4; y < maxY - 4; y += 4) {
[...]

PR #2780, opened 2025-06-21
Palette API cleanup

TheMode:
Single implementation for the sealed Palette interface.

`AdaptivePalette` was sort of a mistake, added a lot of unnecessary code and redirection.

Added a newly exposed `Palette#optimize` (I welcome name suggestions for the enum)

Opening as draft as I may want to add some more low-level methods and perhaps ensuring no performance regression.


TheMode:
Performance is identical, fixed an additional deserializer bug

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -2,87 +2,102 @@
...
- * Palette able to take any value anywhere. May consume more memory than required.
- */
-final class PaletteIndirect implements SpecializedPalette, Cloneable {
+final class PaletteImpl implements Palette {
     private static final ThreadLocal<int[]> WRITE_CACHE = ThreadLocal.withInitial(() -> new int[4096]);
 
-    // Specific to this palette type
-    private final int dimension;
-    private final int maxBitsPerEntry;
+    static final int BLOCK_DIMENSION = 16;
```
kermandev: You probably could pull these out of the impl and into Palette.

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -2,87 +2,102 @@
...
         this.values = values;
 
-        this.paletteToValueList = new IntArrayList(palette.length);
-        this.valueToPaletteMap = new Int2IntOpenHashMap(palette.length);
-        this.valueToPaletteMap.defaultReturnValue(-1);
-
-        for (int i = 0; i < palette.length; i++) {
-            this.paletteToValueList.add(palette[i]);
-            this.valueToPaletteMap.put(palette[i], i);
+        if (hasPalette()) {
```
kermandev, line 45: Whats the memory footprint of this new impl? You could probably check this in https://github.com/openjdk/jol

TheMode:
May solve #2757

kermandev (commented):
Could you run some JMH benchmarks and post them here with master vs here, to see what exactly speedup it is. Also what is the memory footprint increase, cause there would be one from pulling away from the speicalized single, requiring nullptr padding.

mworzala (approved):

PR #2782, opened 2025-06-22
Add InstanceBlockUpdateEvent

TheMode:
Add a new `InstanceBlockUpdateEvent` event called for each block changes initiated from the instance. Useful if you want to know about placement rules.

Added an additional `InstanceSectionInvalidateEvent` currently not called anywhere, which I am seeing as an opaque event for users to use when a large part of the section has been invalidated bypassing the instance (e.g. palette manipulation).

An `InstancePreBlockUpdateEvent` cancellable event could probably be added if deemed necessary.

Additionally added a few coord converter funtions in `CoordConversion`, mostly serving as an helper for `MultiBlockChangePacket`.
Some more are available on my project (like section indexes) but I couldn't find a way to use it in the main codebase.


DersWasTaken:
Hmmm. I think I've seen this somewhere before :P

PR #2787, opened 2025-06-23
Fix: Correct return type of PlayerChangeHeldSlotEvent#getOldSlot()

Skubiak0903:
Changed the return type of PlayerChangeHeldSlotEvent#getOldSlot() from int to byte for consistency


mworzala:
Realized I could do it myself :)

PR #2790, opened 2025-06-24
Registry changes in preinitalization and freezing

kermandev:
Allows the server to "detour" the registry initialization back to the user. For example, when registering dialogs, you can instead register a detour and have the server freeze the dynamic registry. This option will always be open because no players can be connected before initialization is complete, but the detour must be registered before the initial call to #init, which can lead to some class loading issues if done wrong. 

## Further comments
Tests are required before merging.
This PR might include guard clauses in static registries to prevent their initialization before the class is "unsealed (Server initializing/initalized)"


mworzala:

mworzala:

kermandev:
Superseded by #2951 and future server builder API.

PR #2792, opened 2025-06-25
fix(registry): Allow nullable pack in DynamicRegistry with unsafeOps

kermandev:
## Proposed changes

Fixes bug with null packs in copyOf with unsafeops.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
+++ b/src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
@@ -140,11 +140,11 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
             if (id == -1) {
                 idToValue.add(object);
                 idToKey.add(registryKey);
-                packById.add(pack);
+                if (pack != null) packById.add(pack); // Dont add null packs, handled by getPack
             } else {
                 idToValue.set(id, object);
                 idToKey.set(id, registryKey);
-                packById.set(id, pack);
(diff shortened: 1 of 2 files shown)


mworzala:
Requiring these at the end is a super gross handling of this case I think, why not just make an immutable view of the list instead or something?

kermandev:
I did think of that, but for registry freezing I would rather not have to deal with long term storage of null entries instead of a immutable list view, I did think of another solution but I think it would cause more issues by hiding this behavior (changing id) by popping off the null entries and adding them to the end. If your fine with the unpredictable behavior inherently dealing with unsafe registry ops, where the entry id/index is not guaranteed until all registration is finished, I could implement this behavior.

mworzala:
That also feels like worse confusing behavior. We could perhaps require a pack entry and have a default `unnamed:default` or something

kermandev:
That would work; But it still seems like having null though, but currently I dont think Minestom allows you to create packs at all; Maybe this api needs work?

mworzala:
yeah its definitely a bit of a work in progress. I am happy to have null but introducing a restriction of nulls at the end only just because we want to use List.copyOf is very dumb

kermandev:
I think thats acceptable behavior currently as there is no way to register datapacks; and all registers of custom objects will appear at the end regardless; using the add behavior. I was kind of basing this pr on the fact that we already do bound checks for getPack, so why not make it a simple bound check for any null entries?

PR #2795, opened 2025-06-27
SetBlock Ordering & Cleanup

Garrett-EMS:
## Proposed changes

todo

## Types of changes

## Checklist

merging your code._

## Further comments

todo

### DIFF
--- a/demo/src/main/java/net/minestom/demo/block/placement/DripstonePlacementRule.java
+++ b/demo/src/main/java/net/minestom/demo/block/placement/DripstonePlacementRule.java
@@ -3,7 +3,11 @@ package net.minestom.demo.block.placement;
 import net.minestom.server.coordinate.Point;
 import net.minestom.server.coordinate.Vec;
 import net.minestom.server.instance.block.Block;
+<<<<<<< HEAD
 import net.minestom.server.instance.block.BlockFace;
+=======
+import net.minestom.server.instance.block.BlockChange;
+>>>>>>> cc02c79fb (Cleanup)
 import net.minestom.server.instance.block.rule.BlockPlacementRule;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -20,8 +24,16 @@ public class DripstonePlacementRule extends BlockPlacementRule {
     }
 
     @Override
+<<<<<<< HEAD
     public @Nullable Block blockPlace(@NotNull PlacementState placementState) {
         var blockFace = Objects.requireNonNullElse(placementState.blockFace(), BlockFace.TOP);
+=======
+    public @NotNull Block blockPlace(@NotNull BlockChange mutation) {
+        if (!(mutation instanceof BlockChange.Player mut)) {
+            return mutation.block(); // not a player placement
+        }
+        var blockFace = mut.blockFace();
+>>>>>>> cc02c79fb (Cleanup)
         var direction = switch (blockFace) {
             case TOP -> "up";
             case BOTTOM -> "down";
@@ -35,10 +47,17 @@ public class DripstonePlacementRule extends BlockPlacementRule {
     }
 
     @Override
+<<<<<<< HEAD
     public @NotNull Block blockUpdate(@NotNull UpdateState updateState) {
         var direction = updateState.currentBlock().getProperty(PROP_VERTICAL_DIRECTION).equals("up");
         var newThickness = getThickness(updateState.instance(), updateState.blockPosition(), direction);
         return updateState.currentBlock().withProperty(PROP_THICKNESS, newThickness);
+=======
+    public @NotNull Block blockUpdate(@NotNull BlockChange mutation) {
+        var direction = mutation.block().getProperty(PROP_VERTICAL_DIRECTION).equals("up");
+        var newThickness = getThickness(mutation.instance(), mutation.blockPosition(), direction);
+        return mutation.block().withProperty(PROP_THICKNESS, newThickness);
+>>>>>>> cc02c79fb (Cleanup)
     }
 
     private @NotNull String getThickness(@NotNull Block.Getter instance, @NotNull Point blockPosition, boolean direction) {
--- a/demo/src/main/java/net/minestom/demo/block/placement/DripstonePlacementRule.java
+++ b/demo/src/main/java/net/minestom/demo/block/placement/DripstonePlacementRule.java
@@ -3,14 +3,9 @@ package net.minestom.demo.block.placement;
 import net.minestom.server.coordinate.Point;
 import net.minestom.server.coordinate.Vec;
 import net.minestom.server.instance.block.Block;
-<<<<<<< HEAD
-import net.minestom.server.instance.block.BlockFace;
-=======
 import net.minestom.server.instance.block.BlockChange;
->>>>>>> cc02c79fb (Cleanup)
 import net.minestom.server.instance.block.rule.BlockPlacementRule;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
 import java.util.Map;
 import java.util.Objects;
@@ -24,22 +19,17 @@ public class DripstonePlacementRule extends BlockPlacementRule {
     }
 
     @Override
-<<<<<<< HEAD
-    public @Nullable Block blockPlace(@NotNull PlacementState placementState) {
-        var blockFace = Objects.requireNonNullElse(placementState.blockFace(), BlockFace.TOP);
-=======
-    public @NotNull Block blockPlace(@NotNull BlockChange mutation) {
-        if (!(mutation instanceof BlockChange.Player mut)) {
-            return mutation.block(); // not a player placement
+    public @NotNull Block blockPlace(@NotNull BlockChange blockChange) {
+        if (!(blockChange instanceof BlockChange.Player mut)) {
+            return blockChange.block(); // not a player placement
         }
         var blockFace = mut.blockFace();
->>>>>>> cc02c79fb (Cleanup)
         var direction = switch (blockFace) {
             case TOP -> "up";
             case BOTTOM -> "down";
-            default -> Objects.requireNonNullElse(placementState.cursorPosition(), Vec.ZERO).y() < 0.5 ? "up" : "down";
+            default -> Objects.requireNonNullElse(mut.player().getPosition().direction(), Vec.ZERO).y() < 0.5 ? "up" : "down";
         };
-        var thickness = getThickness(placementState.instance(), placementState.placePosition(), direction.equals("up"));
+        var thickness = getThickness(mut.instance(), mut.blockPosition(), direction.equals("up"));
         return block.withProperties(Map.of(
                 PROP_VERTICAL_DIRECTION, direction,
                 PROP_THICKNESS, thickness
@@ -47,17 +37,10 @@ public class DripstonePlacementRule extends BlockPlacementRule {
     }
 
     @Override
-<<<<<<< HEAD
-    public @NotNull Block blockUpdate(@NotNull UpdateState updateState) {
-        var direction = updateState.currentBlock().getProperty(PROP_VERTICAL_DIRECTION).equals("up");
-        var newThickness = getThickness(updateState.instance(), updateState.blockPosition(), direction);
-        return updateState.currentBlock().withProperty(PROP_THICKNESS, newThickness);
-=======
     public @NotNull Block blockUpdate(@NotNull BlockChange mutation) {
         var direction = mutation.block().getProperty(PROP_VERTICAL_DIRECTION).equals("up");
         var newThickness = getThickness(mutation.instance(), mutation.blockPosition(), direction);
         return mutation.block().withProperty(PROP_THICKNESS, newThickness);
->>>>>>> cc02c79fb (Cleanup)
     }
 
     private @NotNull String getThickness(@NotNull Block.Getter instance, @NotNull Point blockPosition, boolean direction) {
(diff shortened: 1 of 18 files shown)


src/main/java/net/minestom/server/instance/block/BlockMutation.java
```diff
@@ -0,0 +1,49 @@
...
+    @NotNull
+    Point blockPosition();
+
+    @NotNull
+    Block block();
+
+    @Nullable
+    BlockFace blockFace();
+
+    BlockMutation setBlock(@NotNull Block newBlock);
```
mworzala: missing annotation

src/main/java/net/minestom/server/instance/block/BlockMutation.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    record Player(
+        @NotNull Block.Getter instance, @NotNull Point blockPosition,
+        @NotNull Block block, @NotNull BlockFace blockFace,
+        @NotNull net.minestom.server.entity.Player player
+    ) implements BlockMutation {
+
+        public @NotNull net.minestom.server.entity.Player player() {
```
mworzala: we should entirely represent the placement here if including things like block face (ie cursor position, item being used, etc)

src/main/java/net/minestom/server/instance/block/BlockMutation.java
```diff
@@ -0,0 +1,49 @@
...
+    Block block();
+
+    @Nullable
+    BlockFace blockFace();
+
+    BlockMutation setBlock(@NotNull Block newBlock);
+
+    record Instance(
+        @NotNull Block.Getter instance, @NotNull Point blockPosition,
+        @NotNull Block block, @Nullable BlockFace blockFace
```
mworzala: blockFace is weird for a non-player placement

src/main/java/net/minestom/server/instance/block/BlockMutation.java
mworzala: to me setblock implies that we are adding a new change, but it actually replaces the change here. `withBlock` would be better

src/main/java/net/minestom/server/instance/block/BlockMutation.java
```diff
@@ -0,0 +1,49 @@
+package net.minestom.server.instance.block;
+
+import net.minestom.server.coordinate.Point;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public sealed interface BlockMutation {
```
mworzala: Would prefer `BlockChange` or something, `BlockMutation` makes me think its mutable

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -96,49 +102,47 @@ public void setBlock(int x, int y, int z, @NotNull Block block,
...
         // Handler
         final BlockHandler handler = block.handler();
-        final Block lastCachedBlock;
-        if (handler != null || block.hasNbt() || block.registry().isBlockEntity()) {
-            lastCachedBlock = this.entries.put(index, block);
-        } else {
-            lastCachedBlock = this.entries.remove(index);
-        }
+        final boolean shouldCache = handler != null || block.hasNbt() || block.registry().isBlockEntity();
+        final Block lastCachedBlock = shouldCache ? this.entries.get(index) : null;
```
mworzala: this should just be a remove always, we are going to remove the old block regardless of whether the current block needs to be stored.

src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -139,116 +138,81 @@ public void setBlock(int x, int y, int z, @NotNull Block block, boolean doBlockU
...
-        PlayerBlockBreakEvent blockBreakEvent = new PlayerBlockBreakEvent(player, block, Block.AIR, new BlockVec(blockPosition), blockFace);
+        PlayerBlockBreakEvent blockBreakEvent = new PlayerBlockBreakEvent(mutation.player(), block, Block.AIR, new BlockVec(mutation.blockPosition()), mutation.blockFace());
         EventDispatcher.call(blockBreakEvent);
         final boolean allowed = !blockBreakEvent.isCancelled();
         if (allowed) {
             // Break or change the broken block based on event result
-            final Block resultBlock = blockBreakEvent.getResultBlock();
-            UNSAFE_setBlock(chunk, x, y, z, resultBlock, null,
-                    new BlockHandler.PlayerDestroy(block, this, blockPosition, player), doBlockUpdates, 0);
+            mutation = (BlockMutation.Player) mutation.setBlock(blockBreakEvent.getResultBlock());
```
mworzala: specialize BlockMutation.Player#setBlock to return Player

src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -139,116 +138,81 @@ public void setBlock(int x, int y, int z, @NotNull Block block, boolean doBlockU
...
-                            this, block, null, blockPosition,
-                            null, null, null,
-                            false
-                    );
-                }
 
-                block = blockPlacementRule.blockPlace(rulePlacement);
-                if (block == null) block = Block.AIR;
+            final BlockPlacementRule placementRule = MinecraftServer.getBlockManager().getBlockPlacementRule(mutation.block());
+            if(placementRule != null && doBlockUpdates) {
```
mworzala: formatting

src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -139,116 +138,81 @@ public void setBlock(int x, int y, int z, @NotNull Block block, boolean doBlockU
...
+                mutation = placementRule.blockPlace(mutation);
```
mworzala: should do an explicit null check here

src/main/java/net/minestom/server/instance/block/BlockMutation.java
Garrett-EMS: Fixed as of 66d67c4

Garrett-EMS: Fixed as of 66d67c4

src/main/java/net/minestom/server/instance/block/BlockMutation.java
Garrett-EMS: Fixed as of 66d67c4

src/main/java/net/minestom/server/instance/InstanceContainer.java
Garrett-EMS: Fixed as of 66d67c4

src/main/java/net/minestom/server/instance/block/BlockMutation.java
Garrett-EMS: Fixed in 2814ca1

src/main/java/net/minestom/server/instance/block/BlockMutation.java
Garrett-EMS: Fixed in 840088c

src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
```diff
@@ -54,34 +46,12 @@ public int maxUpdateDistance() {
...
-            @NotNull Block.Getter instance,
-            @NotNull Block block,
-            @Nullable BlockFace blockFace,
-            @NotNull Point placePosition,
-            @Nullable Point cursorPosition,
-            @Nullable Pos playerPosition,
-            @Nullable ItemStack usedItemStack,
-            boolean isPlayerShifting
-    ) {
+    public boolean isClientPredicted() {
```
mworzala: Why is this here?

demo/src/main/java/net/minestom/demo/block/placement/DripstonePlacementRule.java
```diff
@@ -20,25 +19,28 @@ public DripstonePlacementRule() {
     }
 
     @Override
-    public @Nullable Block blockPlace(@NotNull PlacementState placementState) {
-        var blockFace = Objects.requireNonNullElse(placementState.blockFace(), BlockFace.TOP);
+    public @NotNull Block blockPlace(@NotNull BlockChange blockChange) {
```
mworzala, line 22: Why not make blockPlace take a player change? we only trigger on place anyway which comes from a player (or with zero-ed other values). seems like kinda annoying boilerplate added to every placement rule.

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -77,14 +78,19 @@ protected DynamicChunk(@NotNull Instance instance, int chunkX, int chunkZ, @NotN
     }
 
     @Override
-    public void setBlock(int x, int y, int z, @NotNull Block block,
-                         @Nullable BlockHandler.Placement placement,
-                         @Nullable BlockHandler.Destroy destroy) {
+    public @NotNull Block setBlock(@NotNull BlockChange mutation) {
```
mworzala, line 80: Seems kinda weird for only one setBlock call to return the block, what is the use case for this return value?

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -96,49 +102,49 @@ public void setBlock(int x, int y, int z, @NotNull Block block,
...
-        // Block tick
-        if (handler != null && handler.isTickable()) {
-            this.tickableMap.put(index, block);
-        } else {
-            this.tickableMap.remove(index);
+        final boolean shouldCache = handler != null || block.hasNbt() || block.registry().isBlockEntity();
+
+        final Block lastCachedBlock = this.entries.get(index);
+
+        this.entries.remove(index);
```
mworzala: remove returns the removed value or null if there wasn't one, can replace the get above.

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -96,49 +102,49 @@ public void setBlock(int x, int y, int z, @NotNull Block block,
...
-            lastCachedBlock = this.entries.put(index, block);
-        } else {
-            lastCachedBlock = this.entries.remove(index);
-        }
-        // Block tick
-        if (handler != null && handler.isTickable()) {
-            this.tickableMap.put(index, block);
-        } else {
-            this.tickableMap.remove(index);
+        final boolean shouldCache = handler != null || block.hasNbt() || block.registry().isBlockEntity();
```
mworzala: should just use prior logic here of replace or remove, eg

```
if (handler != null || block.hasNbt() || block.registry().isBlockEntity()) {
            lastCachedBlock = this.entries.put(index, block);
        } else {
            lastCachedBlock = this.entries.remove(index);
        }
```

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -96,49 +102,49 @@ public void setBlock(int x, int y, int z, @NotNull Block block,
...
+            this.entries.put(index, block);
         }
 
-        // Update block handlers
-        var blockPosition = new Vec(x, y, z);
         if (lastCachedBlock != null && lastCachedBlock.handler() != null) {
-            // Previous destroy
-            lastCachedBlock.handler().onDestroy(Objects.requireNonNullElseGet(destroy,
-                    () -> new BlockHandler.Destroy(lastCachedBlock, instance, blockPosition)));
+            block = lastCachedBlock.handler().onDestroy(new BlockChange.Instance(instance, mutation.blockPosition(), block));
```
mworzala: why do we annihilate the potential player change here? seems better to preserve whatever change we received in onDestroy and onPlace. Also im not totally convinced on onDestroy being able to overwrite a place entirely. what is the use case for this?

It is a bit tricky because you have already added the block above, but now it can be changed to a new block with a new handler which we now need to handle. I think the whole thing adds a lot of complexity and potential confusion for little benefit (and currently has logic issues).

This method and Instance#setBlock need a lot of added tests for this change

src/main/java/net/minestom/server/instance/DynamicChunk.java
Garrett-EMS: Fixed in 6429ebc

src/main/java/net/minestom/server/instance/DynamicChunk.java
Garrett-EMS: Fixed in 6429ebc

src/main/java/net/minestom/server/instance/block/rule/BlockPlacementRule.java
Garrett-EMS: Fixed in 6429ebc

demo/src/main/java/net/minestom/demo/block/placement/DripstonePlacementRule.java
Garrett-EMS: We do actually trigger for Instance#setBlock which doesnt have a player. I feel this should stay this way as it allows blockplacementrules to apply for blocks placed by the instance, which is better for world gen and similar stuff.

src/main/java/net/minestom/server/instance/DynamicChunk.java
Garrett-EMS: We need it because BlockChange is now not mutable, and we need to know the final block to send packets so that its not wrong if Chunk#setblock mutates the value since it no longer loops and sends another setblock to instance.

src/main/java/net/minestom/server/instance/DynamicChunk.java
Garrett-EMS: Thats mostly a mistake on my part, it should be a player change. as for the last place im not entirely sure what you mean but im assuming you dont think destroy should be able to mutate the final block? I disagree because it is useful for modifying the block when it is broken to custom behaviors. For instance if someone was to recreate hypixel skyblock (i know it's never been done...) you may want to replace the broken block with bedrock to recreate that behavior.

tests will be added soon

Garrett-EMS:
go to https://github.com/Minestom/Minestom/pull/2816

PR #2796, opened 2025-06-27
Validate input key format in DialogInput records

FrederikHeinrich:
This pull request introduces validation for input keys in various `DialogInput` record types within the `src/main/java/net/minestom/server/dialog/DialogInput.java` file. The validation ensures that input keys follow a specific pattern (`^[a-z0-9_]+) to maintain consistency and prevent invalid keys.

### Key Validation Enhancements:

* **`Boolean` record**: Added a constructor validation to check that the `key` matches the required pattern. Throws an exception if the condition is not met.
* **`NumberRange` record**: Similar validation added to ensure the `key` adheres to the pattern.
* **`SingleOption` record**: Introduced validation for the `key` in its constructor, ensuring compliance with the pattern.
* **`Text` record**: Added validation for the `key` to enforce the pattern requirement.Added constructor checks to ensure that the 'key' field in all DialogInput record types matches the pattern [a-zA-Z0-9_]+. This enforces consistent and valid input keys across dialog input implementations.

## Proposed changes

- Inline validation added to all DialogInput records to ensure keys are well-formed
- Avoided external utility class for simplicity and performance
- Improved robustness of dialog input structure construction
- Reduced potential for user-facing decoding errors due to malformed keys

## Types of changes

## Checklist

merging your code._

## Further comments

This change was made to ensure long-term consistency and validation safety across all implementations of DialogInput. Instead of repeating the regex condition in every record, the check is now reusable, easier to test, and future-proof.

Let me know if you'd prefer the validation logic to live somewhere else (e.g. DialogInput itself or in a base class), or if additional patterns should be supported.

### DIFF
--- a/src/main/java/net/minestom/server/dialog/DialogInput.java
+++ b/src/main/java/net/minestom/server/dialog/DialogInput.java
@@ -31,6 +31,7 @@ public sealed interface DialogInput {
             @NotNull String onTrue,
             @NotNull String onFalse
     ) implements DialogInput {
+
         public static final StructCodec<Boolean> CODEC = StructCodec.struct(
                 "key", Codec.STRING, Boolean::key,
                 "label", Codec.COMPONENT, Boolean::label,
@@ -38,6 +39,10 @@ public sealed interface DialogInput {
                 "on_true", StructCodec.STRING.optional("true"), Boolean::onTrue,
                 "on_false", StructCodec.STRING.optional("false"), Boolean::onFalse,
                 Boolean::new);
+
+        public Boolean {
+            Check.argCondition(!key.matches("^[a-z0-9_]+$"), "Invalid input key: " + key + ". Must match [a-zA-Z0-9_]+");
+        }
 
         @Override
         public @NotNull StructCodec<? extends DialogInput> codec() {
@@ -53,6 +58,7 @@ public sealed interface DialogInput {
             @Nullable Float initial,
             @Nullable Float step
     ) implements DialogInput {
+
         public static final StructCodec<NumberRange> CODEC = StructCodec.struct(
                 "key", Codec.STRING, NumberRange::key,
                 "width", Codec.INT.optional(DEFAULT_WIDTH), NumberRange::width,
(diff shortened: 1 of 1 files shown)


pernydev:
Can I ask how this helps? I fail to see the purpose here. Runtime exceptions feel more of an annoyance than a safety guard.

FrederikHeinrich:
An error in the console is better than a client that loses the connection because you send a wrong key.

pernydev:
And where did you get that pattern from?

FrederikHeinrich:
https://minecraft.wiki/w/Dialog#text
grafik

mworzala (commented):
This is perhaps a minor difference but if we are going to validate we should get it right.

Vanilla uses `Character.isLetterOrDigit(theChar) || theChar == '_'` to decide if a character is valid in one of these keys, we should use the same.

src/main/java/net/minestom/server/dialog/DialogInput.java
```diff
@@ -24,6 +24,13 @@ public sealed interface DialogInput {
             Map.entry(Key.key("text"), Text.CODEC));
     @NotNull StructCodec<DialogInput> CODEC = Codec.RegistryTaggedUnion(REGISTRY, DialogInput::codec, "type");
 
+    static boolean keyIsValidate(String key) {
```
mworzala: grammar error, but should be `validateKey(key)` and throw inside so we dont have to duplicate the message repeatedly.

src/main/java/net/minestom/server/dialog/DialogInput.java
```diff
@@ -41,7 +48,7 @@ record Boolean(
                 Boolean::new);
 
         public Boolean {
-            Check.argCondition(!key.matches("^[a-z0-9_]+$"), "Invalid input key: " + key + ". Must match [a-zA-Z0-9_]+");
+            Check.argCondition(keyIsValidate(key), "Invalid input key: " + key + ". Must match [a-zA-Z0-9_]+");
```
kermandev: Check.argCondition will throw if the result of keyIsValidate/validateKey is true, this seems not what you want.

Also Check provides formatting so you could replace `+ key +` with {0} and pass the key argument.

PR #2799, opened 2025-06-28
Add PlayerBrandEvent

mudkipdev:
Makes it easier to get the client brand. Listening to the `PlayerPluginMessageEvent` and using `getMessageString()` can be misleading as this does not account for the length prefix of the string.

Alternative could be to create some kind of extensible `CustomPayload` registry to make serializing your own plugin messages easier, inspired by [Minecraft's source code](https://github.com/NikitaCartes-archive/MinecraftDeobfuscated-Mojang/tree/snapshot/minecraft/src/net/minecraft/network/protocol/common/custom) (which seems to have many plugin messages that are debug only and not sent).

### DIFF
--- a/src/main/java/net/minestom/server/listener/common/PluginMessageListener.java
+++ b/src/main/java/net/minestom/server/listener/common/PluginMessageListener.java
@@ -10,13 +10,13 @@ import net.minestom.server.network.packet.client.common.ClientPluginMessagePacke
 
 public class PluginMessageListener {
     private static final String BRAND_CHANNEL = "minecraft:brand";
-    private static final ListenerHandle<PlayerBrandEvent> brandHandle = EventDispatcher.getHandle(PlayerBrandEvent.class);
+    private static final ListenerHandle<PlayerBrandEvent> BRAND_HANDLE = EventDispatcher.getHandle(PlayerBrandEvent.class);
 
     public static void listener(ClientPluginMessagePacket packet, Player player) {
         PlayerPluginMessageEvent pluginMessageEvent = new PlayerPluginMessageEvent(player, packet.channel(), packet.data());
         EventDispatcher.call(pluginMessageEvent);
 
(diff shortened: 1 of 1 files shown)


kermandev:
Please use a handle (See PlayerPacketOutEvent implementation) to check if there are any listeners before reading.

Also a integration test would also be nice.

src/main/java/net/minestom/server/event/player/PlayerPluginMessageEvent.java
```diff
@@ -46,6 +46,7 @@ public byte[] getMessage() {
      * @return the message
      */
     @NotNull
+    @Deprecated(forRemoval = true)
```
mworzala, line 49: Why is this deprecated?

mworzala (changes requested):
would also like to see a test

src/main/java/net/minestom/server/event/player/PlayerPluginMessageEvent.java
mudkipdev: Because it simply creates a string from the sent bytes, it doesnt take into account strings prefixed with lengths. Since the only plugin messages in minecraft dont work like this i dont see any reason to keep it.

mudkipdev:
I am not sure how to structure a test for this

kermandev:
Add a listener, see if the event is called, and ensure the buffer has the contents your expecting.

src/main/java/net/minestom/server/event/player/PlayerPluginMessageEvent.java
mworzala: The brand plugin messages happen to have a length prefix but some other mod or proxy plugin may not choose to structure it like that. It's just arbitrary bytes to me having a string method is reasonable

mudkipdev: Ok, I will keep it and simply add a note in the javadoc

PR #2802, opened 2025-06-29
perf: registry key speed improvements

Bloeckchengrafik:
## Proposed changes

Speed up DynamicRegistry#getId() dramatically

```
DynamicRegistryGetIdBenchmark.getId:before  avgt   30  679.928 ± 10.940  ns/op
DynamicRegistryGetIdBenchmark.getId:map     avgt   30   21.094 ±  0.209  ns/op
```

This is specifically done to enhance biome setting in generators. It's now about the same speed as setAllRelative when executed for every fourth block (best possible way to execute it)

## Types of changes

- [x] Performance boost

## Checklist

merging your code._

- [x] I have added benchmarks that prove my fix is effective or that my feature works

## Further comments

I have no idea why, but for tests to pass, the dimension type registry can't be static. If you know why, please let me know

### DIFF
--- a/jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
+++ b/jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
@@ -1,8 +1,8 @@
 package net.minestom.server.registry;
 
-import net.minestom.server.MinecraftServer;
 import net.minestom.server.world.biome.Biome;
 import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
 
 import java.util.concurrent.TimeUnit;
 
@@ -14,7 +14,6 @@ import java.util.concurrent.TimeUnit;
 @State(Scope.Benchmark)
 public class DynamicRegistryGetIdBenchmark {
     private DynamicRegistry<Biome> registry;
-    private final RegistryKey<Biome> biome = Biome.PALE_GARDEN;
 
     @Setup
     public void setup() {
@@ -22,7 +21,7 @@ public class DynamicRegistryGetIdBenchmark {
     }
 
     @Benchmark
-    public int getId() {
-        return registry.getId(biome);
+    public void getId(Blackhole blackhole) {
+        blackhole.consume(registry.getId(RegistryKey.unsafeOf("pale_garden")));
     }
 }
--- a/jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
+++ b/jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
@@ -1,0 +1,28 @@
+package net.minestom.server.registry;
+
+import net.minestom.server.MinecraftServer;
+import net.minestom.server.world.biome.Biome;
+import org.openjdk.jmh.annotations.*;
+
+import java.util.concurrent.TimeUnit;
+
+@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 10, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Fork(3)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+public class DynamicRegistryGetIdBenchmark {
+    private DynamicRegistry<Biome> registry;
+    private final RegistryKey<Biome> biome = Biome.PALE_GARDEN;
+
+    @Setup
+    public void setup() {
+        registry = Biome.createDefaultRegistry();
+    }
+
+    @Benchmark
+    public int getId() {
+        return registry.getId(biome);
+    }
+}
--- a/src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
+++ b/src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
@@ -37,6 +37,7 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
 
     private final List<T> idToValue;
     private final List<RegistryKey<T>> idToKey;
+    private final Map<RegistryKey<T>, Integer> keyToId;
     private final Map<Key, T> keyToValue;
     private final Map<T, RegistryKey<T>> valueToKey;
     private final List<DataPack> packById;
@@ -52,6 +53,7 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
         // Expect stale data possibilities with unsafe ops.
         this.idToValue = new ArrayList<>();
         this.idToKey = new ArrayList<>();
+        this.keyToId = new HashMap<>();
         this.keyToValue = new HashMap<>();
         this.valueToKey = new HashMap<>();
         this.packById = new ArrayList<>();
@@ -68,10 +70,15 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
         this.codec = codec;
         this.idToValue = idToValue;
         this.idToKey = idToKey;
+        this.keyToId = new HashMap<>();
         this.keyToValue = keyToValue;
         this.valueToKey = valueToKey;
         this.packById = packById;
         this.tags = tags;
+
+        for (int id = 0; id < idToKey.size(); id++) {
+            keyToId.put(idToKey.get(id), id);
+        }
     }
 
     @Override
@@ -116,7 +123,7 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
 
     @Override
     public int getId(@NotNull RegistryKey<T> key) {
-        return idToKey.indexOf(key);
+        return keyToId.get(key);
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.concurrent.TimeUnit;
+
+@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 10, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
+@Fork(3)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Benchmark)
+public class DynamicRegistryGetIdBenchmark {
```
kermandev, line 15: This test is invalid because your storing state between tests. Please construct a new RegistryKey#unsafeOf in setup

jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
```diff
@@ -0,0 +1,28 @@
...
+    private DynamicRegistry<Biome> registry;
+    private final RegistryKey<Biome> biome = Biome.PALE_GARDEN;
+
+    @Setup
+    public void setup() {
+        registry = Biome.createDefaultRegistry();
+    }
+
+    @Benchmark
+    public int getId() {
```
kermandev: Please use a blackhole here instead of returning

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -68,10 +73,15 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
         this.codec = codec;
         this.idToValue = idToValue;
         this.idToKey = idToKey;
+        this.keyToId = new HashMap<>();
         this.keyToValue = keyToValue;
         this.valueToKey = valueToKey;
         this.packById = packById;
         this.tags = tags;
+
+        for (int id = 0; id < idToKey.size(); id++) {
```
kermandev: See above

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -146,15 +166,18 @@ public int getId(@NotNull RegistryKey<T> key) {
     @Override
     public boolean remove(@NotNull Key key) throws UnsupportedOperationException {
         if (isFrozen()) throw new UnsupportedOperationException(UNSAFE_REMOVE_MESSAGE);
+        version++;
```
kermandev: Move into synchronized block

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -116,25 +126,35 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
...
+            registryKey.upgrade(version, id);
+            return id;
+        }
+        return keyToId.get(key);
     }
 
     @Override
     public @NotNull RegistryKey<T> register(@NotNull Key key, @NotNull T object, @Nullable DataPack pack) {
         if (isFrozen()) throw new UnsupportedOperationException(UNSAFE_REMOVE_MESSAGE);
+        version++;
```
kermandev: synchronized block please

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -116,7 +116,7 @@
  * You can easily create your own implementation of this and use it with {@link ConnectionManager#setPlayerProvider(PlayerProvider)}.
  */
 public class Player extends LivingEntity implements CommandSender, HoverEventSource<ShowEntity>, Identified, NamedAndIdentified {
-    private static final DynamicRegistry<DimensionType> DIMENSION_TYPE_REGISTRY = MinecraftServer.getDimensionTypeRegistry();
+    private final DynamicRegistry<DimensionType> DIMENSION_TYPE_REGISTRY = MinecraftServer.getDimensionTypeRegistry();
```
kermandev: This change doesnt make sense; Regardless these definitions should be removed but thats another thing.

Likely issues stems from you sharing versioning across tests.

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -68,10 +73,15 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
         this.codec = codec;
         this.idToValue = idToValue;
         this.idToKey = idToKey;
+        this.keyToId = new HashMap<>();
```
kermandev: Please just make this pass the existing keyToId instead, see compact

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -46,12 +47,16 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
     private final Key key;
     private final Codec<T> codec;
 
+    // Effectively automatic cache invalidation on the RegistryKeyImpl
+    private int version = 0;
```
kermandev: versioning has the opportunity to leak between tests

kermandev: Make volatile too

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
```diff
@@ -116,25 +126,35 @@ final class DynamicRegistryImpl<T> implements DynamicRegistry<T> {
 
     @Override
     public int getId(@NotNull RegistryKey<T> key) {
-        return idToKey.indexOf(key);
+        if (key instanceof RegistryKeyImpl<T> registryKey) {
+            int id = registryKey.getLastRegistryId(version);
```
kermandev: need to pull out version into a final ex: `final int version = this.version`

kermandev (changes requested):
I dont think RegistryKeyImpl should be stateful at all, as they are hard to keep clean in tests. There is a potential performance improvement though; so there might be something done with a backed implementation.

I am fine with storing keys to id's though; as you are right, it is a hotpath.

Bloeckchengrafik:
The performance boost of stateful registry keys is a lot. See the benchmarks above:
The first line is the baseline on the current 1.21.6, and the second is after the first commit, so without stateful RegistryKeys. The third one is the full set of optimizations. 

I'm fine with dropping the second commit as this is something that can be implemented user-side as well.

I'll address the other review comments when I'm back at my PC in a couple of hours.

Thanks for the thorough review!

DasBabyPixel:
From a code clarity/safety point of view I am also very much against stateful registry keys. It is obviously blazingly fast, because instead of accessing a second data structure you are only executing a single getfield instruction. I'd love to know however what the implications are, can there ever be multiple dynamic registries sharing the same registry key? I haven't looked too much into this yet. 
If it is safe to have mutable registry keys and there is no better/equal alternative that can achieve similar performance, I am for mutable registry keys. However, the reason for the mutability should be added with comments in the code.

Bloeckchengrafik:
I'd say proving that it is safe is probably not particularily doable. I'll remove the second commit since it can be implemented user-side with about the same performance. It's a big performance increase like this as well.

jmh-benchmarks/src/jmh/java/net/minestom/server/registry/DynamicRegistryGetIdBenchmark.java
Bloeckchengrafik: Done

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
Bloeckchengrafik: not applicable anymore

src/main/java/net/minestom/server/entity/Player.java
Bloeckchengrafik: Still needed for tests to pass. What do you mean by removing the definition? Should I just inline it?

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
Bloeckchengrafik: not applicable anymore

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
Bloeckchengrafik: not applicable anymore

src/main/java/net/minestom/server/registry/DynamicRegistryImpl.java
Bloeckchengrafik: not applicable anymore

src/main/java/net/minestom/server/entity/Player.java
kermandev: Can you see if its still required if you have rollbacked your changes

Bloeckchengrafik: before my changes, it works fine, after, it does not

kermandev: Your not handling DynamicRegistery#getId for null values that is why; Please return -1 if the id value is null; then remove this change.

Bloeckchengrafik: 🤦🏼

Bloeckchengrafik:
For future reference: https://github.com/PolyhedralDev/Terra/pull/513/files#diff-c6364c4370a494254755a25680941d709b40a43cc0b30c65f64750127e42e151R64

This is the way caching behaviour can be implemented user-side. A better API for this would be great but out-of-scope for this PR

DasBabyPixel:
Maybe it is possible to add an implementation for the registry key that does caching, but it is up to the user to pass an instance of that implementation. I am not very happy with the concept of that, a fully out-of-the-box working solution would be much nicer, but it should be better than the user having to create his own implementation for the sake of performance

DasBabyPixel:
What are the benchmarks with/without the caching, keeping the remaining changes?

Bloeckchengrafik:
We're at 20ns vs ~1.3ns for the two. By implementing it the way I did above, it's even faster. I can't even see it on my flame graph. 

I think two registry key impls are not a great way to go although I'd love to get the opinion of a maintainer on the topic.

mworzala (approved):
The id by key lookup makes a lot of sense to me, ill merge that now.

On caching values in a registry key, it has a lot of potential bad implication. There is not necessarily a single instance of a DynamicRegistry for each type (there is in Minestom by default, but that may not be the case long term or with user code), and accidentally getting the wrong value because you used the same key in two registries would be extremely confusing.

I do think its reasonable to cache them in your own isolated logic (where you know the keys are being used by one registry, and that the registry isnt changing during that runtime) but im not very convinced on Minestom doing it by default.

Bloeckchengrafik:
works for me

PR #2811, opened 2025-07-03
Add version to default brand name

TheMode:


GoldenStack:
I feel like the purpose of the brand name area is for the brand name. Like I don't know if it really makes sense to also have the version here.

PR #2815, opened 2025-07-03
Add InstanceSectionInvalidateEvent

TheMode:
Let users manually mess with palette and let the instance be responsible for ensuring players get up-to-date packets.

The event also prevent keeping outdated state.

### DIFF
--- a/src/main/java/net/minestom/server/event/instance/InstanceSectionInvalidateEvent.java
+++ b/src/main/java/net/minestom/server/event/instance/InstanceSectionInvalidateEvent.java
@@ -1,0 +1,39 @@
+package net.minestom.server.event.instance;
+
+import net.minestom.server.event.trait.InstanceEvent;
+import net.minestom.server.instance.Instance;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * This event is triggered when a section of an instance is manually marked as invalid.
+ * <p>
+ * Changes in this case are not known but indicate that its content must be reinterpreted.
+ */
+public class InstanceSectionInvalidateEvent implements InstanceEvent {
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -279,6 +280,14 @@ public void unloadChunk(int chunkX, int chunkZ) {
         unloadChunk(chunk);
     }
 
+    public void invalidateSection(int sectionX, int sectionY, int sectionZ) {
+        final Chunk chunk = getChunk(sectionX, sectionZ);
+        if (chunk != null) {
```
mworzala, line 285: we need to reevaluate the heightmap (needsCompleteHeightmapRefresh in DynamicChunk) in DynamicChunk#invalidate also (probably a mistake that its not there already, but exposed by this pr)

src/main/java/net/minestom/server/event/instance/InstanceSectionInvalidateEvent.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minestom.server.event.trait.InstanceEvent;
+import net.minestom.server.instance.Instance;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * This event is triggered when a section of an instance is manually marked as invalid.
+ * <p>
+ * Changes in this case are not known but indicate that its content must be reinterpreted.
+ */
+public class InstanceSectionInvalidateEvent implements InstanceEvent {
```
mworzala, line 15: Should mention that this should be triggered through Instance#invalidateSection rather than called manually (and maybe mark constructor internal additionally)

PR #2818, opened 2025-07-04
Fix PaintingVariant ordering

kermandev:
## Proposed changes

The current ordering is wrong, the assetId should be after the width and height.

## Types of changes

## Checklist

merging your code._

## Further comments
Codec change not required since its backed by map, but might as well have the same correct ordering.
Chose not to change the signature of create as its better if the key was prefixed, but this could be deprecated and changed later.


kermandev:
Do not merge, metadata seems to be wrong will look at later.

mworzala (approved):

PR #2819, opened 2025-07-08
Palette#count(value) and Palette#any(value)

TheMode:
Further optimizable, but prefer getting them in this state first

### DIFF
--- a/demo/src/main/java/net/minestom/demo/Main.java
+++ b/demo/src/main/java/net/minestom/demo/Main.java
@@ -31,6 +31,7 @@ import java.util.List;
 public class Main {
 
     public static void main(String[] args) {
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -99,6 +99,16 @@ static Palette sized(int dimension, int minBitsPerEntry, int maxBitsPerEntry, in
...
 
+    /**
+     * Returns the number of entries in this palette that match the given value.
+     *
+     * @param value the value to count
+     * @return the number of entries matching the value
+     */
+    int count(int value);
+
+    boolean any(int value);
```
mworzala, line 116: add javadoc

PR #2821, opened 2025-07-08
Slight event cleanup

TheMode:
Dont reinvent `Function`
Avoid anonymous listener impl
Recursive test seems to somehow pass now

### DIFF
--- a/src/main/java/net/minestom/server/event/EventListener.java
+++ b/src/main/java/net/minestom/server/event/EventListener.java
@@ -19,9 +19,11 @@ import java.util.function.Predicate;
  *
  * @param <T> The event type being handled.
  */
-public interface EventListener<T extends Event> extends Function<T, EventListener.Result> {
+public interface EventListener<T extends Event> {
 
     @NotNull Class<T> eventType();
+
+    @NotNull Result run(@NotNull T event);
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/event/EventListener.java
```diff
@@ -17,12 +19,10 @@
  *
  * @param <T> The event type being handled.
  */
```
kermandev: Could you add FunctionalInterface annotation here

mudkipdev:
Why change name from `run` to `apply`? This will break a lot of code for no reason

src/main/java/net/minestom/server/event/EventListener.java
TheMode: It has more than 1 abstract function so it doesnt fit

TheMode:
What code will it break? Its not a function you are supposed to call yourself

src/main/java/net/minestom/server/event/EventListener.java
kermandev: Oh, i see.

mudkipdev:
Implementing EventListener on a class

mworzala (approved):

PR #2825, opened 2025-07-11
Avoid supplier in StructCodec/NetworkTemplate

TheMode:
Kept the supplier version as it could remain useful in some very niche situations (not in this repo I suppose) and to avoid a breaking change

### DIFF
--- a/src/main/java/net/minestom/server/codec/StructCodec.java
+++ b/src/main/java/net/minestom/server/codec/StructCodec.java
@@ -55,6 +55,21 @@ public interface StructCodec<R> extends Codec<R> {
         };
     }
 
+    static <R> StructCodec<R> struct(R value) {
+        final Result<R> ok = new Result.Ok<>(value);
+        return new StructCodec<>() {
+            @Override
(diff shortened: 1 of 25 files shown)


src/main/java/net/minestom/server/codec/StructCodec.java
```diff
@@ -55,6 +55,21 @@ default StructCodec<R> orElseStruct(@NotNull StructCodec<R> other) {
         };
     }
 
+    static <R> StructCodec<R> struct(R value) {
+        final Result<R> ok = new Result.Ok<>(value);
```
kermandev, line 59: I don't think we should wrap the value; as Result objects aren't really meant to be long lived.

src/main/java/net/minestom/server/item/component/ConsumeEffect.java
```diff
@@ -57,7 +57,7 @@ final class ClearAllEffects implements ConsumeEffect {
 
         public static final NetworkBuffer.Type<ClearAllEffects> NETWORK_TYPE = NetworkBuffer.UNIT
```
kermandev: Could you change this to the new template. (There is also 2 more usages in SlotDisplay aswell)

Also Codec with DamageImmunityEffect

PR #2826, opened 2025-07-11
Split block occlusion and collision shapes

TogAr2:
## Proposed changes

This PR separates the occlusion and collision shapes of blocks. Previously, those were linked together in the `ShapeImpl` object.

For example, `ShapeImpl#isOccluded` would perform an occlusion check using the _occlusion shape_, but `ShapeImpl#relativeStart` (on the same shape) would return the relative start of the _collision shape_. Besides being very confusing, this made it impossible to perform occlusion checks on the _collision shape_ or get the relative start of the _occlusion shape_.
Especially the ability to do occlusion checks on the collision shape is very useful for reimplementing certain vanilla features (my current use case is fluids).

With this PR it is possible to do the following:
`block.registry().occlusionShape().relativeStart()`
`block.registry().collisionShape().isOccluded(...)`

## Types of changes

This is a breaking change, since existing uses of `shape.collisionShape().isOccluded()` will now no longer check occlusion for the _occlusion shape_, but for the _collision shape_.

## Checklist

## Further comments

This does slightly increase the memory usage of block shapes, since the `relativeStart`, `relativeEnd` and `lightEmission` fields will now be tracked for both shape types. However, with the added functionality I think it is worth it.

Also: previously, `ShapeImpl` had a `CollisionData` field and a `LightData` field, which were independent of each other. With this PR, it has a `ShapeData` and an `OcclusionData` field, but these are no longer independent, since the `OcclusionData` is made using the `ShapeData`. Would this matter? In my opinion it is cleaner this way, but we could also just add all the data fields to `ShapeImpl` directly.

### DIFF
--- a/src/main/java/net/minestom/server/collision/CollisionUtils.java
+++ b/src/main/java/net/minestom/server/collision/CollisionUtils.java
@@ -177,12 +177,22 @@ public final class CollisionUtils {
         return newPosition;
     }
 
-    public static Shape parseBlockShape(Map<Object, Object> internCache, String collision, String occlusion, boolean occludes, byte lightEmission) {
-        record ShapeEntry(String collision, String occlusion, boolean occludes, byte lightEmission) {} // Easy way to Hashcode
-        ShapeEntry entry = new ShapeEntry(collision, occlusion, occludes, lightEmission);
+    public static Shape parseCollisionShape(Map<Object, Object> internCache, String shape) {
+        record ShapeEntry(String shape) {} // Easy way to Hashcode
+        ShapeEntry entry = new ShapeEntry(shape);
         final Shape cachedShape = (Shape) internCache.get(entry);
         if (cachedShape != null) return cachedShape;
-        final Shape parsedShape = ShapeImpl.parseBlockFromRegistry(collision, occlusion, occludes, lightEmission);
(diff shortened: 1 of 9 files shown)


src/main/java/net/minestom/server/collision/CollisionUtils.java
```diff
@@ -177,12 +177,22 @@ public static Entity canPlaceBlockAt(Instance instance, Point blockPos, Block b)
         return newPosition;
     }
 
-    public static Shape parseBlockShape(Map<Object, Object> internCache, String collision, String occlusion, boolean occludes, byte lightEmission) {
-        record ShapeEntry(String collision, String occlusion, boolean occludes, byte lightEmission) {} // Easy way to Hashcode
-        ShapeEntry entry = new ShapeEntry(collision, occlusion, occludes, lightEmission);
+    public static Shape parseCollisionShape(Map<Object, Object> internCache, String shape) {
+        record ShapeEntry(String shape) {} // Easy way to Hashcode
```
mworzala: you can also just hash the string directly, no?

mworzala (commented):
Looks reasonable to me, but it needs to target the `next` branch because it is breaking.

PR #2829, opened 2025-07-11
Send InventoryCloseEvent on player inventory close

GoldenStack:
## Proposed changes

Fixes the event not previously being sent due to it relying on having a non-player inventory open.

## Types of changes

## Checklist

merging your code._


GoldenStack:
This should fix that (and I added a test) but there's a potential race condition if the open inventory is modified between the listener and the check inside the function. I forgot if race conditions are something we're typically concerned about here though, so LMK if I should try another solution (that's uglier IMO) instead of this.

mworzala (commented):
im not in love with the implicit handling in Player, why not pass the window id being closed into the internal close call and check there? So there are no possible mistakes

GoldenStack:
Ya, I was thinking passing it around would end up being pretty ugly, but ig it didn't turn out that bad
Should be good to go now

mworzala (approved):

PR #2834, opened 2025-07-13
BlockVec canonical constructor should take ints

TheMode:
Current implementation forces 3 flooring calls and require casting to retrieve block coordinates. Overhead should be moved to the case where doubles are explicitly provided.

This should be non-breaking in regard to `Point` but may still affect hash code and perhaps reflection.

I doubt it, but probably require review to ensure I didn't fail my Replace All.

edit: ended up deprecating static converters as well in favor of default Point methods

### DIFF
--- a/src/main/java/net/minestom/server/coordinate/Point.java
+++ b/src/main/java/net/minestom/server/coordinate/Point.java
@@ -297,4 +297,31 @@ public sealed interface Point permits Vec, Pos, BlockVec {
     default boolean sameBlock(@NotNull Point point) {
         return sameBlock(point.blockX(), point.blockY(), point.blockZ());
     }
+
+    @Contract(pure = true)
+    default @NotNull Pos asPosition() {
+        return switch (this) {
+            case Pos pos -> pos;
+            case Vec vec -> new Pos(vec.x(), vec.y(), vec.z());
+            case BlockVec blockVec -> new Pos(blockVec.blockX(), blockVec.blockY(), blockVec.blockZ());
+        };
+    }
+
+    @Contract(pure = true)
+    default @NotNull Vec asVec() {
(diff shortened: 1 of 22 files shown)


src/main/java/net/minestom/server/coordinate/Point.java
```diff
@@ -295,4 +297,31 @@ default boolean sameBlock(int blockX, int blockY, int blockZ) {
     default boolean sameBlock(@NotNull Point point) {
         return sameBlock(point.blockX(), point.blockY(), point.blockZ());
     }
+
+    @Contract(pure = true)
+    default @NotNull Pos asPosition() {
```
kermandev: If were already deprecating old usage of fromPoint; why not also deprecate this name and move to asPos() to line up with the other ones?

TheMode: I dont mind it just felt a bit duplicate for such a small change, can wait for @mworzala opinion

mworzala: Would prefer to rename yeah

src/main/java/net/minestom/server/coordinate/Pos.java
```diff
@@ -39,6 +39,7 @@ public Pos(@NotNull Point point) {
      *
      * @param point the point to convert
      * @return the converted position
+     * @deprecated use {@link Point#asPosition()} instead
      */
     public static @NotNull Pos fromPoint(@NotNull Point point) {
```
mworzala, line 45: add deprecated annotation also, same with other deprecations

mworzala (approved):
looks good after rename

mworzala (approved):

PR #2837, opened 2025-07-15
Declare JPMS & use NullMarked

kermandev:
## Proposed changes

Add JPMS to Minestom for NullMarked by JSpecify, so we assume anything unmarked is implicitly NotNull. This change should not affect any logic, and should be mostly documentation. This does show things that have failed to be marked like Builder's implementations. Unsure if we should address these now or in a future PR, as a whole project chore, as logically we should be checking inputs.

## Types of changes

Possible breaking change for people who use an unnamed JPMS.


mworzala:
I guess im not really opposed to using jspecify if im in the minority here, but I really just dont see the point at this stage. We use a lot of other JB annotations and I feel it will add a bunch of confusion if I ask someone to add nullability markers and they go use the wrong package/reviewers have to check imports/annotations closely every time.

I think adding notnullbydefault to package infos is reasonable (I know a bit more writing but not actually that bad). If jspecify ends up with more annotations in the future im sure my take will change, but curious other peoples.

kermandev:
Ive gone ahead and rebased it to master (hopefully this wont need to be rebased again)

Issue #2804, opened 2025-06-30
[Bug]: Saving tag data for chunks


Kingsthere:
### Affected Version

I'm using minestom ebaa2bb

### Describe the bug

Hi, I am trying to use AnvilLoader and Chunk.setTag to store data in chunks permanently, turns out the tag data is stored only in memory and is gone when restart (with AnvilLoader.saveInstance and AnvilLoader.saveChunksToStorage invoked)

### Code sample

```java
[Paste your code here]
```

SorkoPiko:
is this still present after #2840?

PR #2840, opened 2025-07-16
Anvil loader optimization

TheMode:
Tests going from 12sec to 4 (or 8 to 3 without the chunk lock)

Didn't do any additional check so perhaps there should be more tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,10 @@ gradle-app.setting
 .idea/
 **/out/
 
+# VSCode files
+.vscode/
+**/bin/
+
 # # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
 # gradle/wrapper/gradle-wrapper.properties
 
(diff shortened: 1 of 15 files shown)


TheMode:
May also solve #2804

mworzala (commented):
I dont have any obvious complaints from reading, would feel better with some more tests tho.

Also I have two not-really-related things that I would like to sneak in here because I am reminded of them:
* BlockLight and SkyLight both copy the light array no matter what so the sanity check in Section#setBlockLight of the entirely full or empty arrays is useless and we create a ton of light arrays. The copy is unnecessary besides being defensive about mutation, I think its internal enough that we dont need to care, but we could also copy if the array != full/empty.
* We should make RegionFile a public api, it is necessary for reading entity chunks which many people do and shouldn't have to reinvent region file reading

mworzala (approved):

PR #2843, opened 2025-07-17
fix two NPE's when for some reason palette values are null

p1k0chu:
## Proposed changes

not sure if its the best way to do it but sometimes the Palette's `values` array and also `paletteToValueList` are null,
which causes NPE
this fixes the exceptions by just checking if its null
`PaletteImpl#get` returns air if values are null now
`PaletteImpl#paletteIndexToValue` checks for `hasPalette()` AND if the array isn't null together now

## Types of changes

## Checklist

## Further comments


mworzala:
Do you have repro code for triggering an NPE here? I am unsure default of air is correct/how we end up in the problematic scenario and would like to look some more.

PR #2846, opened 2025-07-17
Codegen cleanup

TheMode:
Allow shortcuts like `ItemStack.DIAMOND` instead of `ItemStack.of(Material.DIAMOND)`.

Made some additional codegen cleanup so I could remove the ItemStack change if deemed unnecessary.


mworzala:
Consensus seems against the item stack constants, would like to keep the codegen cleanup tho for sure

PR #2852, opened 2025-07-20
Fix block break speed being incorrect for instant break blocks and zero mining speed

### LINKED ISSUE #2851: [Bug]: BlockBreakCalculation.breakTicks(Block, Player) incorrect when `BLOCK_BREAK_SPEED` attribute is 0 and block hardness is 0
### Affected Version

2025.07.11-1.21.7

### Describe the bug

If the player's `BLOCK_BREAK_SPEED` is 0 and you try to break a block that you would normally instamine (such as scaffolding), the block still breaks instantly instead of being unbreakable.

1. Give the player a base attribute value of 0 for `BLOCK_BREAK_SPEED`
2. Break an instamine block such as a scaffolding block

### Code sample

```java
player.getInventory().addItemStack(ItemStack.of(Material.SCAFFOLDING));
player.getAttribute(Attribute.BLOCK_BREAK_SPEED).setBaseValue(0);
```

Fisher2911:
## Proposed changes

This fixes the issue with breaking a block that has zero hardness still instamines it even if the player has zero mining speed.
This fixes #2851.

## Types of changes

## Checklist

merging your code._

## Further comments
I'm not sure if there is a better way to do this, but this should work. It just adds a check for block hardness being zero and the player's mining speed being zero.

### DIFF
--- a/src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
+++ b/src/main/java/net/minestom/server/utils/block/BlockBreakCalculation.java
@@ -89,8 +89,6 @@ public class BlockBreakCalculation {
             speedMultiplier /= 5;
         }
 
-        double damage = speedMultiplier / blockHardness;
-
         // if speed multiplier is 0, the block is unbreakable
         if (speedMultiplier == 0) {
             return UNBREAKABLE;
@@ -101,6 +99,7 @@ public class BlockBreakCalculation {
             return 0;
         }
 
+        double damage = speedMultiplier / blockHardness;
         if (isBestTool) {
             damage /= 30;
         } else {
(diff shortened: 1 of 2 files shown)


DasBabyPixel (approved):
I'd consider moving the checks above the calculation of `double damage`, but other than that looking good

mworzala (approved):

mworzala:
Actually, would also like to see a test for this - the prior behavior was wrong I believe (the division by zero check was after it would divide by zero :))

Fisher2911:
> Actually, would also like to see a test for this - the prior behavior was wrong I believe (the division by zero check was after it would divide by zero :))

You're right, I accidentally put the division by zero check after the division, but it shouldn't change the behavior because it's in a floating point context so it doesn't throw a divide by zero error.

PR #2855, opened 2025-07-21
add AsyncEvent trait

TheMode:
Basic check to ensure async events are called within virtual threads.

### DIFF
--- a/src/main/java/net/minestom/server/event/EventNodeImpl.java
+++ b/src/main/java/net/minestom/server/event/EventNodeImpl.java
@@ -2,6 +2,7 @@ package net.minestom.server.event;
 
 import net.minestom.server.MinecraftServer;
 import net.minestom.server.ServerFlag;
+import net.minestom.server.event.trait.AsyncEvent;
 import net.minestom.server.event.trait.RecursiveEvent;
 import net.minestom.server.utils.validate.Check;
 import org.jetbrains.annotations.Contract;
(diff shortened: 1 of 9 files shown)


src/main/java/net/minestom/server/event/EventNodeImpl.java
```diff
@@ -343,6 +344,8 @@ final class Handle<E extends Event> implements ListenerHandle<E> {
 
         @Override
         public void call(@NotNull E event) {
+            assert !(event instanceof AsyncEvent) || Thread.currentThread().isVirtual() :
```
kermandev, line 347: ignore for ServerFlag.INSIDE_TEST?

TheMode: I don't think it should. Special cases should be handled separately

kermandev: maybe but in testing we want blocking tests until completion.

TheMode: You can block inside the test. Some events may not have a guaranteed firing timing (like chunk un/load) and having to handle this case for every single async event will quickly become awkward, on top of missing potential breaking changes.

kermandev: Are we missing conformance tests for this then, to ensure it throws? Also why not make this a runtime check as well, as you can implement this interface?

src/main/java/net/minestom/server/extras/lan/OpenToLAN.java
```diff
@@ -103,20 +103,21 @@ public static boolean isOpen() {
...
-            eventCooldown.refreshLastUpdate(System.nanoTime());
-        }
-
-        try {
-            socket.send(packet);
-        } catch (IOException e) {
-            LOGGER.warn("Could not send Open to LAN packet!", e);
-        }
+        Thread.startVirtualThread(() -> {
+            if (packet == null || eventCooldown.isReady(System.nanoTime())) {
```
mworzala: should catch over this entire thing and send to exception handler

kermandev (approved):

PR #2856, opened 2025-07-21
initial implementation of transfer intent

Sup33r:
Updated version of https://github.com/Minestom/Minestom/pull/2484

Resolves https://github.com/Minestom/Minestom/issues/2196, closes https://github.com/Minestom/Minestom/pull/2473, https://github.com/Minestom/Minestom/pull/2484

Adds OutgoingTransferEvent.
Adds PlayerConnection#transfer(String, int).
Adds PlayerConnection#wasTransferred().
Adds /transfer command to the demo project.
Tested with bungee, and works with both transfers to and from.

### DIFF
--- a/src/main/java/net/minestom/server/network/player/PlayerConnection.java
+++ b/src/main/java/net/minestom/server/network/player/PlayerConnection.java
@@ -7,6 +7,7 @@ import net.minestom.server.crypto.PlayerPublicKey;
 import net.minestom.server.entity.Entity;
 import net.minestom.server.entity.Player;
 import net.minestom.server.event.EventDispatcher;
+import net.minestom.server.event.player.OutgoingTransferEvent;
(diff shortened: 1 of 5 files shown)


src/main/java/net/minestom/server/network/player/PlayerConnection.java
```diff
@@ -40,6 +42,7 @@ public abstract class PlayerConnection {
     private volatile ConnectionState connectionState;
     private PlayerPublicKey playerPublicKey;
     volatile boolean online;
+    private Boolean wasTransferred;
```
AstroSquared, line 45: doesn't look like `wasTransferred` is ever set to false, well, anywhere? so it appears the `wasTransferred` method will only throw an error or return true

mworzala:
Superseded by https://github.com/Minestom/Minestom/pull/2484

PR #2857, opened 2025-07-22
Acquirable strict

TheMode:
Updated version of #2146

Most notably: `Acquirable#assertOwnership` throwing if you do not own an entity and either `ServerFlag.ACQUIRABLE_STRICT` is true or assertions are enabled. Flag is false by default.

instanceSet currently failing because I had to remove the assertion as failing when joining outside of tests.

### DIFF
--- a/build-src/src/main/kotlin/minestom.java-library.gradle.kts
+++ b/build-src/src/main/kotlin/minestom.java-library.gradle.kts
@@ -58,6 +58,7 @@ tasks.withType<Test> {
     // Viewable packets make tracking harder. Could be re-enabled later.
     jvmArgs("-Dminestom.viewable-packet=false")
     jvmArgs("-Dminestom.inside-test=true")
+    jvmArgs("-Dminestom.acquirable-strict=true")
     minHeapSize = "512m"
     maxHeapSize = "1024m"
 }


src/main/java/net/minestom/server/thread/Acquirable.java
```diff
@@ -20,15 +21,12 @@ public sealed interface Acquirable<T> permits AcquirableImpl {
...
-    static @NotNull Stream<@NotNull Entity> currentEntities() {
-        final Thread currentThread = Thread.currentThread();
-        if (currentThread instanceof TickThread) {
-            return ((TickThread) currentThread).entries().stream()
-                    .flatMap(partitionEntry -> partitionEntry.elements().stream())
-                    .filter(tickable -> tickable instanceof Entity)
-                    .map(tickable -> (Entity) tickable);
-        }
-        return Stream.empty();
+    static @NotNull Stream<@NotNull Entity> locals() {
```
mworzala: should keep `localEntities` I guess because acquirable is technically not restricted to entity

src/main/java/net/minestom/server/thread/AcquirableImpl.java
```diff
@@ -1,66 +1,145 @@
...
-        VarHandle.releaseFence();
     }
 
-    static @Nullable ReentrantLock enter(@NotNull Thread currentThread, @Nullable TickThread elementThread) {
-        if (elementThread == null) return null;
-        if (currentThread == elementThread) return null;
-        final ReentrantLock currentLock = currentThread instanceof TickThread ? ((TickThread) currentThread).lock() : null;
-        final ReentrantLock targetLock = elementThread.lock();
-        if (targetLock.isHeldByCurrentThread()) return null;
+    private static final boolean ASSERTIONS_ENABLED = AcquirableImpl.class.desiredAssertionStatus();
```
mworzala: move to top of class

src/main/java/net/minestom/server/thread/AcquirableOwnershipException.java
```diff
@@ -0,0 +1,86 @@
+package net.minestom.server.thread;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Exception thrown when an acquirable element is accessed without proper ownership.
+ */
+public final class AcquirableOwnershipException extends RuntimeException {
```
mworzala, line 10: this is nice, big fan

mworzala (approved):

mworzala:
oh wrong status but whatever the changes are nitpicky so close enough

src/main/java/net/minestom/server/thread/Acquired.java
```diff
@@ -13,31 +10,8 @@
...
-        safeCheck();
-        return value;
-    }
-
-    public void unlock() {
-        safeCheck();
-        this.unlocked = true;
-        AcquirableImpl.leave(lock);
-    }
+public sealed interface Acquired<T> permits AcquiredImpl {
```
kermandev, line 13: Couldnt this extend `Supplier<T>`?

TheMode: I guess it could, though its a temporary object that will throw once unlocked

PR #2865, opened 2025-07-28
Fix loading block entities from vanilla worlds

Fisher2911:
## Proposed changes

This will allow the `AnvilLoader` to not save all blocks with block handlers as "Block Entities". For example if I register a block handler for dirt, and then place thousands of dirt, currently a block entity is saved for every single dirt block. This is unnecessary because the block handler can just be registered in the block manager and is applied when the block is set in the instance.

This also fixes `AnvilLoader` no longer loading registered block handlers for vanilla world blocks that aren't block entities.

## Types of changes

## Checklist

merging your code._

## Further comments

I'm not sure about the naming of the method `defaultHandler()`, let me know if it should be changed.


Fisher2911:
Actually this should be double checked because [https://github.com/Minestom/Minestom/commit/d8da67c307151163d32720622aab473ccd0af127](https://github.com/Minestom/Minestom/commit/d8da67c307151163d32720622aab473ccd0af127) might mean this doesn't work because block handlers are no longer set when loading it looks like.

Fisher2911:
Benchmark:
```
Benchmark                                                 Mode  Cnt          Score            Error   Units
AnvilLoaderLoadBenchmark.chunkLoading                     avgt   30  952037946.333 ± 1269349917.905   ns/op
AnvilLoaderLoadBenchmark.chunkLoading:gc.alloc.rate       avgt   30         11.803 ±          5.157  MB/sec
AnvilLoaderLoadBenchmark.chunkLoading:gc.alloc.rate.norm  avgt   30    5664968.516 ±    1051407.372    B/op
AnvilLoaderLoadBenchmark.chunkLoading:gc.count            avgt   30          9.000                   counts
AnvilLoaderLoadBenchmark.chunkLoading:gc.time             avgt   30        185.000                       ms
AnvilLoaderLoadBenchmark.chunkLoading:stack               avgt                 NaN                      ---
```

PR #2866, opened 2025-07-29
Block entity incorrectly saving fix

Fisher2911:
## Proposed changes

This fixes block entities not loading correctly if they aren't in the zeroth section of the chunk. The y level was saved as the section y level and not the actual y level. The change that caused this is here [https://github.com/Minestom/Minestom/commit/dbd342bac44525359c61aed99a4e83a530fb4af5#diff-e87892fe7151975ae081eca1960bd979a7a57956c0856311b28317edaf3723cc](https://github.com/Minestom/Minestom/commit/dbd342bac44525359c61aed99a4e83a530fb4af5#diff-e87892fe7151975ae081eca1960bd979a7a57956c0856311b28317edaf3723cc).

## Types of changes

## Checklist

merging your code._

## Further comments

It would be nice if this could be fixed quickly because anyone who loads and then saves a world using the current version will have all their block entities destroyed from what I can tell.

### DIFF
--- a/src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
+++ b/src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
@@ -321,13 +321,16 @@ public class AnvilLoader implements IChunkLoader {
         final int chunkZ = chunk.getChunkZ();
 
         // Find the region file or create an empty one if missing
+
+        final int regionX = CoordConversion.chunkToRegion(chunkX);
+        final int regionZ = CoordConversion.chunkToRegion(chunkZ);
+
         RegionFile mcaFile;
         fileCreationLock.lock();
         try {
             mcaFile = getMCAFile(chunkX, chunkZ);
+
             if (mcaFile == null) {
-                final int regionX = CoordConversion.chunkToRegion(chunkX);
-                final int regionZ = CoordConversion.chunkToRegion(chunkZ);
                 final String regionFileName = RegionFile.getFileName(regionX, regionZ);
                 try {
                     Path regionFile = regionPath.resolve(regionFileName);
@@ -346,6 +349,14 @@ public class AnvilLoader implements IChunkLoader {
             }
         } finally {
             fileCreationLock.unlock();
+
+            this.perRegionLoadedChunksLock.lock();
+            try {
+                this.perRegionLoadedChunks.computeIfAbsent(new IntIntImmutablePair(regionX, regionZ), k -> new HashSet<>())
+                        .add(new IntIntImmutablePair(chunkX, chunkZ));
+            } finally {
+                this.perRegionLoadedChunksLock.unlock();
+            }
         }
 
         try {
--- a/src/test/java/net/minestom/server/instance/anvil/AnvilLoaderIntegrationTest.java
+++ b/src/test/java/net/minestom/server/instance/anvil/AnvilLoaderIntegrationTest.java
@@ -319,6 +319,61 @@ public class AnvilLoaderIntegrationTest {
         instance.loadChunk(point).join();
     }
 
+    @Test
+    public void saveChunks(Env env) throws IOException {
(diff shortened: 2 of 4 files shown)


src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
```diff
@@ -421,7 +422,7 @@ private void saveSectionData(@NotNull Chunk chunk, @NotNull CompoundBinaryTag.Bu
                             if (originalNBT != null) blockEntityTag.put(originalNBT);
                             if (handler != null) blockEntityTag.putString("id", handler.getKey().asString());
                             blockEntityTag.putInt("x", x + Chunk.CHUNK_SIZE_X * chunk.getChunkX());
-                            blockEntityTag.putInt("y", y);
+                            blockEntityTag.putInt("y", y + (finalSectionY * Chunk.CHUNK_SECTION_SIZE));
```
kermandev: Please use `y + globalSectionY` instead of attempting to capture sectionY again

src/test/java/net/minestom/server/instance/anvil/AnvilLoaderIntegrationTest.java
```diff
@@ -237,10 +239,23 @@ public void loadAndSaveBlockNBT(Env env) throws IOException {
     }
 
     @Test
-    public void loadAndSaveBlockHandler(Env env) throws IOException, InterruptedException {
+    public void loadAndSaveBlockHandler(Env env) throws IOException {
+        loadAndSaveBlockHandler(env, BlockVec.ZERO);
+        loadAndSaveBlockHandler(env, BlockVec.ZERO.add(0, 15, 0));
```
kermandev: new BlockVec(0, 15,0), instead of adding from zero please.

Ideally just make this a parameterized test, but this is fine.

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -127,10 +128,9 @@ public void setBlock(int x, int y, int z, @NotNull Block block,
         if (handler != null) {
             // New placement
 
-            var absoluteBlockPosition = new Vec(getChunkX() * 16 + x, y, getChunkZ() * 16 + z);
             final Block finalBlock = block;
             handler.onPlace(Objects.requireNonNullElseGet(placement,
-                    () -> new BlockHandler.Placement(finalBlock, instance, absoluteBlockPosition)));
+                    () -> new BlockHandler.Placement(finalBlock, instance, blockPosition)));
```
kermandev: This change is wrong as generators are section relative from 0-15; because we pass an Instance they have to be the global coordinates The correct fix is included in #2581 if you want to copy it and its accompanying test.

kermandev (changes requested):

src/main/java/net/minestom/server/instance/DynamicChunk.java
Fisher2911: Doesn't `Chunk#setBlock` us global coordinates? Otherwise it seems like `InstanceContainer` is wrong https://github.com/Minestom/Minestom/blob/2341c1dec2e1c866267b9ac48a331501e090f4eb/src/main/java/net/minestom/server/instance/InstanceContainer.java#L136.
I'm pretty sure the code in that PR and directly using the `x`, `y`, `z` parameters end up giving the same position, but I could be wrong.

kermandev: This is not the case for generators if i remember correctly, because the chunk does accepts relative coordinates even it messes up the block index later. We still have to reconstruct the global Vec if we allow any case of relative setting inside the chunk.

Just reconstruct the Vec, not a BlockVec (breaking change), from the sectionRealativeX and sectionRealativeZ, the Y should be the global coordinate.

kermandev (changes requested):
See above comment

PR #2867, opened 2025-07-29
Fix AnvilLoader not closing the region file if chunks weren't originally loaded by AnvilLoader

Fisher2911:
## Proposed changes

This fixes `AnvilLoader` not closing region files if chunks weren't originally loaded by `AnvilLoader`. This can happen when generating a new instance and then saving the chunks. Then if you unregister the instance or unload the chunks, the region files are not closed.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
+++ b/src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
@@ -321,21 +321,14 @@ public class AnvilLoader implements IChunkLoader {
         final int chunkZ = chunk.getChunkZ();
 
         // Find the region file or create an empty one if missing
+
+        final int regionX = CoordConversion.chunkToRegion(chunkX);
+        final int regionZ = CoordConversion.chunkToRegion(chunkZ);
+
         RegionFile mcaFile;
         fileCreationLock.lock();
         try {
             mcaFile = getMCAFile(chunkX, chunkZ);
-
-            final int regionX = CoordConversion.chunkToRegion(chunkX);
-            final int regionZ = CoordConversion.chunkToRegion(chunkZ);
-
-            this.perRegionLoadedChunksLock.lock();
-            try {
-                this.perRegionLoadedChunks.computeIfAbsent(new IntIntImmutablePair(regionX, regionZ), k -> new HashSet<>())
-                        .add(new IntIntImmutablePair(chunkX, chunkZ));
-            } finally {
-                this.perRegionLoadedChunksLock.unlock();
-            }
 
             if (mcaFile == null) {
                 final String regionFileName = RegionFile.getFileName(regionX, regionZ);
@@ -356,6 +349,14 @@ public class AnvilLoader implements IChunkLoader {
             }
         } finally {
             fileCreationLock.unlock();
+
+            this.perRegionLoadedChunksLock.lock();
+            try {
+                this.perRegionLoadedChunks.computeIfAbsent(new IntIntImmutablePair(regionX, regionZ), k -> new HashSet<>())
+                        .add(new IntIntImmutablePair(chunkX, chunkZ));
+            } finally {
+                this.perRegionLoadedChunksLock.unlock();
+            }
         }
 
         try {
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
```diff
@@ -343,6 +343,13 @@ public void saveChunk(@NotNull Chunk chunk) {
                     MinecraftServer.getExceptionManager().handleException(e);
                     return;
                 }
+                try {
+                    this.perRegionLoadedChunksLock.lock();
+                    this.perRegionLoadedChunks.computeIfAbsent(new IntIntImmutablePair(regionX, regionZ), k -> new HashSet<>())
+                            .add(new IntIntImmutablePair(chunkX, chunkZ));
+                } finally {
+                    this.perRegionLoadedChunksLock.unlock();
+                }
```
kermandev: Pretty sure this fix will fail since after the first chunk lets say 0,0 is created a MCA file, since the MCA file exists this wont be added to the loaded chunks, I think we have to pull this out and do it for every call.

src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
```diff
@@ -343,6 +343,13 @@ public void saveChunk(@NotNull Chunk chunk) {
                     MinecraftServer.getExceptionManager().handleException(e);
                     return;
                 }
+                try {
+                    this.perRegionLoadedChunksLock.lock();
```
kermandev: Please pull out this lock out of the try

kermandev (changes requested):
Please add a test to ensure regions are being saved correctly

Fisher2911:
> Please add a test to ensure regions are being saved correctly

I added a test, if you want me to make any changes just let me know.

src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
```diff
@@ -325,9 +325,19 @@ public void saveChunk(@NotNull Chunk chunk) {
...
+            final int regionX = CoordConversion.chunkToRegion(chunkX);
+            final int regionZ = CoordConversion.chunkToRegion(chunkZ);
+
+            this.perRegionLoadedChunksLock.lock();
+            try {
+                this.perRegionLoadedChunks.computeIfAbsent(new IntIntImmutablePair(regionX, regionZ), k -> new HashSet<>())
+                        .add(new IntIntImmutablePair(chunkX, chunkZ));
+            } finally {
+                this.perRegionLoadedChunksLock.unlock();
+            }
```
kermandev: This should not be here locking the file creation, pull this out and put it after the file creation is probably fine.

src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
```diff
@@ -346,6 +349,14 @@ public void saveChunk(@NotNull Chunk chunk) {
             }
         } finally {
             fileCreationLock.unlock();
+
+            this.perRegionLoadedChunksLock.lock();
```
kermandev, line 353: This doesnt need to be in the finally block

Fisher2911: If there is an error (other than `IOException`) anywhere after `getMCAFile` succeeds then the `perRegionLoadedChunks` will be missing the coordinate, but the file will still have been created so it might not be released later.

kermandev: Possibly but that error would bubble up into a throw, and if we throw an IOException we dont have the file to save to (so how is the data the same?), so it makes no sense to possibly create in this finally block.

We actually should probably move this after we write the data out (below mcaFile.writexxx (after the try block & not in a finally block) cause that would mean it failed to write, which means the MCA region cant be loaded from.

Should probably make perRegionLoadedChunks should only hold data which could be reconstructed from the file in this case. Which would mean the putting this addition at the end if it was written to fully.

Fisher2911: If an error other than an IOException is thrown though then that means the file might have been created, and after the finally block executes nothing else in the method will execute unless this is done in the finally block.

The whole point of this is to ensure that the file gets released properly when unloading the chunks, so it needs to be done whenever the file is created.

Another option is to just move it to the `getMCAFile` method.

kermandev: Yeah, this is probably good enough for a bug fix without a bigger refactor then.

kermandev (approved):

PR #2868, opened 2025-07-29
Default block handlers

Fisher2911:
## Proposed changes

This adds default handlers for blocks to prevent having to create new blocks repeatedly for the same blocks. It also allows people to just register default handlers and then not need to worry about setting handlers every time they set a block in an instance.
This also allows `AnvilLoader` to no longer need to load block handlers for non-block entities.

This uses the branch from https://github.com/Minestom/Minestom/pull/2865 so that pull request is no longer necessary if this one gets merged.

## Types of changes

## Checklist

merging your code._

## Further comments

I think this is useful because it could potentially prevent thousands or even hundreds of thousands of unnecessary block objects from being created. One example would be a superflat world with a height of 40 that has only stone blocks. If you have 4 chunks loaded, that would be 16 * 16 * 40 *  4 = 40,960 blocks. If you wanted a block handler for stone blocks, currently a new block instance is created for each world if loading using `AnvilLoader`, while this PR makes it only a single stone block instance.

This is marked as a breaking change because it's different behavior than from before https://github.com/Minestom/Minestom/commit/d8da67c307151163d32720622aab473ccd0af127.

One potential concern could be that retrieving block handlers is slower now, so to fix that we could potentially add back the old system and just use both, where `AnvilLoader` would still assign block handlers directly to blocks.

### DIFF
--- a/jmh-benchmarks/src/jmh/java/net/minestom/server/instance/loading/AnvilLoaderLoadBenchmark.java
+++ b/jmh-benchmarks/src/jmh/java/net/minestom/server/instance/loading/AnvilLoaderLoadBenchmark.java
@@ -1,10 +1,12 @@
 package net.minestom.server.instance.loading;
 
+import net.kyori.adventure.key.Key;
 import net.minestom.server.MinecraftServer;
 import net.minestom.server.instance.Instance;
 import net.minestom.server.instance.InstanceContainer;
 import net.minestom.server.instance.anvil.AnvilLoader;
 import net.minestom.server.instance.block.Block;
+import net.minestom.server.instance.block.BlockHandler;
 import net.minestom.server.world.DimensionType;
 import org.jetbrains.annotations.NotNull;
 import org.openjdk.jmh.annotations.*;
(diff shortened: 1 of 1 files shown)


Fisher2911:
Benchmark: 
```
Benchmark                                                 Mode  Cnt           Score           Error   Units
AnvilLoaderLoadBenchmark.chunkLoading                     avgt   30  1108316035.736 ± 699578621.363   ns/op
AnvilLoaderLoadBenchmark.chunkLoading:gc.alloc.rate       avgt   30           2.184 ±         1.000  MB/sec
AnvilLoaderLoadBenchmark.chunkLoading:gc.alloc.rate.norm  avgt   30     1420540.851 ±    577708.068    B/op
AnvilLoaderLoadBenchmark.chunkLoading:gc.count            avgt   30           2.000                  counts
AnvilLoaderLoadBenchmark.chunkLoading:gc.time             avgt   30          20.000                      ms
AnvilLoaderLoadBenchmark.chunkLoading:stack               avgt                  NaN                     ---
```

PR #2871, opened 2025-07-31
Block handler destroy

Fisher2911:
## Proposed changes

This allows block handlers to see the previous block when using `Placement`, and the new block when using `Destroy`.
This is potentially a breaking change for people who directly use the `Placement`, `PlayerPlacement`, `Destroy`, and `PlayerDestroy` constructors, but they *are* marked as internal API.

## Types of changes

## Checklist

merging your code._

## Further comments

This is useful because currently whenever you use `Instance#setBlock`, it calls the destroy and then the placement. This means that if you do something like adding NBT to a block (for example storing a chest inventory), the placement and destroy are both called. If you are storing the block somewhere else (for example you want to store block entities separately so you can access them without iterating over all blocks in the instance), then this would mean you have to remove the block entity and then add it back, which is unnecessary in this case because the block you are destroying and then placing are for all intents and purposes the same block.

This allows you to do something like this:

```java
@Override
public void onDestroy(@NotNull Destroy destroy) {
  if (destroy.getNewBlock().handler() == this) {
    return;
  }
  // remove stored block entity
}

@Override
public void onPlace(@NotNull Placement placement) {
  if (getBlockIdFor(placement.getBlock()).equals(getBlockIdFor(placement.getPreviousBlock)) && placement.getPreviousBlock().handler() == this) {
    return;
  }
  // add stored block entity
}
```
where `getBlockIdFor` is some way of identifying a block, such as storing a `Key` in the block NBT.

This is also useful for cases where a block display other things visually, for example a crafting table that might display the currently crafted items using item displays, and stores the currently items in the inventory in the NBT as they are placed in the inventory. With the current system, you would have to remove all the item entities every time you updated the inventory and then respawn them, but with this new way you would have the ability to only need to remove 1 item entity at most, and add 1 item entity at most.


src/main/java/net/minestom/server/instance/block/BlockHandler.java
```diff
@@ -107,6 +109,10 @@ public Placement(Block block, Instance instance, Point blockPosition) {
             return block;
         }
 
+        public @NotNull Block getPreviousBlock() {
```
kermandev: This can be null, this is has weird behavior as the lastCachedBlock is not done for every block (only ones inside the entries, since we dont pass the existing stateId back from the Palette#set operation, its kind of hard to always get the correct block here.

Fisher2911: Oh I thought it returned `AIR` if it didn't exist, I'll fix that thank you.

src/main/java/net/minestom/server/instance/block/BlockHandler.java
```diff
@@ -93,12 +95,14 @@ default byte getBlockEntityAction() {
      */
     sealed class Placement permits PlayerPlacement {
         private final Block block;
+        private final Block previousBlock;
         private final Instance instance;
         private final Point blockPosition;
 
         @ApiStatus.Internal
-        public Placement(Block block, Instance instance, Point blockPosition) {
+        public Placement(Block block, @Nullable Block previousBlock,  Instance instance, Point blockPosition) {
```
mworzala: double space

src/main/java/net/minestom/server/instance/block/BlockHandler.java
```diff
@@ -107,6 +111,10 @@ public Placement(Block block, Instance instance, Point blockPosition) {
             return block;
         }
 
+        public @Nullable Block getPreviousBlock() {
```
mworzala: it seems pretty confusing for this to be nullable from an api perspective. Ie there _is_ a block there it just didn't have a handler so we don't look it up.

I think it's worth looking up the block in this scenario for clarity (specifically only right before this is called if lastCachedBlock is null). It is another lookup in relatively hot code, but it is only when specifically you set a block handler to a block which did not have one, so I think it's acceptable.

cc @kermandev if you have opinions

mworzala:
p.s. needs rebase

src/main/java/net/minestom/server/instance/block/BlockHandler.java
kermandev: the block palette should instead be updated to return the existing stateId so you could get an accurate reading before doing this addition, so it could have consistent behavior.

mworzala: I definitely am not convinced on that, doing the extra palette read when entirely unnecessary because you only care about block handler seems less than ideal

Fisher2911: > it seems pretty confusing for this to be nullable from an api perspective. Ie there _is_ a block there it just didn't have a handler so we don't look it up.
> 
> I think it's worth looking up the block in this scenario for clarity (specifically only right before this is called if lastCachedBlock is null). It is another lookup in relatively hot code, but it is only when specifically you set a block handler to a block which did not have one, so I think it's acceptable.
> 
> cc @kermandev if you have opinions

This makes sense, I can do this.

mworzala (approved):

PR #2872, opened 2025-07-31
Chunk block entities

Fisher2911:
## Proposed changes

Right now the chunk stores a block as a block entity if it's a vanilla block entity, but that doesn't seem to be necessary if they don't have NBT or a block handler. The only thing I can think of is that the client needs to be sent the block entities, but I don't think that's the case. I don't think this is a breaking change but I'm not completely sure so I checked it off. 

## Types of changes

## Checklist

merging your code._


TheMode:
Did you verify it? I believe that clients dont render block entities unless specified in the chunk packet.

Fisher2911:
> Did you verify it? I believe that clients dont render block entities unless specified in the chunk packet.

You're right, I had only tested it by placing blocks and not generating chunks.

PR #2873, opened 2025-08-01
Convert numbers to their most specific type in JSON transcoder

Fisher2911:
## Proposed changes

This fixes the JSON encoder always using double for number values.

## Types of changes

## Checklist

merging your code._


Fisher2911:
I'm not sure what the best way to fix this for lists/arrays would be.

mworzala:
When parsing using gson you always get `LazilyParsedNumber` values anyway, which will still result in a double no matter what. There is a more complex way to accomplish this, see: https://github.com/Mojang/DataFixerUpper/blob/d0f713b0719b5f1ad77506e6a019c41782b8bcf1/src/main/java/com/mojang/serialization/JsonOps.java#L57

Fisher2911:
> When parsing using gson you always get `LazilyParsedNumber` values anyway, which will still result in a double no matter what. There is a more complex way to accomplish this, see: https://github.com/Mojang/DataFixerUpper/blob/d0f713b0719b5f1ad77506e6a019c41782b8bcf1/src/main/java/com/mojang/serialization/JsonOps.java#L57

I'm not sure if that will work either because that looks like it's just taking the smallest possible number type. With the way the values are stored in JSON format, this might not be possible without changing the CODEC to allow a collection of tags to deserialize to be passed in to the deserialization.

For example, if you store `{ myFavoriteNumber : 1 }`, that would become a byte tag, but the original tag might be an integer tag.

So maybe there could be another method for CODECS like `Codec.NBT_COMPOUND.decode(Transcoder, JSON, Collection<Tag>)`.

mworzala:
The smallest number is often fine, the reality is that json is a lossy format for numbers so there's only so much you can do. but more specifically minecraft uses the smallest valid approach of DFU so we should match that anyway

PR #2875, opened 2025-08-02
Update Codegen

kermandev:
## Proposed changes
Current JavaPoet is out of date, so I updated to the recommended replacement. Ive also cleaned up codegen to prevent future mistakes, also sealed the generated classes to ensure they are implemented so the constants become public.

## Types of changes

## Further comments
Throwing on failure should prevent particles.json from going unnoticed, as it got renamed to particle.json 
Sealed classes should prevent issues like GameEvent missing its extends GameEvents. 
Removed SLF4j as its not required for 2 print statements anymore.


mworzala:
needs rebase but otherwise seems like a nice cleanup (please just merge whenever rebased)

PR #2876, opened 2025-08-03
Auth interface rework

TheMode:
Avoid the globally mutable auth state, simplify pattern matching.

Only tested with offline/online. Although I just moved the code around.


TheMode:
I welcome naming suggestions, Bungee/Velocity is perhaps not the most ideal, but neither is Legacy/Modern

emortaldev:
Either would be fine, I personally prefer Legacy/Modern, maybe with Forwarding somewhere in the name but that's a little long

mworzala (approved):
Looks good but needs rebase on top of notnullbydefault

PR #2877, opened 2025-08-03
optimize palette resizing

AidanMars2:
## Proposed changes

This PR optimizes palette resizing by using a specialized function for it.
There are a couple bug fixes associated with this PR:
- remove the reliance on the assumption that `valueToPaletteIndex(0) == 0`, Notchian anvil files don't follow this assumption.
- `optimize(Optimization.SIZE)` can now actually downsize the palette.
- `valueToPaletteIndex` doesn't prematurely resize anymore.

## Types of changes

## Checklist

merging your code._

## Further comments

No further tests were added as the compare function was upgraded with an early return if count doesn't match.


TheMode:
I think the first bug fix could have some tests.

I am also a bit unsure about the `computeIfAbsent` call being faster

AidanMars2:
I benchmarked the `computeIfAbsent` call against the old version, turns out the function creation makes it 3x slower. It also turns out that the assert at the end of the old version nearly halved performance.

PR #2881, opened 2025-08-06
Use the BlocksAttacks component in UseItemListener

GreatWyrm:
## Proposed changes

The blocking animation does not properly show on other clients if you aren't using a shield, due to only checking for the shield material rather than the component.

Additionally also fixes a minor bug in the BlocksAttacks component where the Sound Events in the network serializer were not marked as optional.

## Types of changes

## Checklist

merging your code._

## Further comments

~~Will do some quick testing to ensure this is the proper fix and will update this text once this is done~~
Completed some testing, before the change another client would not see the blocking animation if the item was not a shield, but still had the blocks attacks component

### DIFF
--- a/src/main/java/net/minestom/server/listener/UseItemListener.java
+++ b/src/main/java/net/minestom/server/listener/UseItemListener.java
@@ -11,6 +11,7 @@ import net.minestom.server.inventory.PlayerInventory;
 import net.minestom.server.item.ItemAnimation;
 import net.minestom.server.item.ItemStack;
 import net.minestom.server.item.Material;
+import net.minestom.server.item.component.BlocksAttacks;
 import net.minestom.server.item.component.Consumable;
 import net.minestom.server.item.component.Equippable;
 import net.minestom.server.item.component.InstrumentComponent;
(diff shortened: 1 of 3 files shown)


src/main/java/net/minestom/server/listener/UseItemListener.java
```diff
@@ -40,7 +42,7 @@ public static void useItemListener(ClientUseItemPacket packet, Player player) {
             // client they can hold it forever
             useItemTime = 7200;
             useAnimation = ItemAnimation.CROSSBOW;
-        } else if (material == Material.SHIELD) {
+        } else if (blocksAttacks != null) {
```
mudkipdev:
```suggestion
        } else if (itemStack.has(DataComponents.BLOCKS_ATTACKS)) {
```

GreatWyrm: Done, thanks for the suggestion

mworzala (commented):
would like to see a test for the new behavior (and old matching shield items which should still have the component in prototype)

GreatWyrm:
@mworzala Added

PR #2884, opened 2025-08-10
Automatically generate Graal native-image metadata from tests

TheMode:
There may be a better way to do it, and there should perhaps be an exclusion list to not include gradle and test classes

### DIFF
--- a/src/main/java/net/minestom/server/advancements/Advancement.java
+++ b/src/main/java/net/minestom/server/advancements/Advancement.java
@@ -93,7 +93,7 @@ public class Advancement {
         return tab;
     }
 
-    protected void setTab(AdvancementTab tab) {
+    protected void setTab(@Nullable AdvancementTab tab) {
         this.tab = tab;
     }
 
@@ -283,7 +283,7 @@ public class Advancement {
         return parent;
     }
 
-    protected void setParent(Advancement parent) {
+    protected void setParent(@Nullable Advancement parent) {
         this.parent = parent;
     }
 
--- a/src/main/java/net/minestom/server/command/builder/Command.java
+++ b/src/main/java/net/minestom/server/command/builder/Command.java
@@ -11,6 +11,7 @@ import net.minestom.server.command.builder.condition.CommandCondition;
 import net.minestom.server.utils.StringUtils;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.UnknownNullability;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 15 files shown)


src/main/resources/META-INF/native-image/minestom/README.md
```diff
@@ -0,0 +1,24 @@
+# GraalVM Native Image Metadata
```
mworzala: this file should not be in resources, it will be included in the final jar :|

src/main/resources/META-INF/native-image/minestom/serialization-config.json
```diff
@@ -1,2 +1,101 @@
...
+      "name":"java.util.concurrent.locks.ReentrantLock"
+    },
+    {
+      "name":"java.util.concurrent.locks.ReentrantLock$NonfairSync"
+    },
+    {
+      "name":"java.util.concurrent.locks.ReentrantLock$Sync"
+    },
+    {
+      "name":"org.gradle.api.internal.tasks.testing.filter.TestFilterSpec"
```
mworzala: everything here and below should be filtered out

build.gradle.kts
```diff
@@ -56,6 +56,49 @@ tasks.jar {
     }
 }
 
+// GraalVM Native Image configuration
+tasks.register<Test>("testWithAgent") {
+    group = "verification"
+    description = "Runs all tests with GraalVM native-image agent to generate metadata."
+
+    val metadataOutputDir = layout.buildDirectory.dir("native-image-metadata").get().asFile
+    val resourcesTargetDir = layout.projectDirectory.dir("src/main/resources/META-INF/native-image/minestom").asFile
```
mworzala: nitpick I guess, but conventionally the directory for our metadata should be

```
META-INF/native-image/net.minestom/minestom/*.json
```

build.gradle.kts
```diff
@@ -56,6 +56,49 @@ tasks.jar {
...
+    doFirst {
+        delete(metadataOutputDir)
+        metadataOutputDir.mkdirs()
+
+        filterFile.parentFile.mkdirs()
+        filterFile.writeText("""
+{
+  "rules": [
+    {"excludeClasses": "org.junit.**"},
+    {"excludeClasses": "org.opentest4j.**"},
```
mworzala, line 79: seems like this either isnt working or needs to be regenerated since being added, since there are still targeted classes in the output.

TheMode: It works for the reflect-config but not the rest

src/main/resources/META-INF/native-image/minestom/serialization-config.json
TheMode: Didn't figure out how

TheMode:
Demo server tested using
```
./gradlew demo:shadowJar
native-image -jar demo/build/libs/minestom-demo.jar --no-fallback -o minestom-demo-native
```

Though without logging as Minestom doesn't log unless an slf4j implementation is specified

PR #2885, opened 2025-08-12
Remove NotNull annotations from tests

TheMode:


kermandev:
Probably need to add package infos and modules to tests, and demo to actually use the not null by default annotation.

kermandev (approved):
Ive changed my view on doing modules on test as doing modules on tests is not preferred as if done would require things that are inherently file private to be opened for tests as they cannot be under the same module.

The slight cleanup in the main nullability is also fine. (This is going to be a requirement for alot of the project as nullability annotations usually fell after the first field implementation.

PR #2887, opened 2025-08-14
Fix nullability documentation of Codec

kermandev:
## Proposed changes

Fixes nullability warnings across Minestom with codecs.

## Types of changes

## Further comments
Cleanup some other warnings in the codec package
Cleanup union's with revised version
Changed how optional defaultValue cannot be null as optional exists and we can gaurrentee its nullable.
Yes I know the extends Object, looks bad but thats how the spec is works (jspecify does this too)
Maybe result should not allow nullable values and instead always be an null error then a Result#orNull?


mworzala:
nvm its fine for some reason I thought there was a conflict

PR #2891, opened 2025-08-20
Accessible block entities

Fisher2911:
## Proposed changes

This allows you to get the block entities in a chunk directly, instead of iterating over all blocks.

## Types of changes

## Checklist

merging your code._


kermandev:
Whats the use case for this; your not given block locations so you could only really check if there is block entities/block handlers inside this chunk.

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -337,7 +338,19 @@ public ChunkSnapshot updateSnapshot(SnapshotUpdater updater) {
                 tagHandler().readableCopy());
     }
 
+    /**
+     *
+     * @return a copy of the block entities currently in this chunk
+     */
+    @Override
+    public @UnmodifiableView Collection<Block> getBlockEntities() {
```
kermandev: This is the wrong annotation for this

Fisher2911:
> Whats the use case for this; your not given block locations so you could only really check if there is block entities/block handlers inside this chunk.

Sorry about that, I forgot that block entities don't have a reference to their position.

src/main/java/net/minestom/server/instance/Chunk.java
```diff
@@ -265,12 +266,23 @@ public boolean isLoaded() {
...
-    public void onGenerate() {}
+    public void onGenerate() {
+    }
+
+    /**
+     *
+     * @return the current block entities in this chunk, not guaranteed to stay consistent with the chunk's
+     * block entities if block entities are added/removed from the chunk after calling this method.
+     */
+    public @Unmodifiable Map<Point, Block> getBlockEntities() {
```
mworzala: shouldn't be implemented here

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -339,8 +345,59 @@ public ChunkSnapshot updateSnapshot(SnapshotUpdater updater) {
                 tagHandler().readableCopy());
     }
 
+    /**
+     * @return a copy of the block entities currently in this chunk
+     */
+    @Override
+    public @Unmodifiable Map<Point, Block> getBlockEntities() {
+        synchronized (this) {
+            return this.entries.int2ObjectEntrySet().stream()
```
kermandev: Streamless version please

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -339,8 +345,59 @@ public ChunkSnapshot updateSnapshot(SnapshotUpdater updater) {
                 tagHandler().readableCopy());
     }
 
+    /**
+     * @return a copy of the block entities currently in this chunk
+     */
+    @Override
+    public @Unmodifiable Map<Point, Block> getBlockEntities() {
+        synchronized (this) {
```
kermandev:
```suggestion
    public synchronized @Unmodifiable Map<Point, Block> getBlockEntities() {
```

src/main/java/net/minestom/server/instance/Chunk.java
```diff
@@ -265,12 +268,33 @@ public boolean isLoaded() {
...
+     *
+     * Iterates over all block entities in the chunk
+     */
+    public abstract void forEachBlockEntity(BiConsumer<Point, Block> consumer);
+
+    /**
+     * Finds all block entities that match {@code filter}
+     * @return map of blocks that pass the filter
+     */
+    public abstract Map<Point, Block> filterBlockEntities(BiPredicate<Point, Block> filter);
```
kermandev: These dont make sense to have, since the same could achieved with getBlockEntitites(), and these force them to be called inside the synchronized scope, but they are unmodifiable....

src/main/java/net/minestom/server/instance/Chunk.java
```diff
@@ -265,12 +268,33 @@ public boolean isLoaded() {
...
-    public void onGenerate() {}
+    public void onGenerate() {
+    }
+
+    /**
+     *
+     * @return the current block entities in this chunk, not guaranteed to stay consistent with the chunk's
+     * block entities if block entities are added/removed from the chunk after calling this method.
+     */
+    public abstract @Unmodifiable Map<Point, Block> getBlockEntities();
```
kermandev, line 290: Missing javadocs above return.

kermandev (changes requested):

src/main/java/net/minestom/server/instance/Chunk.java
Fisher2911: These would allow you to iterate the entities and perform side effects without having to synchronize again if you use `getBlockEntities`. These are not necessarily unmodifiable, because the `BlockHandler` of the block might not be immutable, so you would be able to modify the block entity in that way.

Fisher2911:
Looks like I accidentally pushed my merge from https://github.com/Minestom/Minestom/pull/3308 to fix the chunk issue locally, if that doesn't end up getting merged in I'll revert that commit.

src/main/java/net/minestom/server/instance/Chunk.java
kermandev: I think its good to keep the simple API surface for this, and remove these methods, we make no synchronization on BlockHandler regardless, as it can be freely shared. Its not necessaryly safe either during the read lock as it can be shared with other threads.

src/test/java/net/minestom/server/instance/DynamicChunkTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.*;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@EnvTest
+public final class DynamicChunkTest {
```
kermandev: Likely can just be ChunkIntegrationTest, as you are adding the methods to Chunk.

PR #2892, opened 2025-08-23
Update Chunk View Distance

### LINKED ISSUE #2455: Chunks at the edge of view distance suddenly pop into existance
Minestom loads chunks at a slightly different distance from vanilla causing the fog to not be effective in hiding chunk loading.
I think this is due to the fact that vanilla adds an additional chunk to the actual view distance when deciding how many chunks to load.

See examples below comparing a vanilla server and a minestom server with equivalent view distance (8) and the same client.

## Vanilla:
vanilla

## Minestom:
minestom

### LINKED ISSUE #2485: Dynamic chunk/entity view distance
Currently, the chunk/entity view distances are set via a server flag, making it impossible to change them during runtime.
Would be very useful to change without having to restart the server entirely.

GreatWyrm:
Bumps the effective view distance of players by 1 to resolve #2455, and adds per-instance view distance to resolve #2485.

## Proposed changes

## Types of changes

- [X] Documentation Update (if none of the other choices apply)

## Checklist

merging your code._

## Further comments

None

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -2377,11 +2377,9 @@ public class Player extends LivingEntity implements CommandSender, HoverEventSou
      * @return The effective chunk view distance range of the client
      */
     public int effectiveViewDistance() {
-        if (this.instance == null) {
-            return Math.min(settings.viewDistance(), ServerFlag.CHUNK_VIEW_DISTANCE) + 1;
-        } else {
-            return Math.min(settings.viewDistance(), this.instance.viewDistance()) + 1;
-        }
+        Instance instance = this.instance;
+        int maxViewDistance = instance != null ? instance.viewDistance() : ServerFlag.CHUNK_VIEW_DISTANCE;
+        return Math.min(settings.viewDistance(), maxViewDistance) + 1;
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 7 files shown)


src/main/java/net/minestom/server/network/player/ClientSettings.java
```diff
@@ -52,10 +52,6 @@ public record ClientSettings(Locale locale, byte viewDistance,
         Objects.requireNonNull(mainHand);
     }
 
-    public int effectiveViewDistance() {
```
mworzala, line 55: deprecate this instead of deleting, then its not a breaking change

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2372,6 +2372,18 @@ private int compareChunkDistance(long chunkIndexA, long chunkIndexB) {
         return Integer.compare(chunkDistanceA, chunkDistanceB);
     }
 
+    /**
+     * Gets the client's 'effective' view distance, which is the minimum of the client's view distance settings, and the local instance settings, plus one
+     * @return The effective chunk view distance range of the client
+     */
+    public int effectiveViewDistance() {
+        if (this.instance == null) {
```
mworzala: Should read the instance once, eg

```java
		Instance instance = this.instance;
		int maxViewDistance = instance != null ? instance.viewDistance() : ServerFlag.CHUNK_VIEW_DISTANCE;
            return Math.min(settings.viewDistance(), maxViewDistance) + 1;
```

(also I think removing the if and only adding once makes it a bit more legible)

src/main/java/net/minestom/server/network/player/ClientSettings.java
GreatWyrm: Done

PR #2902, opened 2025-08-30
Entity AI overhaul

TogAr2:
## Proposed changes

This pull request adds a new entity AI system, to hopefully make working with it easier and allow for extended functionality compared to the current system.

### The main change

The main change is the new `GoalSelector`. The current AI system consist of independent AI groups, each with an active goal and target selectors. In the new system, there is no such thing as a group, all goals will be running in the same goal selector. They can occupy a slot, and slots can be user created, still allowing for multiple goals to be run simultaneously.

The reason for this change is that it allows for a goal to block multiple slots. For example, it is now possible to have an entity which can walk randomly and look randomly at the same time, but when attacking, stop both of these goals. This would be accomplished by creating a `move` slot, and a `look` slot, allowing the attack goal to block both. The setup would be similar to this:

```java
final var move = new GoalSelector.Slot();
final var look = new GoalSelector.Slot();

entity.getAi()
    .addGoal(move, look, new AttackGoal(entity, ...))
    .addGoal(move, new RandomMoveGoal(entity, ...))
    .addGoal(look, new RandomLookGoal(entity, ...))
```

The priority of the goals would be determined by the order of the list, so the attack goal would always be preferred if its `canStart()` method returns `true`.
I am open to any suggestions for improvements of the api :)

### Other changes

Other changes include the abstraction of the `Navigator` class, so it is now possible for users to create a custom navigator implementation, allowing them to change the way pathfinding & following works.

I also changed the implementations of the currently existing goals within Minestom. The attack goals (`MeleeAttackGoal`, `RangedAttackGoal`) are now no longer responsible for navigating towards the entity. This can be done with the `FollowTargetGoal`. `CombinedAttackGoal` has been removed, since the same behavior can be accomplished by using both attack goals in the same slot, with different range parameters.
One note on this: with the removal of navigation for attack goals, the `comeClose` behavior for ranged attacks has also been removed (this meant: if the entity is not in sight, move as close as possible). There is currently no builtin replacement, but one can easily create a custom `FollowTargetGoal` with this behavior.

Another implication of the changes: an entity can no longer have _multiple targets_ at the same time. Previously, target selectors were local to the AI group, and different goals could have different targets. With the new api, the entity's target is synchronized across all goals (`entity.getAi().getTarget()`). Of course it would still possible to recreate the previous behavior by yourself. I personally think it makes sense that all builtin goals just use the centralized entity target.

## Types of changes

## Checklist


mworzala:

mworzala:

kermandev:

PR #2908, opened 2025-09-07
Ensure nullability constraints at runtime for Codecs

kermandev:
Should make known any initialization issues; also fixes #2893 
Also adds type and docs to TranscoderProxy & fixes MinecraftServer#process nullability being dumb. (Immutable server processes will ensure its never null at runtime during the first call).

## Types of changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation Update (if none of the other choices apply)

Issue #2904, opened 2025-09-02
[Bug]: `statistic_type.json` no longer exists and causes an error when trying to use `StatisticType`


cosrnic:
### Affected Version

`2025.08.29-1.21.8`

### Describe the bug

When using `StatisticType` it tries to load the file `statistic_type.json`, however this file no longer exists in the data dependency and causes an error.

It seems like this was mistyped to `statistic_type.json` from `custom_statistics.json` or the data gen wasn't updated with the new name.
https://github.com/Minestom/Minestom/blob/a7951ca0ae34f12e8abf7915967f392c1c4f0698/src/main/java/net/minestom/server/statistic/StatisticTypeImpl.java#L8

log:
```
Exception in thread "main" java.lang.ExceptionInInitializerError
	at net.minestom.server.statistic.StatisticTypes.<clinit>(StatisticTypes.java:10)
	at wales.cosmic.test.Main.main(Main.java:49)
Caused by: java.lang.IllegalArgumentException: Failed to load required registry file: statistic_type.json
	at net.minestom.server.utils.validate.Check.fail(Check.java:53)
	at net.minestom.server.registry.RegistryData.load(RegistryData.java:134)
	at net.minestom.server.registry.RegistryData.createStaticRegistry(RegistryData.java:149)
	at net.minestom.server.statistic.StatisticTypeImpl.<clinit>(StatisticTypeImpl.java:8)
	... 2 more
```

### Steps to reproduce the bug

1. Run the sample
2. Experience the error
3. Comment out the line involving `StatisticType`
4. No longer have the error

### Code sample

```java
public static void main(String[] args) {
    MinecraftServer server = MinecraftServer.init();

    // Comment this out to remove the error
    System.out.println(StatisticType.SLEEP_IN_BED);

    server.start("localhost", 25565);
}
```

Issue #2903, opened 2025-09-01
RespawnPacket fields swapped order


Blintii:
### Description
Here in RespawnPacket:
https://github.com/Minestom/Minestom/blob/a7951ca0ae34f12e8abf7915967f392c1c4f0698/src/main/java/net/minestom/server/network/packet/server/play/RespawnPacket.java#L34 the seaLevel and copyData (Data kept) fields in swapped order based on https://minecraft.wiki/w/Java_Edition_protocol/Packets#Respawn protocol description. The Sea level field comes in protocol 769 (1.21.4). Until protocol 767 (1.21.1) the RespawnPacket doesn't contain Sea level field yet. So this issue is relevant from 1.21.4+, this new field placed in wrong place
### Solution
Please fix RespawnPacket fields order to match the 1.21.4+ protocol description like this picture shows:
<img width="450" height="617" alt="Image" src="https://github.com/user-attachments/assets/b4670507-b531-409c-9a47-dbfa92536540" />
### Steps to reproducate
- Send a RespawnPacket to player (1.21.4+ joined client) with sea level above 127 (0b0111 1111)
- Client will disconnect because of crash that image shows:
<img width="1106" height="380" alt="Image" src="https://github.com/user-attachments/assets/118b772a-c70e-4870-bd55-c1a1a13e7d29" />

### Reason
Varint value starting from 128 (0b1000 0000) coded any more with 2 bytes instead of 1. But the order swapped, so client parse the varint first byte with regular byte (as Data kept named field). After that, client detects 1 more byte in the packet, but can't recognize why, so close the connection with the server and shows the network protocol error screen.

Issue #2748, opened 2025-05-20
[Bug]: Components not serializing/deserializing correctly


Jazzkuh:
### Affected Version

1.21.5 - d243be627e

### Describe the bug

Hi, does anyone know why ``ItemStack.fromItemNBT(tag)`` cuts a CompoundBinaryTag that has a lot of components like: 

```
{count:1,components:{"minecraft:item_model":"food:fruit_banana","minecraft:lore":[{color:"white",italic:0b,text:"쀁"},{:""},{text:"Nutri Score: ",color:"gray",italic:0b,extra:[{color:"white",text:"퀁"}]},{color:"gray",italic:0b,text:""},{color:"dark_gray",italic:0b,text:"Categorie: Fruit"}],"minecraft:max_stack_size":5,"minecraft:food":{saturation:0.3f,nutrition:2},"minecraft:tooltip_display":{hidden_components:["minecraft:base_color","minecraft:dyed_color","minecraft:attribute_modifiers","minecraft:unbreakable","minecraft:enchantments"]},"minecraft:custom_name":{color:"#626262",italic:0b,text:"Banaan"},"minecraft:custom_data":{nutri_score:"A",item_key:"banana"},"minecraft:tooltip_style":"mineville:common"},id:"minecraft:apple"}
```

to an item like this: 
```
ItemStackImpl[material=minecraft:apple, amount=1, components=DataComponentMapImpl[components={}]]
```

I am using this code to deserialize an item: 
```java

    public static ItemStack deserialize(String rawData) {
        try {
            CompoundBinaryTag tag = TAG_IO.asCompound(rawData);
            System.out.println(rawData);
            System.out.println(ItemStack.fromItemNBT(tag));
            return ItemStack.fromItemNBT(tag);
        } catch (IOException e) {
            throw new RuntimeException("Error while deserializing ItemStack", e);
        }
    }
```

kermandev:
Ensure you are using `MinestomAdventure.tagStringIO()` to parse, as we do not parse legacy. As legacy will emit `{:""}` instead of `""`

PR #2909, opened 2025-09-07
Resolve #2898 #2903 #2904 #2720 #2748 & more

### LINKED ISSUE #2898: ServerDataPacket not match protocol description
### Description
ServerDataPacket (packet ID 0x49) still valid until protocol version 765 (... - 1.20.5)
But starting protocol version 766 (1.20.6 - ...) it's no longer matching the protocol description.

Image

### Issue
Wrong line of code here:

https://github.com/Minestom/Minestom/blob/b7c72b9818a5c0013018f2d701eba3064ffb2250/src/main/java/net/minestom/server/network/packet/server/play/ServerDataPacket.java#L16

### Solution
Have to remove the "enforcesSecureChat" boolean field from `net/minestom/server/network/packet/server/play/ServerDataPacket`

### Steps to reproducate

- send any ServerDataPacket to player in any 1.20.6+ version
- player client will instantly crash with network protocol error screen

Image

kermandev:
## Proposed changes
Fix #2898 #2903 #2904 #2720 #2748
Also fixes data components not being key based like other registries.
Fixes nullability of MinecraftServer taken from #2908 
## Types of changes

## Further comments

### DIFF
--- a/src/main/java/net/minestom/server/MinecraftServer.java
+++ b/src/main/java/net/minestom/server/MinecraftServer.java
@@ -70,7 +70,7 @@ public final class MinecraftServer implements MinecraftConstants {
     public static final int TICK_MS = 1000 / TICK_PER_SECOND;
 
     // In-Game Manager
-    private static volatile ServerProcess serverProcess;
+    private static volatile @UnknownNullability ServerProcess serverProcess;
 
     private static int compressionThreshold = 256;
     private static String brandName = "Minestom";
--- a/src/main/java/net/minestom/server/component/DataComponent.java
+++ b/src/main/java/net/minestom/server/component/DataComponent.java
@@ -1,6 +1,7 @@
 package net.minestom.server.component;
 
 import net.kyori.adventure.key.Key;
+import net.kyori.adventure.key.KeyPattern;
 import net.minestom.server.codec.Codec;
 import net.minestom.server.item.enchant.EffectComponent;
 import net.minestom.server.network.NetworkBuffer;
@@ -57,12 +58,12 @@ public sealed interface DataComponent<T> extends StaticProtocolObject<DataCompon
     T read(NetworkBuffer reader);
     void write(NetworkBuffer writer, T value);
 
-    static @Nullable DataComponent<?> fromKey(String key) {
-        return DataComponentImpl.NAMESPACES.get(key);
+    static @Nullable DataComponent<?> fromKey(@KeyPattern String key) {
+        return fromKey(Key.key(key));
     }
 
     static @Nullable DataComponent<?> fromKey(Key key) {
-        return fromKey(key.asString());
+        return DataComponentImpl.NAMESPACES.get(key);
     }
 
     static @Nullable DataComponent<?> fromId(int id) {
--- a/src/main/java/net/minestom/server/component/DataComponentImpl.java
+++ b/src/main/java/net/minestom/server/component/DataComponentImpl.java
@@ -1,6 +1,7 @@
 package net.minestom.server.component;
 
 import net.kyori.adventure.key.Key;
+import net.kyori.adventure.key.KeyPattern;
 import net.minestom.server.codec.Codec;
 import net.minestom.server.codec.Result;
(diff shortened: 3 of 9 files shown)


DasBabyPixel:
While we are here can we talk about InstrumentComponent?
```java
public record InstrumentComponent(Either<Holder<Instrument>, RegistryKey<Instrument>> instrument)

and

interface RegistryKey<Instrument> extends Holder<Instrument>
```

why is this an Either and not just a Holder? Makes no sense to me. Could even go as far as removing the InstrumentComponent class as a result, because it's just an Instrument similar to SoundEvent

kermandev:
> While we are here can we talk about InstrumentComponent?
> 
> ```java
> public record InstrumentComponent(Either<Holder<Instrument>, RegistryKey<Instrument>> instrument)
> 
> and
> 
> interface RegistryKey<Instrument> extends Holder<Instrument>
> ```
> 
> why is this an Either and not just a Holder? Makes no sense to me. Could even go as far as removing the InstrumentComponent class as a result, because it's just an Instrument similar to SoundEvent

EitherHolders are different prefixed Holders as they with the either's T/F, so we probably could do this, just need to keep this mind. At minimum we probably should provide a convince constructor for this record, then this would be a next if removal.

I think that should be a separate PR though, as thats more feature than bugs.

DasBabyPixel:
> EitherHolders are different prefixed Holders as they with the either's T/F, so we probably could do this, just need to keep this mind.

I don't understand what you're saying here.
I'm not that deep into networking code, but from what I see on the minecraft wiki, the network type should accept either an ID (int) or a definition. Not a key. So in any case having a RegistryKey not associated in the relevant registry is an error and can't/mustn't happen. Also RegistryTag#uncheckedNetworkType and RegistryTag#uncheckedCodec is only used in InstrumentComponent. This makes me think that it is simply an oversight of something that was used once during development of the tag rework and then forgotten.
TLDR; remove and replace with Holder
@mworzala can you say why this is the way it is? Or is it a mistake/oversight?


> I think that should be a separate PR though, as thats more feature than bugs.

I kind of disagree and I'd argue it's a bug (see above). In any case it can't hurt discussing it though

EDIT: I noticed MC's source code also has an EitherHolder...
My take may be wrong here, would love to know the reason why an EitherHolder exists (I don't get it)

mworzala:
@DasBabyPixel I dont have that good of an explanation either - I have many rants in various places about the holder & friends types they are very confusing. I dont think I ever actually figured out in what scenario a resource key should end up there. On the client, if you pass a resource key it will be looked up in the registry anyway so you might as well have passed an int anyway.

However it does also serialize differently on the network, so we do need to recognize the difference somewhere at least. And notably, it is valid to send a key so I think it's reasonable for us to represent it in that type. My current theory is that it was a mistake on their end and nobody has fixed it because nobody cares about the instrument component (but thats just a theory).

There are some other fun cases if you're interested, like PaintingVariant taking a holder for its network type (meaning you can send an unregistered painting structure), but the codec (used for saving the entity notably) only supports a registry key. Fun stuff all around :)

DasBabyPixel:
> @DasBabyPixel I dont have that good of an explanation either - I have many rants in various places about the holder & friends types they are very confusing. I dont think I ever actually figured out in what scenario a resource key should end up there. On the client, if you pass a resource key it will be looked up in the registry anyway so you might as well have passed an int anyway.
> 
> However it does also serialize differently on the network, so we do need to recognize the difference somewhere at least. And notably, it is valid to send a key so I think it's reasonable for us to represent it in that type. My current theory is that it was a mistake on their end and nobody has fixed it because nobody cares about the instrument component (but thats just a theory).
> 
> There are some other fun cases if you're interested, like PaintingVariant taking a holder for its network type (meaning you can send an unregistered painting structure), but the codec (used for saving the entity notably) only supports a registry key. Fun stuff all around :)

well damn.
Time to go on the mojang bug tracker and rant about their horrible internals I or something like that 😂

mworzala (approved):

PR #2915, opened 2025-09-20
implement PlayerInstanceEvent in PlayerSpectateEvent

TropicalShadow:
## Proposed changes

Update PlayerEvent to PlayerInstanceEvent for PlayerSpectateEvent

## Types of changes

## Checklist

## Further comments

no extra comments

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerSpectateEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerSpectateEvent.java
@@ -2,13 +2,13 @@ package net.minestom.server.event.player;
 
 import net.minestom.server.entity.Entity;
 import net.minestom.server.entity.Player;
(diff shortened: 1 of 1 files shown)


kermandev:
This event doesn't make alot of sense because PlayerSpectateEvent is going to be called when the player may be leaving that instance (teleport), which begs what instance is it?

Same reason that EntityTeleportEvent probably isn't a EntityInstanceEvent (but that also is cause they could not exist before that).

TropicalShadow:
I felt that Teleport event didn't have the InstanceEvent due to it potentially not having an initial instance, but with Spectate, I feel that the instance should be the instance in which the action was performed within. Maybe some doc strings could clarify?

my other thought is to extend this event with an instance-based version, but I feel that is just adding extra weight to something that could work like this impl.

PR #2925, opened 2025-09-29
Rename IChunkLoader to ChunkLoader

kermandev:
## Proposed changes

Currently the only interface using the interface Ixxx pattern, which seems odd to have for a interface here.

## Types of changes

## Further comments

Breaking some signatures if you rely on typed IChunkLoader which is highly likely.


mworzala (commented):
I think we can just delete ichunkloader tbh, its going in with a mc version bump anyway

kermandev:
I think it might be better to still have the class until the next version bump but deprecated, so the people that don’t read the release notes don’t question why the class isn’t there  (already going to be bad enough to bump to Jdk 25)

DasBabyPixel:
@kermandev in my own personal experience, if people (me) are missing a class they are gonna go to release notes anyway. I wouldn't bother

PR #2932, opened 2025-10-04
Fix command condition bypass bug

mudkipdev:
In `GraphImpl.java`, when a command had a condition and you added a zero-argument syntax without its own condition, the syntax's null condition would overwrite the command's condition. This let anyone bypass permission checks and run admin-only commands.

I also added some utilities like `setCondition(Predicate<CommandSender>...)`, changed `setCondition(CommandCondition)` to varargs, and `Conditions#not(CommandCondition)`.

### DIFF
--- a/src/main/java/net/minestom/server/command/builder/Command.java
+++ b/src/main/java/net/minestom/server/command/builder/Command.java
@@ -38,7 +38,7 @@ import java.util.stream.Stream;
  * you can set it using {@link #setDefaultExecutor(CommandExecutor)}.
  * <p>
  * Before any syntax to be successfully executed the {@link CommandSender} needs to validated
- * the {@link CommandCondition} sets with {@link #setCondition(CommandCondition)} (ignored if null).
+ * the {@link CommandCondition} sets with {@link #setCondition(CommandCondition...)} (ignored if null).
  * <p>
  * Some {@link Argument} could also require additional condition (eg: a number which need to be between 2 values),
  * in this case, if the whole syntax is correct but not the argument condition,
--- a/src/main/java/net/minestom/server/command/builder/Command.java
+++ b/src/main/java/net/minestom/server/command/builder/Command.java
@@ -223,7 +223,6 @@ public class Command {
         return addConditionalSyntax(null, executor, args);
     }
 
-
     /**
      * Creates a syntax from a formatted string.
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/command/builder/Command.java
```diff
@@ -100,11 +102,40 @@ public CommandCondition getCondition() {
     /**
      * Sets the {@link CommandCondition}.
      *
-     * @param commandCondition the new command condition, null to do not call anything
+     * @param conditions the conditions that must all pass
      * @see #getCondition()
      */
-    public void setCondition(@Nullable CommandCondition commandCondition) {
-        this.condition = commandCondition;
+    public void setCondition(CommandCondition... conditions) {
```
kermandev: This is fine, but this is breaking, please check for Nullable
`if (conditions == null || conditions.length == 0)`

to make it not breaking

src/main/java/net/minestom/server/command/builder/condition/Conditions.java
```diff
@@ -51,4 +51,11 @@ public static boolean playerOnly(CommandSender sender, String commandString) {
     public static boolean consoleOnly(CommandSender sender, String commandString) {
         return sender instanceof ConsoleSender;
     }
+
+    /**
+     * Inverts the result of the given condition.
+     */
+    public static CommandCondition not(CommandCondition condition) {
```
kermandev, line 67: A not null check would also be nice here

src/main/java/net/minestom/server/command/builder/Command.java
```diff
@@ -98,13 +101,23 @@ public CommandCondition getCondition() {
...
+     * Sets the {@link CommandCondition}s required to execute the command.
      *
-     * @param commandCondition the new command condition, null to do not call anything
+     * @param condition the condition that must pass
+     * @param conditions the rest of the conditions that must pass
      * @see #getCondition()
      */
-    public void setCondition(@Nullable CommandCondition commandCondition) {
-        this.condition = commandCondition;
+    public void setCondition(@Nullable CommandCondition condition, CommandCondition... conditions) {
```
kermandev: Revert change to master

kermandev (changes requested):

PR #2935, opened 2025-10-05
Update the section API

TheMode:
Move `Block` storage (block entities, handler, NBT) to `Section` which could later be used to speedup stateful block check, and handle world state versioning.

The only breaking change is the new `Section` factory methods, and its change from a class to interface (ABI)

### DIFF
--- a/src/main/java/net/minestom/server/instance/DynamicChunk.java
+++ b/src/main/java/net/minestom/server/instance/DynamicChunk.java
@@ -4,6 +4,7 @@ import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
 import net.kyori.adventure.nbt.CompoundBinaryTag;
 import net.kyori.adventure.nbt.LongArrayBinaryTag;
 import net.minestom.server.MinecraftServer;
+import net.minestom.server.coordinate.BlockVec;
 import net.minestom.server.coordinate.CoordConversion;
 import net.minestom.server.coordinate.Point;
 import net.minestom.server.entity.Entity;
@@ -35,7 +36,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.*;
 
-import static net.minestom.server.coordinate.CoordConversion.globalToSectionRelative;
+import static net.minestom.server.coordinate.CoordConversion.*;
 import static net.minestom.server.network.NetworkBuffer.SHORT;
 
 /**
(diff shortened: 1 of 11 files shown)


src/main/java/net/minestom/server/instance/Section.java
```diff
@@ -1,66 +1,60 @@
...
+    Light skyLight();
 
-        return new Section(this.blockPalette.clone(), this.biomePalette.clone(), skyLight, blockLight);
-    }
+    Light blockLight();
 
-    public void setSkyLight(byte[] copyArray) {
-        this.skyLight.set(copyArray);
-    }
+    Int2ObjectOpenHashMap<@Nullable Block> entries();
```
kermandev: Dont expose fastutil please, its no longer exposable API

src/main/java/net/minestom/server/instance/SectionImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+record SectionImpl(
+        Palette blockPalette, Palette biomePalette,
+        Light skyLight, Light blockLight,
+        // Key = CoordConversion#sectionBlockIndex
+        Int2ObjectOpenHashMap<@Nullable Block> entries,
+        Int2ObjectOpenHashMap<Block> tickableMap,
+        AtomicLong version
+) implements Section {
+    public void clear() {
```
kermandev, line 23: Missing override

src/main/java/net/minestom/server/instance/SectionImpl.java
```diff
@@ -0,0 +1,60 @@
...
+        this.blockPalette.fill(0);
+        this.biomePalette.fill(0);
+        this.entries.clear();
```
kermandev, line 26: Probably should have a Section#trim aswell

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -183,15 +159,24 @@ public void loadHeightmapsFromNBT(CompoundBinaryTag heightmapsNBT) {
...
-        tickableMap.int2ObjectEntrySet().fastForEach(entry -> {
-            final int index = entry.getIntKey();
-            final Block block = entry.getValue();
-            final BlockHandler handler = block.handler();
-            if (handler == null) return;
-            final Point blockPosition = CoordConversion.chunkBlockIndexGetGlobal(index, chunkX, chunkZ);
-            handler.tick(new BlockHandler.Tick(block, instance, blockPosition));
-        });
+        int i = 0;
+        for (Section section : sections) {
```
kermandev: fori

kermandev (commented):
I dont see the main point of this change, we are going to duplicate entries across each 16x16x16 region (~24 per) (not even including the constant map overhead. To still have it so you still have to synchronize on a chunk for each section access? The only time that would be faster is generation which doesnt require synchronization cause the chunk hasnt been exposed yet. Otherwise you would already be locked on the chunk instance.

Also wouldnt it be better to just have a volatile long in Chunk for versioning too as your guarded due to the synchronization?

Also this should probably be targeted towards 1_21_9

mworzala:

mworzala:

kermandev:

PR #2937, opened 2025-10-06
Debug subscriptions fixes & default handler

kermandev:
Fixes common issue (NPE cause unloaded) and adds default listener and event and exports package.


mworzala (approved):
looks good once build fixed

PR #2944, opened 2025-10-17
Use default entity attributes from vanilla

mudkipdev:
Loads the default `LivingEntity` attributes from vanilla Minecraft for each entity type. For example, currently spawning an iron golem in Minestom would cause it to look "damaged" due to the max health not being the correct value. Requires the data generator PR to be merged.

~~I also fixed the max health attribute not being synchronized with clients in the `heal()` method.~~

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -737,5 +737,4 @@ public class LivingEntity extends Entity implements EquipmentHandler {
     public Acquirable<? extends LivingEntity> acquirable() {
         return (Acquirable<? extends LivingEntity>) super.acquirable();
     }
-
 }


kermandev:
Id prefer if the default attributes were added to the entity type registry instead.

mudkipdev:
@kermandev Requesting review

src/main/java/net/minestom/server/registry/RegistryData.java
```diff
@@ -571,6 +573,26 @@ public EntityEntry(String namespace, Properties main) {
...
+                    Attribute attribute = Attribute.fromKey(entry.getKey());
+                    if (attribute == null) continue;
+
+                    Object value = entry.getValue();
+                    if (!(value instanceof Number number)) continue;
+
+                    attributes.put(attribute, number.doubleValue());
+                }
+
+                this.defaultAttributes = attributes.isEmpty() ? Map.of() : Map.copyOf(attributes);
```
kermandev:
```suggestion
                this.defaultAttributes = Map.copyOf(attributes);
```

src/main/java/net/minestom/server/registry/RegistryData.java
```diff
@@ -571,6 +573,26 @@ public EntityEntry(String namespace, Properties main) {
...
+            Properties defaultAttributesSection = main.section("defaultAttributes");
+
+            if (defaultAttributesSection == null) {
+                this.defaultAttributes = Map.of();
+            } else {
+                Map<Attribute, Double> attributes = new HashMap<>();
+
+                for (var entry : defaultAttributesSection) {
+                    Attribute attribute = Attribute.fromKey(entry.getKey());
+                    if (attribute == null) continue;
```
kermandev:
```suggestion
                    Check.notNull(attribute, "Failed to find attribute {0}", entry.getKey());
```

src/main/java/net/minestom/server/registry/RegistryData.java
```diff
@@ -571,6 +573,26 @@ public EntityEntry(String namespace, Properties main) {
...
+
+                    Object value = entry.getValue();
+                    if (!(value instanceof Number number)) continue;
```
kermandev: Throw here like above

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -98,6 +98,13 @@ public class LivingEntity extends Entity implements EquipmentHandler {
      */
     public LivingEntity(EntityType entityType, UUID uuid) {
         super(entityType, uuid);
+        var defaultAttributes = entityType.registry().defaultAttributes();
+
+        if (!defaultAttributes.isEmpty()) {
```
kermandev: `applyDefaultAttributes` public method instead. or a better name if you can come up with one.

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -432,7 +439,7 @@ public void setHealth(float health) {
      * Retrieved from {@link #getAttributeValue(Attribute)} with the attribute {@link Attribute#MAX_HEALTH}.
      */
     public void heal() {
-        setHealth((float) getAttributeValue(Attribute.MAX_HEALTH));
+        setHealth((float) getAttribute(Attribute.MAX_HEALTH).getValue());
```
kermandev: Why was this changed?

kermandev (changes requested):

src/main/java/net/minestom/server/entity/LivingEntity.java
mudkipdev: `getAttribute` creates the attribute instance and syncs the value with the client, without this the client would just assume the attribute is at its default value which could be wrong

kermandev: But shouldnt this be updated if the default health was changed, since it still looks into the map.

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -733,4 +734,17 @@ public void takeKnockback(float strength, final double x, final double z) {
...
+
+    /**
+     * Sets the attributes of this entity to their default vanilla values.
+     */
+    public void applyDefaultAttributeValues() {
+        var defaultAttributes = entityType.registry().defaultAttributes();
+
+        if (!defaultAttributes.isEmpty()) {
+            for (var entry : defaultAttributes.entrySet()) {
+                getAttribute(entry.getKey()).setBaseValue(entry.getValue());
```
kermandev: You probably should omit setting values == defaultValue of the attribute, if they arent set.

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -733,4 +734,17 @@ public void takeKnockback(float strength, final double x, final double z) {
...
         return (Acquirable<? extends LivingEntity>) super.acquirable();
     }
+
+    /**
+     * Sets the attributes of this entity to their default vanilla values.
+     */
+    public void applyDefaultAttributeValues() {
+        var defaultAttributes = entityType.registry().defaultAttributes();
+
+        if (!defaultAttributes.isEmpty()) {
```
kermandev:
```suggestion
        if (defaultAttributes.isEmpty()) return;
```

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -462,6 +464,8 @@ public AttributeInstance getAttribute(Attribute attribute) {
      */
     protected void onAttributeChanged(AttributeInstance attributeInstance) {
         if (!shouldSendAttributes()) return;
+        // batch packets when applying default attributes
+        if (applyingDefaultAttributes) return;
```
kermandev: This check probably doesnt need to exist due to the entity likely having no viewers before the object is fully created.

kermandev (changes requested):

mworzala:

src/test/java/net/minestom/server/entity/EntityAttributeTest.java
```diff
@@ -89,4 +91,45 @@ public void testDirectlyAddAttributes(Env env) {
         player.setItemInMainHand(itemStack);
         assertEquals(0, Double.compare(player.getAttribute(Attribute.MAX_HEALTH).getValue(), baseHealth + addition));
     }
+
+    @Test
+    public void testEntityDefaultAttributes() {
```
kermandev:
```suggestion
    public void testEntityDefaultAttributes(Env ignored) {
```

src/test/java/net/minestom/server/entity/EntityAttributeTest.java
```diff
@@ -89,4 +91,45 @@ public void testDirectlyAddAttributes(Env env) {
...
+        assertEquals(golemHealth.getBaseValue(), ironGolem.getAttributeValue(Attribute.MAX_HEALTH), 0.001);
+
+        var zombieHealth = zombie.getAttribute(Attribute.MAX_HEALTH);
+        assertNotNull(zombieHealth);
+        assertEquals(20.0, zombieHealth.getBaseValue(), 0.001);
+        assertEquals(zombieHealth.getBaseValue(), zombie.getAttributeValue(Attribute.MAX_HEALTH), 0.001);
+    }
+
+    @Test
+    public void testEntitySpawnsWithCorrectHealth() {
```
kermandev:
```suggestion
    public void testEntitySpawnsWithCorrectHealth(Env ignored) {
```

src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -733,4 +734,22 @@ public void takeKnockback(float strength, final double x, final double z) {
     public Acquirable<? extends LivingEntity> acquirable() {
         return (Acquirable<? extends LivingEntity>) super.acquirable();
     }
+
+    /**
+     * Sets the attributes of this entity to their default vanilla values.
+     */
+    public void applyDefaultAttributeValues() {
```
mworzala: should be protected, someone can promote it if really needed but otherwise its internal to the entity.

mworzala: On further thought, this can just not exist. Instead the default value can be provided when constructing the AttributeInstance.

If someone (for an unknown reason to me) wants to remove this behavior its trivial to write a loop setting the attributes to their global defaults.

PR #2947, opened 2025-10-18
Fix: Use instance view distance

whereareiam:
## Proposed changes

- Joining an instance with a small view distance (e.g. 2) still advertised the global default of 8 in the initial JoinGamePacket, so vanilla clients spun up a 19×19 area (361 chunks). The server-side count exposed via Instance#getChunks().size() showed the same overshoot until the client later sent its settings packet, which contradicted the configured radius and inflated memory usage under load.
- The login handshake now uses the instance’s configured view distance (clamped to Mojang’s 2–32 bounds) when building JoinGamePacket, so the very first tick matches the world configuration.
- When a player moves to an instance with a different radius, we immediately push UpdateViewDistancePacket and UpdateSimulationDistancePacket so the client adjusts fog and simulation limits without waiting for another settings exchange.
- With these changes the chunk count retrieved from Instance#getChunks().size() stays at the expected (viewDistance + 1)^2 (e.g. 49 for view distance 2) from the moment the player spawns.

## Types of changes

## Checklist

- [x] I have read the CONTRIBUTING.md (CONTRIBUTING.md)
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~I have added necessary documentation (if appropriate)~~

## Further comments

Scope is limited to server→client view-distance negotiation; generation and persistence code paths remain unchanged. Verified manually by joining an instance set to view distance 2 and observing that the handshake and Instance#getChunks().size() both report the expected 49 chunks immediately, then switching to a larger-radius instance and seeing view distance packets update the client on the fly.

### Before
(client view distance 12; effective view distance 8; instance view distance 2)
image

### After
(client view distance 12; effective view distance 8; instance view distance 2)
image

### DIFF
--- a/src/test/java/net/minestom/server/entity/player/PlayerIntegrationTest.java
+++ b/src/test/java/net/minestom/server/entity/player/PlayerIntegrationTest.java
@@ -300,7 +300,7 @@ public class PlayerIntegrationTest {
         var player = env.createPlayer(connectingInstance, pos);
         assertNotEquals(2, ServerFlag.CHUNK_VIEW_DISTANCE); // Ensure we arent using the same value
 
-        player.refreshSettings(new ClientSettings(Locale.US, (byte) 0x2,
+        player.refreshSettings(new ClientSettings(Locale.US, (byte) 32,
                 ChatMessageType.FULL, true,
                 (byte) 0x7F, ClientSettings.MainHand.RIGHT,
                 true, true,


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2388,12 +2399,25 @@ private int compareChunkDistance(long chunkIndexA, long chunkIndexB) {
...
      * @return The effective chunk view distance range of the client
      */
     public int effectiveViewDistance() {
-        Instance instance = this.instance;
-        int maxViewDistance = instance != null ? instance.viewDistance() : ServerFlag.CHUNK_VIEW_DISTANCE;
-        return Math.min(settings.viewDistance(), maxViewDistance) + 1;
+        return computeEffectiveViewDistance(settings.viewDistance(), this.instance);
+    }
+
+    public int effectiveViewDistance(@Nullable Instance targetInstance) {
```
mworzala: this (and the next two) aren't really specific to a player, they can go in a utils file (eg ChunkUtils?). And please add a few tests for them

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
```diff
@@ -81,4 +83,43 @@ public static Chunk retrieve(Instance instance, Chunk originChunk, double x, dou
...
+     *   <li>Instance VD = 50 → 32 (clamped to maximum)</li>
+     *   <li>Instance = null → clamp(ServerFlag.CHUNK_VIEW_DISTANCE, 2, 32)</li>
+     * </ul>
+     *
+     * @param targetInstance the target instance, or null to use the default server view distance
+     * @return the server view distance, clamped between 2 and 32
+     */
+    public static int computeServerViewDistance(@Nullable Instance targetInstance) {
+        int viewDistance = targetInstance != null ? targetInstance.viewDistance() : ServerFlag.CHUNK_VIEW_DISTANCE;
+        return MathUtils.clamp(viewDistance, 2, 32);
```
kermandev, line 123: Seems weird to upperbound the viewDistance set by the user, the client accepts a varint for the settings, so there is no real reason to do this from a protocol perspective.

kermandev (commented):
Fix seems to work with added test, as it fails on master, but there should be no reason to bound server distance to 2, 32, but the regular gameserver does do this. Minestom shouldn't limit you to such a check. The only feasible check minestom should do is preventing you from going less than 2.

Similarly for future reference, there is no reason anymore for a bounds check in client settings, but I dont expect you to change this.

P.S Matt, while squashing please remove me from coauthor.

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
```diff
@@ -81,4 +83,43 @@ public static Chunk retrieve(Instance instance, Chunk originChunk, double x, dou
...
+     *   <li>Client VD = 10, Instance VD = 16 → min(10, 16) + 1 = 11</li>
+     *   <li>Client VD = 16, Instance VD = 8 → min(16, 8) + 1 = 9</li>
+     *   <li>Client VD = 12, Instance = null → min(12, ServerFlag.CHUNK_VIEW_DISTANCE) + 1</li>
+     * </ul>
+     *
+     * @param clientViewDistance the client's view distance setting
+     * @param targetInstance the target instance, or null to use the default server view distance
+     * @return the effective view distance
+     */
+    public static int computeEffectiveViewDistance(byte clientViewDistance, @Nullable Instance targetInstance) {
```
SorkoPiko, line 102: i don't think it makes much sense to use a byte as an input. i understand it's to match the byte stored in `ClientSettings`, but this method isn't related to that, this is just a helper that could be used by users. additionally, it returns an int anyway, so i don't think passing a byte just to match the `ClientSettings` is appropriate here. ~`ClientSettings.viewDistance()` can be cast to an int without any problems.~ byte is implictly widened by the compiler, so this won't even need any other changes (apart from removing the explicit byte casts elsewhere).
```suggestion
    public static int computeEffectiveViewDistance(int clientViewDistance, @Nullable Instance targetInstance) {
```

src/test/java/net/minestom/server/utils/ChunkUtilsTest.java
```diff
@@ -48,4 +57,35 @@ private static Stream<Arguments> testForDifferingChunksInRangeParams() {
     }
 
     private record ChunkCoordinate(int x, int z) {}
+
+    @Test
+    public void effectiveViewDistanceUsesServerFlagWhenNull() {
+        int result = ChunkUtils.computeEffectiveViewDistance((byte) 10, null);
```
SorkoPiko, line 63:
```suggestion
        int result = ChunkUtils.computeEffectiveViewDistance(10, null);
```
see comment on ChunkUtils.java#102

src/test/java/net/minestom/server/utils/ChunkUtilsTest.java
```diff
@@ -48,4 +57,35 @@ private static Stream<Arguments> testForDifferingChunksInRangeParams() {
...
+        assertEquals(Math.min(10, ServerFlag.CHUNK_VIEW_DISTANCE) + 1, result);
+    }
+
+    @Test
+    public void effectiveViewDistanceUsesInstanceViewDistance(Env env) {
+        Instance instance = env.createFlatInstance();
+        instance.viewDistance(8);
+        
+        assertEquals(9, ChunkUtils.computeEffectiveViewDistance((byte) 16, instance));
```
SorkoPiko, line 72:
```suggestion
        assertEquals(9, ChunkUtils.computeEffectiveViewDistance(16, instance));
```
see comment on ChunkUtils.java#102

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
SorkoPiko, line 123: as kerman said, no need for an upper bound.
```suggestion
        return Math.max(2, viewDistance);
```

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
```diff
@@ -1,9 +1,11 @@
 package net.minestom.server.utils.chunk;
 
+import net.minestom.server.ServerFlag;
 import net.minestom.server.coordinate.CoordConversion;
 import net.minestom.server.coordinate.Point;
 import net.minestom.server.instance.Chunk;
 import net.minestom.server.instance.Instance;
+import net.minestom.server.utils.MathUtils;
```
SorkoPiko, line 8:
```suggestion
```
as mentioned in `computeServerViewDistance`

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
mworzala: +1

src/main/java/net/minestom/server/utils/chunk/ChunkUtils.java
mworzala: +1

PR #2948, opened 2025-10-18
Palette optimization

AidanMars2:
## Proposed changes

`fill` has a new overload that partially fills the palette with a cuboid.

This PR optimizes several functions by using the new `PaletteIndexMap`:
- `setAll` and `replaceAll` don't immediately default to direct palette anymore.
- `optimize` now makes a new index map directly and returns early if no optimization is possible.

Fixed bugs:
- `copyFrom` now handles negative offsets properly.
- `replace(int oldValue, int newValue)` now properly replaces values if the new value already exists in the palette.
- High values are now properly accounted for in direct palettes.

Miscellaneous optimizations:
- `copyFrom` now defaults to `fill` if the source palette is single value or filled with air.
- `Palettes#remap` now only allocates if forced to.
- `load` now calls `remap` instead of manually setting all values.

Breaking changes:
- `copyFrom` now throws an exception if offset is out of bounds.
- `Palettes` is now marked as internal and some unnecessary methods were removed.
- `Palette#BLOCK_PALETTE_DIRECT_BITS` is now marked as internal.
- `Palette#offset(int)` has been removed.
- `Palette#height` now takes an `IntPredicate` instead of an `EntryPredicate`

## Types of changes

## Checklist

## Further comments

The old system of using an `Int2IntMap` and `IntArrayList` was a mess, using 5 lines of code just to initialize the palette with a single value. 
This led to several bugs in previous development and was too clunky to use locally in functions like `setAll`.
The new `PaletteIndexMap` is also more performant, as it reduces total allocations and exposes several internal functions that reduce double work.

### DIFF
--- a/src/main/java/net/minestom/server/instance/palette/Palette.java
+++ b/src/main/java/net/minestom/server/instance/palette/Palette.java
@@ -1,6 +1,7 @@
 package net.minestom.server.instance.palette;
 
 import it.unimi.dsi.fastutil.ints.Int2IntFunction;
+import net.minestom.server.MinecraftServer;
 import net.minestom.server.coordinate.Point;
 import net.minestom.server.instance.block.Block;
 import net.minestom.server.network.NetworkBuffer;
@@ -32,20 +33,26 @@ public sealed interface Palette permits PaletteImpl {
     int BLOCK_DIMENSION = 16;
     int BLOCK_PALETTE_MIN_BITS = 4;
     int BLOCK_PALETTE_MAX_BITS = 8;
+    @ApiStatus.Internal
     int BLOCK_PALETTE_DIRECT_BITS = MathUtils.bitsToRepresent(Block.statesCount() - 1);
 
     int BIOME_DIMENSION = 4;
     int BIOME_PALETTE_MIN_BITS = 1;
     int BIOME_PALETTE_MAX_BITS = 3;
+
     @ApiStatus.Internal
-    int BIOME_PALETTE_DIRECT_BITS = 7; // Vary based on biome count, this is just a sensible default
+    static int biomePaletteDirectBits() {
+        final var process = MinecraftServer.process();
+        if (process == null) return 7; // Vanilla biome direct bits as of 1.21.10
+        return MathUtils.bitsToRepresent(process.biome().size() - 1);
+    }
 
     static Palette blocks(int bitsPerEntry) {
         return sized(BLOCK_DIMENSION, BLOCK_PALETTE_MIN_BITS, BLOCK_PALETTE_MAX_BITS, BLOCK_PALETTE_DIRECT_BITS, bitsPerEntry);
     }
 
     static Palette biomes(int bitsPerEntry) {
-        return sized(BIOME_DIMENSION, BIOME_PALETTE_MIN_BITS, BIOME_PALETTE_MAX_BITS, BIOME_PALETTE_DIRECT_BITS, bitsPerEntry);
+        return sized(BIOME_DIMENSION, BIOME_PALETTE_MIN_BITS, BIOME_PALETTE_MAX_BITS, biomePaletteDirectBits(), bitsPerEntry);
     }
 
     static Palette blocks() {
@@ -53,7 +60,7 @@ public sealed interface Palette permits PaletteImpl {
     }
 
     static Palette biomes() {
-        return empty(BIOME_DIMENSION, BIOME_PALETTE_MIN_BITS, BIOME_PALETTE_MAX_BITS, BIOME_PALETTE_DIRECT_BITS);
+        return empty(BIOME_DIMENSION, BIOME_PALETTE_MIN_BITS, BIOME_PALETTE_MAX_BITS, biomePaletteDirectBits());
     }
 
     static Palette empty(int dimension, int minBitsPerEntry, int maxBitsPerEntry, int directBits) {
@@ -272,7 +279,7 @@ public sealed interface Palette permits PaletteImpl {
     NetworkBuffer.Type<Palette> BLOCK_SERIALIZER = serializer(BLOCK_DIMENSION, BLOCK_PALETTE_MIN_BITS, BLOCK_PALETTE_MAX_BITS, BLOCK_PALETTE_DIRECT_BITS);
 
     static NetworkBuffer.Type<Palette> biomeSerializer(int biomeCount) {
-        final int directBits = MathUtils.bitsToRepresent(biomeCount);
+        final int directBits = MathUtils.bitsToRepresent(biomeCount - 1);
         return serializer(BIOME_DIMENSION, BIOME_PALETTE_MIN_BITS, BIOME_PALETTE_MAX_BITS, directBits);
     }
 
--- a/src/main/java/net/minestom/server/instance/palette/Palette.java
+++ b/src/main/java/net/minestom/server/instance/palette/Palette.java
@@ -1,15 +1,13 @@
 package net.minestom.server.instance.palette;
 
-import it.unimi.dsi.fastutil.ints.Int2IntFunction;
 import net.minestom.server.MinecraftServer;
 import net.minestom.server.coordinate.Point;
 import net.minestom.server.instance.block.Block;
 import net.minestom.server.network.NetworkBuffer;
-import net.minestom.server.utils.MathUtils;
-import net.minestom.server.utils.validate.Check;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.function.IntPredicate;
 import java.util.function.IntUnaryOperator;
 
 import static net.minestom.server.network.NetworkBuffer.*;
@@ -34,7 +32,7 @@ public sealed interface Palette permits PaletteImpl {
     int BLOCK_PALETTE_MIN_BITS = 4;
     int BLOCK_PALETTE_MAX_BITS = 8;
     @ApiStatus.Internal
-    int BLOCK_PALETTE_DIRECT_BITS = MathUtils.bitsToRepresent(Block.statesCount() - 1);
+    int BLOCK_PALETTE_DIRECT_BITS = Palettes.paletteBits(Block.statesCount());
 
     int BIOME_DIMENSION = 4;
     int BIOME_PALETTE_MIN_BITS = 1;
@@ -43,8 +41,14 @@ public sealed interface Palette permits PaletteImpl {
     @ApiStatus.Internal
     static int biomePaletteDirectBits() {
         final var process = MinecraftServer.process();
-        if (process == null) return 7; // Vanilla biome direct bits as of 1.21.10
-        return MathUtils.bitsToRepresent(process.biome().size() - 1);
+        // 7 == Vanilla biome direct bits as of 1.21.10
+        final int biomeDirectBits = process != null ? Palettes.paletteBits(process.biome().size()) : 7;
+        if (BiomePaletteHolder.BIOME_DIRECT_BITS == 0) {
+            BiomePaletteHolder.BIOME_DIRECT_BITS = biomeDirectBits;
+        } else if (biomeDirectBits != BiomePaletteHolder.BIOME_DIRECT_BITS) {
+            throw new IllegalStateException("Biome direct bits must not change after chunk creation");
+        }
+        return biomeDirectBits;
     }
 
     static Palette blocks(int bitsPerEntry) {
@@ -63,10 +67,12 @@ public sealed interface Palette permits PaletteImpl {
         return empty(BIOME_DIMENSION, BIOME_PALETTE_MIN_BITS, BIOME_PALETTE_MAX_BITS, biomePaletteDirectBits());
     }
 
+    @ApiStatus.Internal
     static Palette empty(int dimension, int minBitsPerEntry, int maxBitsPerEntry, int directBits) {
         return new PaletteImpl((byte) dimension, (byte) minBitsPerEntry, (byte) maxBitsPerEntry, (byte) directBits);
     }
 
+    @ApiStatus.Internal
     static Palette sized(int dimension, int minBitsPerEntry, int maxBitsPerEntry, int directBits, int bitsPerEntry) {
         return new PaletteImpl((byte) dimension, (byte) minBitsPerEntry, (byte) maxBitsPerEntry, (byte) directBits, (byte) bitsPerEntry);
     }
@@ -110,11 +116,6 @@ public sealed interface Palette permits PaletteImpl {
     default void fill(Point min, Point max, int value) {
         fill(min.blockX(), min.blockY(), min.blockZ(), max.blockX(), max.blockY(), max.blockZ(), value);
     }
-
-    /**
-     * Efficiently offsets all values in the palette by the given offset.
-     */
-    void offset(int offset);
 
     /**
      * Efficiently copies values from another palette with the given offset.
@@ -182,14 +183,12 @@ public sealed interface Palette permits PaletteImpl {
 
     /**
      * Gets the highest y position where predicate returns true, or -1 if none matched.
-     * <p>
-     * It is assumed that the predicate result is independent of passed coordinates.
      *
      * @param x the x coordinate to check at
      * @param z the z coordinate to check at
      * @return the highest y position where predicate returns true, or -1 if none matched.
      */
-    int height(int x, int z, EntryPredicate predicate);
+    int height(int x, int z, IntPredicate predicate);
 
     /**
      * Returns the number of bits used per entry.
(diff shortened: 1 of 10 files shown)


src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -30,13 +31,13 @@ public sealed interface Palette permits PaletteImpl {
...
     int BLOCK_PALETTE_MAX_BITS = 8;
-    int BLOCK_PALETTE_DIRECT_BITS = 15;
+    int BLOCK_PALETTE_DIRECT_BITS = MathUtils.bitsToRepresent(Block.statesCount() - 1);
 
     int BIOME_DIMENSION = 4;
     int BIOME_PALETTE_MIN_BITS = 1;
     int BIOME_PALETTE_MAX_BITS = 3;
     @ApiStatus.Internal
-    int BIOME_PALETTE_DIRECT_BITS = 6; // Vary based on biome count, this is just a sensible default
+    int BIOME_PALETTE_DIRECT_BITS = 7; // Vary based on biome count, this is just a sensible default
```
kermandev: I know these are dynamic, but why dont we set the initial value during startup? Ideally these could be set as late as possible in a class holder idiom.

Likely a biomeDirectBits() static method, with a blockDirectBits() too possibly.

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -17,16 +14,15 @@
 
 final class PaletteImpl implements Palette {
     private static final ThreadLocal<int[]> WRITE_CACHE = ThreadLocal.withInitial(() -> new int[SECTION_BLOCK_COUNT]);
-    final byte dimension, minBitsPerEntry, maxBitsPerEntry, directBits;
+    final byte dimension, minBitsPerEntry, maxBitsPerEntry;
+    byte directBits;
```
kermandev: Any major reason why you just dont throw on mismatch for this and keep it final?

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -191,61 +124,38 @@ public void replace(int oldValue, int newValue) {
...
-                valueToPaletteMap.remove(oldValue);
-                valueToPaletteMap.put(newValue, index);
-                // Update count
-                if (newValue == 0) {
-                    this.count -= count; // Replacing with air
-                } else if (oldValue == 0) {
-                    this.count += count; // Replacing air with a block
-                }
+            int oldIndex;
+            int newIndex;
```
kermandev: final keyword here please (for both)

src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -68,24 +69,35 @@ static Palette sized(int dimension, int minBitsPerEntry, int maxBitsPerEntry, in
...
+     * Efficiently fills a cuboid from {@code (x0, y0, z0)} to {@code (x1, y1, z1)} (inclusive).
+     * <p>
+     * If any part of the cuboid lies outside the bounds of palette,
+     * those out-of-bounds positions are ignored.
+     * <p>
+     * Coordinate order does not matter.
+     *
+     * @param value The value to fill with
+     */
+    void fill(int x0, int y0, int z0, int x1, int y1, int z1, int value);
```
kermandev: add fill(Point, Point, int)

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -260,20 +170,118 @@ public void replace(int x, int y, int z, IntUnaryOperator operator) {
...
+    public void fill(int value) {
+        this.bitsPerEntry = 0;
+        this.count = value;
+        this.values = null;
+        this.paletteIndexMap = null;
+    }
+
+    @Override
+    public void fill(int x0, int y0, int z0, int x1, int y1, int z1, int value) {
+        final int dimensionMinus = dimension - 1;
```
kermandev, line 222: We should not do bounds clamping here, instead we should throw on out of bounds.

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -285,65 +293,31 @@ public void copyFrom(Palette source, int offsetX, int offsetY, int offsetZ) {
...
+        // Calculate the actual copy bounds - only copy what fits within target bounds
+        final int minX = Math.max(0, -offsetX);
+        final int minY = Math.max(0, -offsetY);
+        final int minZ = Math.max(0, -offsetZ);
+        final int maxX = dimension - Math.max(0, offsetX);
+        final int maxY = dimension - Math.max(0, offsetY);
+        final int maxZ = dimension - Math.max(0, offsetZ);
+
+        // Early exit if nothing to copy (offset pushes everything out of bounds)
+        if (minX >= maxX || minY >= maxY || minZ >= maxZ) {
```
kermandev: throw instead (I know its breaking, but it doesnt make sense it turn into a noop for out of bounds)

src/main/java/net/minestom/server/instance/palette/PaletteIndexMap.java
```diff
@@ -0,0 +1,204 @@
...
+package net.minestom.server.instance.palette;
+
+import it.unimi.dsi.fastutil.HashCommon;
+import net.minestom.server.utils.MathUtils;
+
+/**
+ * A performant implementation for transforming values into palette indices.
+ * Derived from {@link it.unimi.dsi.fastutil.ints.Int2IntOpenHashMap} and modified to suit our specific needs.
+ */
+@SuppressWarnings({"StatementWithEmptyBody", "UnusedReturnValue"})
```
kermandev: remove first suppressed

src/main/java/net/minestom/server/instance/palette/PaletteIndexMap.java
```diff
@@ -0,0 +1,204 @@
...
+        PaletteIndexMap c;
+        try {
+            c = (PaletteIndexMap) super.clone();
+        } catch (CloneNotSupportedException cantHappen) {
+            throw new InternalError();
+        }
+        c.containsNullKey = containsNullKey;
+        c.value = value.clone();
+        c.index = index.clone();
+        c.indexToValue = indexToValue.clone();
```
kermandev, line 198: missing cloning the rest

src/main/java/net/minestom/server/instance/palette/PaletteIndexMap.java
```diff
@@ -0,0 +1,204 @@
...
+    private int realSize() {
+        return containsNullKey ? size - 1 : size;
+    }
+
+    @Override
+    public PaletteIndexMap clone() {
+        PaletteIndexMap c;
+        try {
+            c = (PaletteIndexMap) super.clone();
+        } catch (CloneNotSupportedException cantHappen) {
```
kermandev:
```suggestion
        } catch (CloneNotSupportedException _) {
```

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -285,65 +293,31 @@ public void copyFrom(Palette source, int offsetX, int offsetY, int offsetZ) {
...
-                        if (wasAir != isAir) {
-                            this.count += wasAir ? 1 : -1;
-                        }
-                    }
-                }
-            }
+        if (sourcePalette.bitsPerEntry == 0 || sourcePalette.count == 0) {
+            final int dimensionMinus = dimension - 1;
+            fill(offsetX, offsetY, offsetZ,
+                    dimensionMinus + offsetX, dimensionMinus + offsetY, dimensionMinus + offsetZ,
```
kermandev: Wouldnt this always throw if its not zero now, Maybe this shouldnt take account the offsetX/Y/Z for the upper bound

src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -109,8 +121,10 @@ static Palette sized(int dimension, int minBitsPerEntry, int maxBitsPerEntry, in
      */
     void copyFrom(Palette source);
 
+    void load(int[] palette, long[] values);
```
kermandev, line 157: javadoc please

src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -68,24 +69,35 @@ static Palette sized(int dimension, int minBitsPerEntry, int maxBitsPerEntry, in
...
 
     void replaceAll(EntryFunction function);
 
+    void fill(int value);
+
+    /**
+     * Efficiently fills a cuboid from {@code (x0, y0, z0)} to {@code (x1, y1, z1)} (inclusive).
+     * <p>
+     * If any part of the cuboid lies outside the bounds of palette,
+     * those out-of-bounds positions are ignored.
```
kermandev: throws on any out of bounds positions.

Imagine Palette is an array, if you try do something outside of the bounds of the array, the logic of your program is wrong, especially since access to chunk palettes are synchronized, and if you can avoid accessing them for nops you should.

src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -231,19 +256,24 @@ static NetworkBuffer.Type<Palette> serializer(int dimension, int minIndirect, in
...
             public void write(NetworkBuffer buffer, Palette palette) {
                 PaletteImpl value = (PaletteImpl) palette;
-                // Temporary fix for biome direct bits depending on the number of registered biomes
-                if (directBits != value.directBits && !value.hasPalette()) {
-                    PaletteImpl tmp = new PaletteImpl((byte) dimension, (byte) minIndirect, (byte) maxIndirect, (byte) directBits);
-                    tmp.setAll(value::get);
-                    value = tmp;
+                Check.argCondition(dimension != value.dimension,
+                        "Palette must be of dimension {0}, got {1}", dimension, value.dimension);
+                if (directBits != value.directBits && value.isDirect()) {
```
kermandev: Would still prefer the workaround, as it can be alleviated with more general registry initialization changes, Unlikely this is going to be a bottleneck for your server especially if the default bits turns into a class holder idiom which should be lazy enough.

src/main/java/net/minestom/server/instance/palette/Palette.java
```diff
@@ -30,13 +31,13 @@ public sealed interface Palette permits PaletteImpl {
     int BLOCK_DIMENSION = 16;
     int BLOCK_PALETTE_MIN_BITS = 4;
     int BLOCK_PALETTE_MAX_BITS = 8;
-    int BLOCK_PALETTE_DIRECT_BITS = 15;
+    int BLOCK_PALETTE_DIRECT_BITS = MathUtils.bitsToRepresent(Block.statesCount() - 1);
```
kermandev: Could you mark this as internal, as now its dependent on registry initialization?

kermandev (commented):
Thanks for the PR, I do think it does clean it up a bit, but do you have any benchmarks against master?

src/main/java/net/minestom/server/instance/palette/Palette.java
AidanMars2: This would be a good idea, and I tried it, but unfortunately we don't have the biome registry available during testing (it is part of the server process and that is null during the tests as far as I know)

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
AidanMars2: The registries are only frozen after the `MinecraftServer#start()` has been called. 

Before that happens, an instance can be created, which creates biome palettes.
Then you can still register new biome types, potentially changing the direct bits, which pre-existing palettes don't know about.

src/main/java/net/minestom/server/instance/palette/Palette.java
kermandev: You just need to add EnvTest to the accompanying tests with a Env paramater, but realistically sections or chunks should be passing down the direct sizes they require

src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
kermandev: Creating an instance wouldn't initialize palettes, creating chunks would, but I understand your point.

AidanMars2:
> Thanks for the PR, I do think it does clean it up a bit, but do you have any benchmarks against master?

I benchmarked against master and got the following:
`get` and `set` were not significantly affected.
The new `fill` overload is 4x faster than the naive implementation.
`getAll` and `getAllPresent` are now 1.5-2x faster.
`optimize` is now 25% faster.

`setAll` and `replaceAll` are 3x slower than master, using 3-4x less memory and network in exchange.

kermandev:
> > Thanks for the PR, I do think it does clean it up a bit, but do you have any benchmarks against master?
> 
> I benchmarked against master and got the following: `get` and `set` were not significantly affected. The new `fill` overload is 4x faster than the naive implementation. `getAll` and `getAllPresent` are now 1.5-2x faster. `optimize` is now 25% faster.
> 
> `setAll` and `replaceAll` are 3x slower than master, using 3-4x less memory and network in exchange.

Do you have the JMH benchmarks you were using for these tests or are you using the server to test them?

AidanMars2:
https://github.com/AidanMars2/palette-test

note: the claim of less memory and network is based on my understanding of the palette structure

kermandev:
So Ive reread this, i realized it makes no sense that directBits needs to be mutable or is changing at all, its defined as the upper bound so BPE <= directBits, assume its like maxDirectBits.
I think its actually a problem that was overlooked by the current implementations, so if any value is greater the possible max value in maxDirectBits it should fail. Also I think the current tests fail this cause they attempt filling negative values, or values to big to fit into maxDirectBits ever.

Also biomeBits shouldnt change if REGISTRY_UNSAFE_OPS is disabled after startup, but yes there is a window between when biomes are added and possibly instances being created with chunks, so I think the original workaround is more correct, with a possible warning that its having to resize up. (As now it should throw if the value is too big for the max direct bits, which can be diagnosed easier from an initialization order.

Also If either is not a single value:
Optimizing for speed should make BPE = directBits
Optimizing for size should try to shrink BPE as much as possible, even in direct mode.

Also sidenote: Technically since we control the PaletteIndexMap, it KV/I could be optimized in the same packing for space knowing the bounds of key/values. I wouldnt worry about that until a future PR though.

AidanMars2:
> So Ive reread this, i realized it makes no sense that directBits needs to be mutable or is changing at all, its defined as the upper bound so BPE <= directBits, assume its like maxDirectBits.

This assumption is incorrect, as per the [minecraft wiki](https://minecraft.wiki/w/Java_Edition_protocol/Chunk_format#Paletted_Container_structure).

> Values not listed in the following table are rounded upwards to the next one specified, or downwards if larger than the value for Direct. Therefore such values will lead to unexpected results, and should not be used.

So the following needs to hold true exactly `bpe == 0 || (bpe >= minBits && bpe <= maxBits) || bpe == directBits` to be able to serialize without resizing.

kermandev:
> > So Ive reread this, i realized it makes no sense that directBits needs to be mutable or is changing at all, its defined as the upper bound so BPE <= directBits, assume its like maxDirectBits.
> 
> This assumption is incorrect, as per the [minecraft wiki](https://minecraft.wiki/w/Java_Edition_protocol/Chunk_format#Paletted_Container_structure).
> 
> > Values not listed in the following table are rounded upwards to the next one specified, or downwards if larger than the value for Direct. Therefore such values will lead to unexpected results, and should not be used.
> 
> So the following needs to hold true exactly `bpe == 0 || (bpe >= minBits && bpe <= maxBits) || bpe == directBits` to be able to serialize without resizing.

Yes protocol wise it is required directBits are required to be equal from both sides, but form a compactness its possible to compress them smaller than that until writing.

Still we do not need to support directBits resizing, instead only bpe should change, and any value that is greater than the maximum bpe defined in directBits cannot be stored.

AidanMars2:
So I benchmarked the dynamic direct bits and found the following:
`get` and `set` are not measurably affected when not resizing.
1632–3504 bytes can be saved when using lower direct bits (out of 8192 bytes by default).
Serializing becomes 30x slower according to [this benchmark](https://github.com/AidanMars2/palette-test/blob/d2932b3f93b472b6f789c59f993362c145cc106c/src/main/java/com/aidanmars/bench/WriteValuesBench.java), where the values are funneled into a `BlackHole`.

If you decide you want the memory savings in exchange for the performance hit in serialization, that's fine with me. Just let me know.

Side note: we currently cache full chunk packets in a way where all palettes in the chunk are essentially stored twice in memory. I think it's a better optimization to serialize the palettes on the fly.

kermandev:
I think it might be an idea to be more conformant to the protocol with the later option to let users #optimize(SIZE) that performs that direct size optimization. Also with the chunk packet cache, its actually quite important to have this, and users can disable these anyways behind server flags. (Not really documented though!)

AidanMars2:
I tend to agree, but I do want to point out that if someone wants to `#optimze(SIZE)` to improve network performance, it can reduce serialization performance. It might be worth adding `#optimize(NETWORK)` to get around this.

kermandev:
Superseeded by #3324, which does some more changes to palettes. Apologies for not getting this in after a lot of back and forth review.

Some changes can likely be reintroduced after it lands. (Likely around 26.3 minestom)

PR #2949, opened 2025-10-18
Add `BlockEntityType`, fix ENTITY_DATA/BLOCK_ENTITY_DATA comps

mworzala:
This is a breaking change, but its a bug fix (missed update) so acceptable.

There is a secondary breaking change to be done in the future of changing BlockEntityDataPacket to use a BlockEntityType, but I will wait for a future update to do that one.

Also turns Codec.COMPOUND_BINARY_TAG into a StructCodec and properly serializes EntityType as its key instead of its int id.


mworzala:
Ended up doing the swap in packet and block registry, preserving the old methods so its non-breaking

PR #2950, opened 2025-10-18
Fix: PlayerChunkUnloadEvent race condition for async chunk loading

whereareiam:
## Proposed changes

- Added a pending unload queue in `InstanceContainer` that stores player UUIDs when an unload is triggered for a non-existent chunk. When that chunk finishes loading, the queued unloads are processed and events fire with the now-available chunk.
- Modified `Player.chunkRemover` to call `instance.handlePlayerChunkUnload()` instead of firing the event directly, routing through the new queue system that checks if chunks exist before firing events.

## Types of changes

## Checklist

### Background

I use `PlayerChunkUnloadEvent` to implement custom chunk management in my server - specifically, unloading chunks when no players are viewing them to prevent unnecessary memory usage. This is a common pattern for servers that want fine-grained control over chunk lifecycle.

However, I noticed that when players flew around quickly, the loaded chunk count (via `Instance#getChunks().size()`) would continuously grow instead of remaining stable. Upon investigation, I discovered a race condition: when a player moves away from a chunk that's still loading asynchronously, `PlayerChunkUnloadEvent` fires immediately with just chunk coordinates. At that moment, the chunk doesn't exist in memory yet, so my event listener can't process it - there's no chunk to check viewers on or to unload. 

When the chunk eventually finishes loading, no unload event fires again. The chunk just sits in memory forever, orphaned. This makes `PlayerChunkUnloadEvent` unreliable for its intended use case of chunk lifecycle management.

### Solution

This fix introduces a pending unload queue that bridges the timing gap. When an unload is triggered for a chunk that doesn't exist yet, I queue the player's UUID. When that chunk loads, I process the pending unloads and fire events with the now-available chunk. This ensures the event fires reliably for all chunks, making it actually usable for chunk management.

### Before
image

### After
image

### DIFF
--- a/src/main/java/net/minestom/server/instance/InstanceContainer.java
+++ b/src/main/java/net/minestom/server/instance/InstanceContainer.java
@@ -1,6 +1,7 @@
 package net.minestom.server.instance;
 
 import it.unimi.dsi.fastutil.ints.Int2ObjectMaps;
+import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import net.kyori.adventure.key.Key;
 import net.kyori.adventure.nbt.CompoundBinaryTag;
 import net.minestom.server.MinecraftServer;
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/instance/InstanceContainer.java
```diff
@@ -77,6 +79,7 @@ public class InstanceContainer extends Instance {
     // used as a monitor when access is required
     private final Long2ObjectSyncMap<Chunk> chunks = Long2ObjectSyncMap.hashmap();
     private final Map<Long, CompletableFuture<Chunk>> loadingChunks = new ConcurrentHashMap<>();
+    private final Long2ObjectOpenHashMap<List<UUID>> pendingChunkUnloads = new Long2ObjectOpenHashMap<>();
```
kermandev, line 82: Shouldnt the loadingChunks be canceled instead?

PR #2954, opened 2025-10-25
fix: change Armadillo class to extend AgeableMob to fix metadata indices disconnecting clients

Bloeckchengrafik:
## Proposed changes

What it says on the tin. This fixes armadillo metadata so it correctly inherits from AgeableMob as seen in the current client decomp: `Armadillo extends Animal` and `Animal extends AgeableMob`. Also looked at some other definitions connected to AgeableMob, those seem all fine.

## Types of changes

## Checklist

merging your code._

## Further comments

I wonder if it would be possible to have an automated way of checking these things. I thought about it for a second but I did not have a good idea.


kermandev (approved):
Thanks

PR #2956, opened 2025-10-28
Add cloud_height

thecodertommy:
## Proposed changes

Add the `cloud_height` field to DimensionType.
Currently, Minestom always sends null for this field, meaning clouds never appear.
This allows clouds to optionally appear in the world as they do in vanilla.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/demo/src/main/java/net/minestom/demo/PlayerInit.java
+++ b/demo/src/main/java/net/minestom/demo/PlayerInit.java
@@ -211,6 +211,12 @@ public class PlayerInit {
                     )));
                 }
             })
+            .addListener(PlayerGameModeRequestEvent.class, event -> {
+                final Player player = event.getPlayer();
+                if (player.getPermissionLevel() >= 2) {
+                    player.setGameMode(event.getRequestedGameMode());
+                }
+            })
             .addListener(PlayerChatEvent.class, event -> {
                 var dialog = new Dialog.MultiAction(
                         new DialogMetadata(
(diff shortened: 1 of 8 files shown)


src/main/java/net/minestom/server/world/DimensionType.java
```diff
@@ -119,12 +122,13 @@ final class Builder {
...
         private boolean hasCeiling = false;
         private float ambientLight = 0f;
-        private Long fixedTime = null;
+        private @Nullable Long fixedTime = null;
         private boolean piglinSafe = false;
         private boolean bedWorks = true;
         private boolean respawnAnchorWorks = false;
         private boolean hasRaids = true;
         private int logicalHeight = VANILLA_MAX_Y - VANILLA_MIN_Y + 1;
+        private @Nullable Integer cloudHeight = 192;
```
kermandev: this shouldnt have a default height

src/main/java/net/minestom/server/world/DimensionType.java
```diff
@@ -45,11 +46,11 @@ public sealed interface DimensionType extends DimensionTypes permits DimensionTy
     static DimensionType create(
             boolean ultrawarm, boolean natural, double coordinateScale, boolean hasSkylight, boolean hasCeiling,
```
kermandev, line 47: Could you duplicate the old one and mark it for removal and set the default value to null

kermandev (changes requested):
Thanks, but a few changes

kermandev (approved):

PR #2677, opened 2025-03-09
NetworkBuffer FFM

KermanIsPretty:
## Proposed changes

Changes network buffers implementation to use an FFM (MemorySegment) backing.

## Types of changes

What types of changes does your code introduce to this project?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of
them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before
merging your code._

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments
This is a draft. We need to see where confined network buffers can be used.

PR #2958, opened 2025-10-29
Network Buffer FFM & Network Conformance

kermandev:
## Proposed changes
Tracking PR for FFM, being a squash of all existing FFM work; See #2677 for more history on changes.
Description will be updated with changes and reasoning.

# FFM Changelog

## NetworkBuffer
* `NetworkBuffer` is unsealed and documented.
* **Type Updates**:
  * `OPTIONAL_NBT` and `OPTIONAL_NBT_COMPOUND` types to cleanly handle optional NBT data without arbitrary branching.
  * `LP_ANGLE` maps 0-360 degrees to a single unsigned byte.
  * `FixedBitSet` and `EnumSet` write proper fixed bitsets.
  * `Type#unionType` uses a single generic bound.
  * `BLOCK_POSITION` now returns `BlockVec` instead of `Vec`.
  * `Recursive`: Added `Recursive` similar to `Codec#Recursive`
* **Management**:
  * `#requestCapacity(long)` added for auto resizing.
  * `#ensureReadable(long)` fails if there aren't enough bytes to read.
  * `#trim()` and `#trimmed(...)` added to compact and shrink resizable buffers.
  * `#slice(...)` returns a new instance backed by a slice of the parent buffer.
  * `#copy(NetworkBufferAllocator, ...)` overloads copy for a specific allocator.
  * `#readOnly()` returns a read only *view*. If the segment is resizeable, when any operation is attempted, the read only view is not required to be the same.
  * `#fill(...)` added along with zeroing documentation.
* **Deprecations & Removals**:
  * `#extractBytes()` deprecated for `#extractReadBytes(...)` or `#extractWrittenBytes(...)`.
  * `#copyTo(long, byte[], long, long)` deprecated.
  * `#equals` deprecated for `#contentEquals`.
* **Usage**:
  * `staticBuffer(..., Registries)` and `resizableBuffer(..., Registries)` now require `Registries` to be non null.
  * `makeArray(Type, T)` has been added as a shortcut. Existing `makeArray(..., Registries)` still accepts a nullable `Registries` to easily allow passing through registries from another `NetworkBuffer` without manual null checks. Note that `makeArray` now uses a confined arena, meaning the returned buffer is strictly bound to the thread that initially created it and will throw an exception if accessed by other threads.
  * `wrap(MemorySegment, ...)` was added as a new experimental feature.
  * `dummy(Registries)` was introduced to return a dummy buffer (useful for `Type#sizeOf`) and uses a nullable `Registries`.

## NetworkBuffer Additions
* **Advanced Memory Allocation API**:
  * `NetworkBufferAllocator`: New functional interface defining how `NetworkBuffer` objects should be allocated for network buffers. Supports explicit `Arena` assignments (`staticAllocator()` or `resizeableAllocator()`).
  * `NetworkBufferSegmentProvider`: Not really intended for general API use yet, but allows advanced users a way to access the backing `MemorySegment` of a buffer.
  * **Native Memory**: Uses native memory allocation (`malloc` and `free`) bypassing standard Java memory clearing when `ServerFlag.ATTEMPT_NATIVE_ALLOCATION` is true.
* **Views**:
  * `NetworkBuffer.IOView`: Extends `DataInput` and `DataOutput` to mostly interface with Java's streams, used with Adventure.
  * `NetworkBuffer.Direct`: Used internally denoting that interactions write directly to the buffer, ignoring indiences rules.

## Reading & Writing
* **Slices**: Packet reading uses read only slices to stop readers from messing with the buffer.
* **Error Handling**: Restructures how invalid packets are handled and logged using server flags:
  * **Packet Misuse**: If a packet handler throws an exception during execution, it triggers a disconnect (`REJECT_MISUSED_PACKET`).
  * **Packet Malformed**: If a packet constructor or any point during reading throws an exception (invalid data), it triggers a disconnect (`REJECT_MALFORMED_PACKET`).
  * **Error Suppression**: Added `SUPPRESS_PACKET_ERROR_LEVEL` to hide packet reading exceptions based on the `ConnectionState` ordinal. This is extremely useful for preventing console log spam from automated server scanners and pingers. Handled securely inside `PlayerSocketConnection` and `ExceptionManager`.
* **Connection Resets**: The server writer ignores "Connection Reset by peer" errors.
* **Disconnection Leak**: Fixed the leak of `PlayerSocketConnection`s when a connection is closed using the new write lock.

## Memory & Caching
* **BufferedPacket**: Sets it's buffer to a read only view in the canonical constructor.
* **CachedPacket** & **LazyPacket**: Uses weaker acquire/release semantics to be better on ARM platforms.
* **LazyPacket**: Now uses synchronized in the slow path, compared to racing till the last winner, as virtual thread pinning has been fixed in JDK 25.
* **FramedPacket**: Its `readIndex` gets set after going read only in the constructor.
* **SendablePacket**: `#extractServerPacket` requires a writer for cached packets now, and may require different fields in the future.

## Packets
* **Action based**: `StopSoundPacket` and `ScoreboardObjectivePacket` are now action based (e.g., `new StopSoundPacket(new StopSoundPacket.All())` instead of `new StopSoundPacket(0, null, null)`).
* **PlayerInfoUpdatePacket**: `PlayerInfoUpdatePacket.Property` removed in favor of `GameProfile.Property`.
* **DamageEventPacket**: Updated to use the registry with nullable attacker and source entity IDs.
* **FacePlayerPacket**: Entity data now uses a nullable `EntityData` instead of special values (e.g `new FacePlayerPacket(FacePosition.EYES, Vec.ZERO, null)` instead of `new FacePlayerPacket(FacePosition.EYES, Vec.ZERO, 0, null)`).
* **StatusRequestPacket**: Renamed to `ClientStatusRequestPacket`.
* **LegacyServerListPingPacket**: `LegacyServerListPingPacket`, `ServerListPingType`, and related methods are gone. These packets have not been handled for quite a while now.
* **ChunkData**: Chunk data is now formed using `BlockEntityInfo`.
* **MagicConstants**: Added flag support to `ClientPlayerPositionAndRotationPacket`, `ClientPlayerPositionPacket`, `ClientPlayerPositionStatusPacket`, `ClientPlayerRotationPacket`, `ClientUpdateStructureBlockPacket`, `RespawnPacket`, and `DeclareCommandsPacket.Node`.
* **ComponentHolding**: Added to `ServerLinksPacket` (which now uses `Either` for Entry), `MapDataPacket`, `PlaceGhostRecipePacket`, and `SetTitleTextPacket` / `SetTitleSubTitlePacket`.
* **ParticlePacket**: Origins and offsets now use `Point`. The old constructors are still available for compatibility but are now deprecated.
* **SoundEffectPacket**: Translates values locally wrapping divisions implicitly inside protocol calls.
* **Templates**: Packets that used to have a manual type are now using templates.
* **Immutability**: Packets should no longer be compared by identity and are generally more immutable overall (e.g., clone applied on byte arrays in `ClientPluginMessagePacket`, `ClientEncryptionResponsePacket`, `ClientLoginPluginResponsePacket`, `MultiBlockChangePacket`, `RegistryDataPacket`, etc).

## New Server Flags
* `SUPPRESS_CONNECTION_ACCEPT_ERRORS`: Suppresses traces caught explicitly during the server's main socket accept phase when a client first connects (default: `true`). Useful to suppress scanners.
* `REJECT_MALFORMED_PACKET`: Drops connections if a packet constructor or reading parsing throws an exception (default: `true`).
* `REJECT_MISUSED_PACKET`: Drops connections if a packet handler throws an exception during execution (default: `true`). This may be unwanted from current behavior if you want players to still be connected if their packet was mishandled.
* `SUPPRESS_PACKET_ERROR_LEVEL`: Suppresses packet reading errors thresholded by >= `ConnectionState` (default: `-1`, no supression).
* `WARN_UNREAD_BYTES_PACKET`: Triggers a console warning if a packet is read but unread bytes remain in the payload (default: `true`).
* `AUTOMATIC_COMPONENT_TRANSLATION`: Automatically attempts to translate components server side before sending (default: `false`). This replaced `MinstomAdventure.AUTOMATIC_COMPONENT_TRANSLATION`.
* `DEFAULT_RESIZEABLE_SIZE`: The baseline size of newly allocated resizable NetworkBuffers (default: `256`).
* `ATTEMPT_NATIVE_ALLOCATION` / `FORCE_NATIVE_ALLOCATION`: Used to bypass standard Arena `calloc` zeroing by directly calling `malloc` and `free` using native memory segments for better performance (defaults: `true` / `false`). Currently only supports 64 bit machines. 
* `DECOMPRESS_POOL_SIZE` / `COMPRESS_POOL_SIZE` / `PACKET_POOL_SIZE` / `VIEWABLE_POOL_SIZE`: Allows configuring the size of various internal memory pools that were previously unconfigurable (default: `32767`).

## Miscellaneous
* **Damage API**: `Damage#getTypeId()` is deprecated in favor of `#getType()`.
* **Codec BlockVec**: `BLOCK_POSITION` Codec now returns `BlockVec` to align with NetworkBuffer's.
* **ComponentSerialization**: Transitioned from `NetworkBufferTypeImpl` interface to returning instances of `NetworkBuffer.Type<Component>`. Lookups for registry payloads use the buffer's `ioView()`.
* **VarInt & VarLong**: Now can be used when the buffer is under 5/10 bytes.
* **VarInt**: Packets using small `VarInt`s (0-127) now serialize using `Byte` instead. This is purely unnoticeable internally and no types should change.
* **MessageSignature**: Fixed to strictly 256 bytes (no longer uses `RAW_BYTES`). `MessageSignature.Packed` network type is now properly implemented.
* **ObjectPool**: Split implementation (still internal), supports `Unpooled`, and moved to the `collections` package.
* **Binary Tags**: `BinaryTagReader` & `BinaryTagWriter` are now records around `DataInput`/`DataOutput`.
* **Packet Listeners**: Registrations in `PacketListenerManager` renamed from `StatusRequestPacket` to `ClientStatusRequestPacket` and updated consumers (`PacketPlayListenerConsumer`, `PacketPrePlayListenerConsumer`) to throw Exception out of `.accept()`.
* **JMH Benchmarks**: Added multiple network buffer, memory allocation, and packet benchmarks.
* **Module Exports**: `module-info` exports `net.minestom.server.network.foreign`.

## Types of changes

## Publishing
These changes are published so you can use them by accessing the snapshots repository.
```
repositories {
    maven(url = "https://central.sonatype.com/repository/maven-snapshots/") {
        content {
            includeModule("net.minestom", "minestom")
        }
    }
    mavenCentral()
}

dependencies {
    implementation("net.minestom:minestom:ffm-SNAPSHOT")
}
```


src/main/java/net/minestom/server/network/NetworkBufferAllocator.java
```diff
@@ -0,0 +1,90 @@
...
+    private NetworkBufferAllocator() {
+    }
+
+    public static MemorySegment allocate(Arena arena, long byteSize) {
+        if (!ENABLE_NATIVE) return arena.allocate(byteSize);
+        Objects.requireNonNull(arena, "arena");
+        try {
+            MemorySegment segment = (MemorySegment) MALLOC_HANDLE.invokeExact(byteSize);
+            if (segment.address() == MemorySegment.NULL.address())
+                throw new OutOfMemoryError("Native failed to allocate: %d".formatted(byteSize));
```
mudkipdev: Here we treat a null pointer from malloc as an allocation failure, but the C spec explicitly allows `malloc(0)` to return NULL even when the call succeeds. As soon as `minestom.force-native-allocation=true` is enabled (which this patch now does for the test JVM), any code path that allocates a zero-length buffer – e.g. `NetworkBuffer.resizableBuffer(0)` (used in NetworkBufferTest) or `NetworkBuffer.trimmed()` on an empty buffer – will pass byteSize == 0 to this method. On a libc that returns NULL for that case, we will now throw OutOfMemoryError even though nothing is wrong, which breaks those APIs. Please special-case zero-length allocations (e.g. bypass the native path or accept the null segment) instead of treating them as OOM.

src/main/java/net/minestom/server/adventure/AdventurePacketConvertor.java
mudkipdev, line 103: Boss-bar flags are accumulated with `val |= flag.ordinal()`. The first flag (ordinal 0) never sets any bits, and the others set the wrong bit positions. Every boss-bar flag payload sent from these helpers is therefore wrong once multiple flags are involved.

src/main/java/net/minestom/server/crypto/MessageSignature.java
mudkipdev, line 43: `Packed.write` only serializes the embedded signature when `id == 0`, but the protocol uses `id == -1` to mean "full signature present". We now emit the varint 0 without the following 256 bytes, so the reader later tries to consume a signature that was never written, corrupting the remainder of the packet.

src/main/java/net/minestom/server/network/NetworkBufferAllocator.java
kermandev: Thanks, didnt see the section under `Nonportable behavior`

src/main/java/net/minestom/server/adventure/AdventurePacketConvertor.java
```diff
@@ -12,39 +11,59 @@
...
+        COLOR_ID_MAP.put(NamedTextColor.WHITE, 15);
+
+        ObjectArray<NamedTextColor> array = ObjectArray.singleThread(16);
+        COLOR_ID_MAP.forEach((key, value) -> array.set(value, key));
+
+        NAMED_TEXT_COLOR_ID_MAP = Map.copyOf(COLOR_ID_MAP);
+        ID_NAMED_TEXT_COLOR_MAP = array.toList();
     }
 
+    @ApiStatus.Experimental
```
mworzala, line 60: why experimental?

src/main/java/net/minestom/server/network/packet/server/play/StopSoundPacket.java
```diff
@@ -1,35 +1,96 @@
...
+import org.jetbrains.annotations.ApiStatus;
+
+import java.util.Objects;
 
 import static net.minestom.server.network.NetworkBuffer.*;
 
-public record StopSoundPacket(byte flags, @Nullable Sound.Source source,
-                              @Nullable String sound) implements ServerPacket.Play {
-    public static final NetworkBuffer.Type<StopSoundPacket> SERIALIZER = new Type<>() {
+public record StopSoundPacket(Action action) implements ServerPacket.Play {
```
mworzala, line 13: seems very complex for what could just be some factory methods on this class

src/main/java/net/minestom/server/adventure/MinestomAdventure.java
```diff
@@ -34,8 +35,10 @@ public final class MinestomAdventure {
 
     /**
      * If components should be automatically translated in outgoing packets.
+     * @deprecated Use {@link ServerFlag#AUTOMATIC_COMPONENT_TRANSLATION} instead.
      */
-    public static boolean AUTOMATIC_COMPONENT_TRANSLATION = false;
+    @Deprecated(forRemoval = true) // make final to prevent builds that activate it as it would be no op.
+    public static final boolean AUTOMATIC_COMPONENT_TRANSLATION = ServerFlag.AUTOMATIC_COMPONENT_TRANSLATION;
```
mworzala, line 41: its a break anyway so why keep?

src/main/java/net/minestom/server/entity/metadata/other/PaintingVariant.java
```diff
@@ -16,8 +16,8 @@ public sealed interface PaintingVariant extends Holder.Direct<PaintingVariant>,
             NetworkBuffer.INT, PaintingVariant::width,
             NetworkBuffer.INT, PaintingVariant::height,
             NetworkBuffer.KEY, PaintingVariant::assetId,
-            NetworkBuffer.COMPONENT.optional(), PaintingVariant::title,
-            NetworkBuffer.COMPONENT.optional(), PaintingVariant::author,
+            NetworkBuffer.OPT_CHAT, PaintingVariant::title,
+            NetworkBuffer.OPT_CHAT, PaintingVariant::author,
```
mworzala, line 20: opt chat is really gross it should be removed and just use optional() or renamed to OPTIONAL_COMPONENT

src/main/java/net/minestom/server/entity/metadata/villager/VillagerMeta.java
```diff
@@ -78,15 +78,15 @@ public enum Level {
...
+            return (byte) (this.ordinal() + 1);  // Villager levels are 1-indexed
         }
 
-        private static Level fromProtocolId(int value) {
+        private static Level fromProtocolId(byte value) {
             return VALUES[value - 1];
         }
 
-        public static final NetworkBuffer.Type<Level> NETWORK_TYPE = NetworkBuffer.VAR_INT.transform(Level::fromProtocolId, Level::toProtocolId);
+        public static final NetworkBuffer.Type<Level> NETWORK_TYPE = NetworkBuffer.BYTE.transform(Level::fromProtocolId, Level::toProtocolId);
```
mworzala, line 89: this is wrong: https://mcsrc.dev/1/26.1.2/net/minecraft/world/entity/npc/villager/VillagerData#L39

src/main/java/net/minestom/server/entity/GameMode.java
```diff
@@ -37,21 +38,18 @@ public boolean instantBreak() {
...
     }
 
-    private static final GameMode[] VALUES = values();
+    public static final NetworkBuffer.Type<GameMode> NETWORK_TYPE = NetworkBuffer.Enum(GameMode.class);
 
-    public static final NetworkBuffer.Type<GameMode> NETWORK_TYPE = BYTE.transform(
-            id -> VALUES[id],
-            gameMode -> (byte) gameMode.ordinal()
-    );
+    public static final NetworkBuffer.Type<@Nullable GameMode> OPT_NETWORK_TYPE = new NetworkBuffer.Type<>() {
```
mworzala, line 43: OPTIONAL_ (ik its not related, but might as well clean it up to be standardized)

src/main/java/net/minestom/server/network/packet/client/common/ClientResourcePackStatusPacket.java
```diff
@@ -8,15 +8,15 @@
...
-import static net.minestom.server.network.NetworkBuffer.VAR_INT;
+import static net.minestom.server.network.NetworkBuffer.BYTE;
 
 public record ClientResourcePackStatusPacket(
         UUID id,
         ResourcePackStatus status
 ) implements ClientPacket.Configuration, ClientPacket.Play {
     public static final NetworkBuffer.Type<ClientResourcePackStatusPacket> SERIALIZER = NetworkBufferTemplate.template(
             UUID, ClientResourcePackStatusPacket::id,
-            VAR_INT.transform(ClientResourcePackStatusPacket::readStatus, ClientResourcePackStatusPacket::statusId), ClientResourcePackStatusPacket::status,
```
mworzala, line 19: this is not correct. we should not represent the types differently even if it will never encode with multiple bytes

src/main/java/net/minestom/server/network/packet/client/handshake/ClientHandshakePacket.java
```diff
@@ -21,7 +21,7 @@ public record ClientHandshakePacket(int protocolVersion, String serverAddress,
             VAR_INT, ClientHandshakePacket::protocolVersion,
             STRING, ClientHandshakePacket::serverAddress,
             UNSIGNED_SHORT, ClientHandshakePacket::serverPort,
-            VAR_INT.transform(Intent::fromId, Intent::id), ClientHandshakePacket::intent,
```
mworzala, line 24: same leave as varint

kermandev:
Is being slowly merged in chunks into master instead of one Pull Request.

PR #2962, opened 2025-10-31
fix: ensure the item stack size of non-air-items is always >0 in ItemStack#withMaterial

Bloeckchengrafik:
## Proposed changes

Fixes the bug shown in [this](https://gist.github.com/Bloeckchengrafik/a29c90e19d7c219ddfc55e5460117739) notebook with which a non-air-item can become 0-amounted. That results in the client disconnecting. Also fix a related bug ensuring 0 amount for air

## Types of changes

## Checklist

merging your code._

_Want to try this PR in your project? Use my [build](https://ci.everbuild.org/everbuild-org/minestom-pr-builds/7/1/2)! For usage instructions, check the bottom of the log_

### DIFF
--- a/src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
@@ -202,4 +202,27 @@ public class EntityMeta {
             metadata.set(MetadataDef.CUSTOM_NAME, (Component) value);
     }
 
+    /**
+     * Retrieves the value of the specified metadata entry.
+     *
+     * @param entry The metadata entry to retrieve the value from.
+     * @param <T>   The type of the metadata value.
+     * @return The value associated with the specified metadata entry.
+     */
+    @ApiStatus.Experimental
+    public <T> T get(MetadataDef.Entry<T> entry) {
+        return metadata.get(entry);
+    }
+
+    /**
+     * Sets the value of the specified metadata entry.
+     *
+     * @param entry The metadata entry to be updated.
+     * @param value The value to assign to the specified metadata entry.
+     * @param <T>   The type of the metadata value.
+     */
+    @ApiStatus.Experimental
+    public <T> void set(MetadataDef.Entry<T> entry, T value) {
+        metadata.set(entry, value);
+    }
 }


src/main/java/net/minestom/server/item/ItemStackImpl.java
```diff
@@ -96,7 +96,10 @@ public ItemStack with(Consumer<ItemStack.Builder> consumer) {
 
     @Override
     public ItemStack withMaterial(Material material) {
-        return new ItemStackImpl(material, amount, components);
+        int newAmount = amount;
+        if (material != Material.AIR && newAmount == 0) newAmount = 1;
+        if (material == Material.AIR) newAmount = 0;
```
kermandev: Whatever solution this likely should just return the air instance

mworzala:
I am comfortable to switch to amount of 1. Amount of zero for non air items is nonsensical and certainly not the intended behavior. I do think it would also be good to set air in the canonical constructor of ItemStackImpl if amount is zero as well as throw in the serializer to make the issue more obvious. However it doesnt need to be part of this PR.

However kermans comment about returning the air instance is relevant prior to merging.

Bloeckchengrafik:
@mworzala Done. Added the check in the canonical constructor and the exception in the serializer as well. Thanks for reviewing!

src/main/java/net/minestom/server/item/ItemStackImpl.java
```diff
@@ -30,6 +30,10 @@ public void write(NetworkBuffer buffer, ItemStack value) {
                     return;
                 }
 
+                if (value.amount() <= 0) {
+                    throw new IllegalArgumentException("ItemStack amount must be greater than 0 if not air");
```
kermandev: This should include information on the item stack, as it helps people narrow it down.

ItemStack \`%s\` must be greater than 0 if not zero, found %d 

Pretty sure the toString is probably fine or the namespace, but include the amount cause its not clamped

Bloeckchengrafik: Should components be shown as well?

kermandev: Sorry for getting back to you late, but a toString on the ItemStack is probably fine.

src/main/java/net/minestom/server/item/ItemStackImpl.java
```diff
@@ -96,7 +103,8 @@ public ItemStack with(Consumer<ItemStack.Builder> consumer) {
 
     @Override
     public ItemStack withMaterial(Material material) {
-        return new ItemStackImpl(material, amount, components);
+        if (material == Material.AIR) return ItemStack.AIR;
+        return new ItemStackImpl(material, amount == 0 ? 1 : amount, components);
```
mworzala: nitpick: Math.min(1, amount) is more legible

kermandev (approved):

PR #2963, opened 2025-11-01
Fix not being able to create blank issues

mudkipdev:


kermandev:
Could you specify in your PR why you would want to allow blank issues, and maybe if the issue template doesn't work for the use case; maybe it should be added.

PR #2980, opened 2025-11-24
feat: add `PlayerInfoUpdateEvent` to control tab list contents

### LINKED ISSUE #2076: Ability to hide player from Tab

SorkoPiko:
## Proposed changes

Add an event which allows control of the recipients of the tab list update packet (`PlayerInfoUpdatePacket`). It's cancellable, and provides access to a mutable set of recipients for granular control.

## Types of changes

## Checklist

merging your code._

## Further comments

Supersedes #2113 and closes #2076

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataDef.java
+++ b/src/main/java/net/minestom/server/entity/MetadataDef.java
@@ -114,9 +114,8 @@ public sealed class MetadataDef {
 
     public static final class AreaEffectCloud extends MetadataDef {
         public static final Entry<Float> RADIUS = index(0, Metadata::Float, 0.5f);
-        public static final Entry<Integer> COLOR = index(1, Metadata::VarInt, 0);
-        public static final Entry<Boolean> IGNORE_RADIUS_AND_SINGLE_POINT = index(2, Metadata::Boolean, false);
-        public static final Entry<Particle> PARTICLE = index(3, Metadata::Particle, Particle.EFFECT);
+        public static final Entry<Boolean> WAITING = index(1, Metadata::Boolean, false);
+        public static final Entry<Particle> PARTICLE = index(2, Metadata::Particle, Particle.EFFECT);
     }
 
     public static final class FishingHook extends MetadataDef {
--- a/src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
@@ -19,20 +19,12 @@ public class AreaEffectCloudMeta extends EntityMeta {
         metadata.set(MetadataDef.AreaEffectCloud.RADIUS, value);
     }
 
-    public int getColor() {
-        return metadata.get(MetadataDef.AreaEffectCloud.COLOR);
+    public boolean isWaiting() {
+        return metadata.get(MetadataDef.AreaEffectCloud.WAITING);
     }
 
-    public void setColor(int value) {
-        metadata.set(MetadataDef.AreaEffectCloud.COLOR, value);
-    }
-
-    public boolean isSinglePoint() {
-        return metadata.get(MetadataDef.AreaEffectCloud.IGNORE_RADIUS_AND_SINGLE_POINT);
-    }
-
-    public void setSinglePoint(boolean value) {
-        metadata.set(MetadataDef.AreaEffectCloud.IGNORE_RADIUS_AND_SINGLE_POINT, value);
+    public void setWaiting(boolean value) {
+        metadata.set(MetadataDef.AreaEffectCloud.WAITING, value);
     }
 
     public Particle getParticle() {


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2402,4 +2401,22 @@ public int effectiveViewDistance() {
...
+        sendPlayPacket(packet, PlayerInfoUpdateEvent.InfoUpdateType.REMOVE);
+    }
+
+    private void sendPlayPacket(ServerPacket.Play packet, PlayerInfoUpdateEvent.InfoUpdateType type) {
+        var event = new PlayerInfoUpdateEvent(this, type, new HashSet<>(MinecraftServer.getConnectionManager().getOnlinePlayers()));
+
+        EventDispatcher.callCancellable(event, () -> {
+            for (Player recipient : event.getRecipients()) {
+                recipient.sendPacket(packet);
+            }
```
```suggestion
            PacketSendingUtils.sendGroupedPacket(event.getRecipients(), packet);
```

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -322,17 +321,17 @@ public CompletableFuture<Void> UNSAFE_init() {
...
-        for (var player : connectionManager.getOnlinePlayers()) {
-            if (player != this) {
-                sendPacket(player.getAddPlayerToList());
-                if (player.displayName != null) {
+        var infoUpdateEvent = new PlayerInfoUpdateEvent(this, PlayerInfoUpdateEvent.InfoUpdateType.UPDATE, new HashSet<>(MinecraftServer.getConnectionManager().getOnlinePlayers()));
+        EventDispatcher.callCancellable(infoUpdateEvent, () -> {
+            for (var player : infoUpdateEvent.getRecipients()) {
+                if (player != this) {
+                    sendPacket(player.getAddPlayerToList());
                     sendPacket(new PlayerInfoUpdatePacket(PlayerInfoUpdatePacket.Action.UPDATE_DISPLAY_NAME, player.infoEntry()));
```
kermandev: player.displayName is null until setDisplayName's display name is set so removing the check is wrong, annotation wise Player fields don't have nullability annotations so your IDE will likely warn.

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2402,4 +2401,22 @@ public int effectiveViewDistance() {
...
+
+    protected void sendPlayPacket(PlayerInfoUpdatePacket packet) {
+        sendPlayPacket(packet, PlayerInfoUpdateEvent.InfoUpdateType.UPDATE);
+    }
+
+    protected void sendPlayPacket(PlayerInfoRemovePacket packet) {
+        sendPlayPacket(packet, PlayerInfoUpdateEvent.InfoUpdateType.REMOVE);
+    }
+
+    private void sendPlayPacket(ServerPacket.Play packet, PlayerInfoUpdateEvent.InfoUpdateType type) {
```
kermandev: These names shouldn't be sendPlayPacket, maybe sendPlayerInfo()?

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2402,4 +2401,22 @@ public int effectiveViewDistance() {
...
+        var event = new PlayerInfoUpdateEvent(this, type, new HashSet<>(MinecraftServer.getConnectionManager().getOnlinePlayers()));
```
kermandev: this copy should be moved into the event

src/main/java/net/minestom/server/event/player/PlayerInfoUpdateEvent.java
```diff
@@ -0,0 +1,53 @@
...
+    private final Player player;
+    private final InfoUpdateType type;
+    private final Set<Player> recipients;
+
+    private boolean cancelled;
+
+    public PlayerInfoUpdateEvent(Player player, InfoUpdateType type, Set<Player> recipients) {
+        this.player = player;
+        this.type = type;
+        this.recipients = recipients;
```
kermandev:
```suggestion
        this.recipients = new HashSet<>(recipients);
```

src/main/java/net/minestom/server/event/player/PlayerInfoUpdateEvent.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public Player getPlayer() {
+        return player;
+    }
+
+    public InfoUpdateType getType() {
+        return type;
+    }
+
+    public Set<Player> getRecipients() {
```
kermandev, line 43: This should likely return the unmodifable view and instead add remove/add recipients methods.

In the future recipients copy might be lazy as copying the all players can be expensive.

SorkoPiko: not too sure why we would want to do this, i was thinking we should keep it mutable to maintain parity with Spigot's `EntityExplodeEvent`, for example, which allows conditional calls like `event.getRecipients().removeIf(p -> !canSee(p, event.getPlayer()));` as opposed to iterating over it one collection and modifying it elsewhere (if i understand you correctly).

src/main/java/net/minestom/server/entity/Player.java
SorkoPiko: yeah that makes more sense, i had been calling it that to keep parity with `PacketSendingUtils.sendPlayPacket` but it would make more sense to name it something more specific, yeah

src/main/java/net/minestom/server/event/player/PlayerInfoUpdateEvent.java
SorkoPiko: also, regarding the lazy copying, i just implemented that on [line 54](https://github.com/Minestom/Minestom/pull/2980/files#diff-feea47e8a6884d7c81b736cc475c272fabb17844082c5234ea4b74d84ca69a77R54).

Andrei965:
Hi! Sorry if I overlooked something, but shouldn’t there be an option to re-evaluate the condition for each player, similar to how [updateViewableRule ](https://github.com/Minestom/Minestom/blob/bb7cdc8232d549cd04191ce83e3086d5751f6926/src/main/java/net/minestom/server/entity/Entity.java#L479) handles it for viewable entities?

SorkoPiko:
it wouldn't make too much sense to have seperate events for each player, as this is an event, not a predicate. instead, you can just check with `.getPlayer()` and switch logic accordingly.

SorkoPiko:
this is meant to allow you to control who sees who, essentially. you have access to the type of change (update/remove), the recipients of the change and the player who is changing. a predicate was initially suggested for this, but for more control, it's easier to use an event as there can be multiple listeners as well as operating on the recipients as you wish, instead of having an individual check per recipient.

SorkoPiko:
the event gives you the type of player entry change (add/remove), the player who's info is being sent out, as well as all the recipients of the update. that seems like enough information to at least control the packets - while it's not a complete tab list API, it's definitely a start and allows developers to easily hide new joining players. i'm working on a more complete API as we speak, but for now i think this PR serves its purpose just fine.

PR #2982, opened 2025-11-25
feat: extensive batch API improvements

### LINKED ISSUE #994: add more BlockBatches utils
thinks like:

`merge(other)`: merges other blockbatch in itself
`mirrorX()`: mirrors the batch on the x pos
`mirrorY()`: mirrors the batch on the y pos
`mirrorZ()`: mirrors the batch on the z pos
`rotate(clockwise, degrees)` rotates the batch

would be cool to have but i dont know how hard it is to implement

### LINKED ISSUE #2766: [Bug]: `AbsoluteBlockBatch` completion callback is never called if no blocks were set
### Affected Version

8209a1d1e9

(The latest version on git still has this problem, code unchanged)

### Describe the bug

Applying an empty `AbsoluteBlockBatch` with a completion callback does not fire the completion callback. There is also no other public way to detect if it was completed because of this, as to my knowledge there is no way to check if an `AbsoluteBlockBatch` is empty or not.

I assume this is not intended behavior as it means that an empty batch will never return, and if it is there should be a way to still check for completion even if the batch is empty.

The reason this happens is because the callback is only executed in the for-loop over batch instructions:

https://github.com/Minestom/Minestom/blob/42e0d212660d0aa307ea5fb731f22aa4e8843674/src/main/java/net/minestom/server/instance/batch/AbsoluteBlockBatch.java#L147-L153 

1. Create an `AbsoluteBlockBatch`
2. Apply it with a callback
3. Notice the callback does not get called

### Code sample

```java
// Use any instance
Instance instance = ...;

final AbsoluteBlockBatch batch = new AbsoluteBlockBatch();
batch.apply(instance, () -> {
    System.out.println("This is never printed");
});
```

SorkoPiko:
## Proposed changes

Add a large collection of API methods to the `RelativeBlockBatch` to greatly expand the possibilities of data transformation within batches, removing the need to evaluate it manually before creating the batch. New transformative methods added (shorthands skipped):
- `rotate(Axis axis, Point origin, int quarterTurns)`: Rotates this batch around the specified axis.
- `rotate(Axis axis, Point origin, double degrees)`: Rotates this batch around the specified axis using degrees. (Experimental)
- `mirror(Axis axis, Point origin)`: Mirrors this batch over the plane perpendicular to the specified axis.
- `mergeWithOffset(RelativeBlockBatch other, int dx, int dy, int dz)`: Merges this batch with another batch, starting with a copy of this batch and then adding blocks from the other batch translated by the specified offset vector. Blocks from the other batch overwrite any conflicts.
- `subtractWithOffset(RelativeBlockBatch other, int dx, int dy, int dz, @Nullable Block defaultBlock)`: Subtracts another batch from this batch with an offset, removing blocks at matching positions and replacing them with a default block.
- `translate(int dx, int dy, int dz)`: Translates this batch by the specified offsets.
- `transform(Function<Block, Block> transformer)`: Transforms this batch by applying a function to each block.
- `copy()`: Creates a copy of this batch.

New getters added:
- `int getBlockCount()`: Returns the number of blocks in this batch.
- `BoundingBox getBounds()`: Returns the bounding box of this batch.

Additionally, the `Batch` interface now has a new method:
- `Set<Long> getAffectedChunks()`: Gets the set of chunk indices that will be affected by applying this batch.

All methods are annotated accordingly and contain extensive documentation.

Additionally, this PR also makes sure the `AbsoluteBlockBatch::apply`'s `callback` is called (as per #2766), regardless of whether any chunks have been affected. This is a potentially breaking change, so the javadoc has been updated accordingly.
This also fixes a previously undocumented bug with the `ChunkBatch` where it would fail to acquire the `Chunk`'s lock before calling `.setBlock()` on it.

## Types of changes

## Checklist

merging your code._

## Further comments

Closes #994, fixes #2766.
If tests are required, let me know what I should be testing (and potentially how).


src/main/java/net/minestom/server/instance/batch/RelativeBlockBatch.java
```diff
@@ -78,13 +87,562 @@ public void setBlock(int x, int y, int z, Block block) {
...
+    /**
+     * Rotates this batch around the Y axis.
+     *
+     * @param origin The origin point to rotate around
+     * @param quarterTurns The number of 90-degree clockwise turns
+     * @return A new rotated batch
+     */
+    @Contract(pure = true)
+    public RelativeBlockBatch rotateY(Point origin, int quarterTurns) {
+        return rotate(Axis.Y, origin, quarterTurns);
```
Bloeckchengrafik: Nit for these (all axis): Wouldn't these profit (speed-wise) from not using the slower rotation method laid out in rotate() below?

I haven't measured the exact performance implications but doing a "direct" quarter rotation might be faster by quite a bit, so it might be worth it to build a fast path for these

SorkoPiko: it's only two trigonometric function invocations per call, so i'm not too sure why it would be slower (or how we would do it faster, other than having a sin/cos lookup table, which most CPUs already kind of have internally anyway).

Bloeckchengrafik: My main issue is that the quarter rotations do a lot of floating point arithmetic while it's guaranteed to be integer-only. But this is probably not a huge deal, although measuring would be best IMO. Might do that later when I have the time.

Note that this is a nitpick and would not be a merge blocker (for me at least). 

Generally, awesome work! I've been looking forward to using batches more for worldgen and this is kinda what I was missing!

SorkoPiko: ah yeah that's true. i'll get an optimised version ready.

SorkoPiko:
important note: these methods don't mutate `this`, they return a copy of the original with the modifications applied.

Bloeckchengrafik (approved):
Looks good to me, thanks!

src/main/java/net/minestom/server/collision/BoundingBox.java
```diff
@@ -18,7 +18,7 @@ public record BoundingBox(Vec relativeStart, Vec relativeEnd) implements Shape {
     private static final BoundingBox SNEAKING = new BoundingBox(0.6, 1.5, 0.6);
     private static final BoundingBox SMALL = new BoundingBox(0.6, 0.6, 0.6);
 
-    final static BoundingBox ZERO = new BoundingBox(Vec.ZERO, Vec.ZERO);
+    public static final BoundingBox ZERO = new BoundingBox(Vec.ZERO, Vec.ZERO);
```
mworzala, line 21: should not be part of this pr (even if good change)

src/main/java/net/minestom/server/instance/batch/AbsoluteBlockBatch.java
```diff
@@ -129,6 +136,22 @@ protected AbsoluteBlockBatch apply(Instance instance, @Nullable Runnable callbac
             AtomicInteger counter = new AtomicInteger();
             Set<Chunk> updated = ConcurrentHashMap.newKeySet();
 
+            if (chunkBatchesMap.isEmpty()) {
+                if (inverse != null) inverse.readyLatch.countDown();
+                if (instance instanceof InstanceContainer) {
+                    // FIXME: put method in Instance instead
+                    ((InstanceContainer) instance).refreshLastBlockChangeTime();
```
mworzala: nothing changed, should not update last change time

src/main/java/net/minestom/server/instance/batch/Batch.java
```diff
@@ -51,6 +53,16 @@ default void awaitReady() {
      */
     void clear();
 
+    /**
+     * Gets the set of chunk indices that will be affected by applying this batch.
+     * <p>
+     * Each chunk index is a {@code long} value representing the unique identifier of a chunk,
+     * computed using {@link CoordConversion#chunkIndex(int, int)}.
```
mworzala, line 61: should mention CoordConversion.chunkIndexGetX/Z since those are the relevant methods to a consumer of this method

src/main/java/net/minestom/server/instance/batch/ChunkBatch.java
```diff
@@ -159,6 +192,7 @@ protected ChunkBatch apply(Instance instance,
     /**
      * Applies this batch in the current thread, executing the callback upon completion.
      */
+    @ApiStatus.Internal
```
mworzala: its private, it doesnt need an internal annotation

src/main/java/net/minestom/server/instance/batch/RelativeBlockBatch.java
```diff
@@ -190,4 +818,8 @@ public AbsoluteBlockBatch toAbsoluteBatch(int x, int y, int z) {
         }
         return batch;
     }
+
+    public enum Axis {
```
mworzala: should go in one of the util packages, there are other use cases besides block batches
(like [hollow-cube/schem](https://github.com/hollow-cube/schem/blob/main/src/main/java/net/hollowcube/schem/util/Axis.java) :D)

src/main/java/net/minestom/server/instance/batch/ChunkBatch.java
```diff
@@ -80,6 +84,35 @@ public void awaitReady() {
...
+     * <p>
+     * Each chunk index is a {@code long} value representing the unique identifier of a chunk,
+     * computed using {@link CoordConversion#chunkIndex(int, int)}.
+     *
+     * @param chunkX The x chunk coordinate of the origin of the batch
+     * @param chunkZ The z chunk coordinate of the origin of the batch
+     * @return A set of chunk indices affected by this batch
+     */
+    @Contract(pure = true)
+    public Set<Long> getAffectedChunks(int chunkX, int chunkZ) {
```
mworzala: since this isnt part of a common interface anyway, I dont think we need an alias of chunkIndex here...

mworzala (commented):
To me the rotation/transformation changes seem like a somewhat weird addition. 
* Unaligned rotation is a somewhat complicated subject and very much subject to different implementations for better results in various cases.
* It doesnt handle block state rotations
* it feels a little like not the point of block batches. The intent is not to be a worldedit selection, just a way to manage application of blocks.

i believe it can all be done externally anyway, perhaps with some methods to get blocks out of the batch. Seems like a better place for a library to deal with to me, and blockbatch can remain about actually applying changes.

src/main/java/net/minestom/server/collision/BoundingBox.java
SorkoPiko: i needed access

src/main/java/net/minestom/server/instance/batch/Batch.java
SorkoPiko: i could, but it kinda clutters the description. this bit is already longer than the main description. it might be worth moving the chunkIndex to a seperate class that has `from(x, z)` and `getX()`/`getY()` to avoid confusion like this.

src/main/java/net/minestom/server/instance/batch/ChunkBatch.java
SorkoPiko: what should `getAffectedChunks` do then? we need to override it. NotImplementedException?

SorkoPiko:
should i keep in rotations/transformations as experimental? while they don't handle state like you mentioned, i think they can still be a useful addition, and it's easier than having to get each project to implement it manually.

mworzala:
No i think they should be removed, they can exist as a nice library

src/main/java/net/minestom/server/instance/batch/ChunkBatch.java
mworzala: This one is not an override (or missing the override annotation). The one actually overriding can exist

src/main/java/net/minestom/server/instance/batch/Batch.java
mworzala: I dont think a separate class is necessary

src/main/java/net/minestom/server/collision/BoundingBox.java
mworzala: for what? I missed it i guess

src/main/java/net/minestom/server/instance/batch/ChunkBatch.java
SorkoPiko: seems kind of strange to just have an alias for `CoordConversion.chunkIndex(0, 0)`. i might just throw NotImplementedException, or something similar. i think it's best to either have it like it is, or just have it do nothing. i don't fully understand what you mean by "the one actually overriding can exist", as it depends on this.

src/main/java/net/minestom/server/collision/BoundingBox.java
SorkoPiko: [RelativeBlockBatch.java#L440](https://github.com/Minestom/Minestom/pull/2982/files#diff-fa4cb9989b2e2c07a378e26c3d18b54d9d910a9f54190abb54dd62867d2012acR440)?

mworzala:
The other transformations should not be here either

SorkoPiko:
i've removed `rotate`. if you want the other ones removed too, can we please talk about it on discord? these are genuinely useful features that i want to use in a project of mine, and i don't understand what's harmful about a transformer and merging/subtracting and translating. you're right, this isn't worldedit, but being able to move the origin of a batch after it's created is a useful feature, no? as well as merging and subtracting - those are just utilities to combine batches to prevent having to manually recreate a batch with both (which you can't even do, as you don't have access to the blocks until it's placed).

mworzala:

kermandev:

PR #2984, opened 2025-11-26
feat: per-instance entity view distance

### LINKED ISSUE #2485: Dynamic chunk/entity view distance

### LINKED ISSUE #2971: [Feature]: Dynamic entity view distance
### Is your feature request related to a problem?

Entity view distance is currently only set using a server flag. This makes it impossible to configure it in the program dynamically or differently for each instance.

For some server applications, the default entity view distance (5) may be too small, so it would make sense to define an appropriate entity view distance directly in the code, not via external system properties. I don't know if this may be considered a separate bug, but for example, even if a display entity sets a larger `viewRange` in its meta properties, it will disappear after around 64 blocks (around 5 chunks, like the default entity view distance), if `setAutoViewable` is set to true. However, if `setAutoViewable` is disabled (and the entity is manually added to the player), it will properly respect its `viewRange`.

This was already requested in #2485 along with the normal view distance, but only the normal view distance was then implemented in #2892.

### Describe the solution you'd like

It should be possible to change the entity view distance for each instance dynamically, like it's already possible for the normal view distance.

### Describe alternatives you've considered

_No response_

SorkoPiko:
## Proposed changes

Allows entity view distance to be configured per-entity and per-instance, with the default set to `ServerFlag.ENTITY_VIEW_DISTANCE` (as per #2971).
New methods:
- `int Entity.getViewDistance()`: gets this entity's view distance, in chunks
- `void Entity.setViewDistance(int)`: sets the entity's view distance to the value specified
- `int Instance.getEntityViewDistance()`: gets the instance's default entity view distance, in chunks
- `void Instance.setEntityViewDistance(int)`: sets the instance's default entity view distance to the value specified

## Types of changes

## Checklist

merging your code._

## Further comments

Closes #2971, closes #2485. Extends #2892.
New tests haven't been added, but existing tests have been updated to use the new instance-based view distance.


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -822,7 +822,7 @@ protected void refreshCurrentChunk(Chunk currentChunk) {
      *
      * @return the entity instance, can be null if the entity doesn't have an instance yet
      */
-    public @UnknownNullability Instance getInstance() {
+    public @Nullable Instance getInstance() {
```
mworzala: not a good change, you frequently know for sure an entity is in an instance and this becomes a very annoying null check.

src/main/java/net/minestom/server/entity/EntityView.java
```diff
@@ -269,8 +268,9 @@ private Collection<T> references() {
             final Instance instance = trackedLocation.instance();
             final Point point = trackedLocation.point();
 
+            int dynamicRange = entity.getInstance() != null ? entity.getInstance().entityViewDistance() : ServerFlag.ENTITY_VIEW_DISTANCE;
```
mworzala: the instance is 3 lines above and already null checked

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
```diff
@@ -351,4 +351,9 @@ public void forEach(Consumer<? super Player> action) {
             }
         }
     }
+
+    private int entityViewDistance(@Nullable Instance instance) {
```
mworzala: static

src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -909,12 +910,31 @@ public int viewDistance() {
     /**
      * Sets the chunk view distance of this instance
      *
-     * @param newViewDistance the new view distance
+     * @param newViewDistance the new chunk view distance
      */
     public void viewDistance(int newViewDistance) {
```
mworzala, line 926: add chunkViewDistance and deprecate `viewDistance`. its confusing with the presence of entityViewDistance

src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -120,6 +120,7 @@ public abstract class Instance implements Block.Getter, Block.Setter,
     private final ChunkCache blockRetriever = new ChunkCache(this, null, null);
 
     protected int chunkViewDistance = ServerFlag.CHUNK_VIEW_DISTANCE;
+    protected int entityViewDistance = ServerFlag.ENTITY_VIEW_DISTANCE;
```
mworzala: probably the EntityTracker for this instance should own the entity view distance, not the instance itself.

Then the slightly odd getInstance checks in entity tracker can be removed, and that is a better chain of responsibility to me

src/main/java/net/minestom/server/instance/Instance.java
SorkoPiko: yeah, i didn't want to change it as it's already pushed to master.

src/main/java/net/minestom/server/instance/Instance.java
SorkoPiko: that would allow for more config, but would very rarely be used. what about something like each `Entity` has its own view distance, which is by default the `Instance`'s, but can be overridden per-entity.

mworzala: Let's start with per instance and can discuss per entity in the future as a further layer perhaps.

SorkoPiko: i already implemented it 😕. i'll push in a few minutes for review.

SorkoPiko:
> needs some tests for different view distances.

didn't see this, oops. i'll implement this when i get home

SorkoPiko:
@mworzala sorry for the delay, should actually be working now, with all tests. let me know what you think 👍🏻

filips123:
Any chance this could get merged please?

SorkoPiko:
it's been a while, yeah. i'll update the branch and request a review when i get home.

Issue #2831, opened 2025-07-12
[Bug]: Exception Instance is null in Navigation Entity Tracker


einfachBlu:
### Affected Version

1.21.7

### Describe the bug

The CustomCreature.java:86 is just the super.tick() call for the Entity
Error is thrown:

java.lang.NullPointerException: Cannot invoke "net.minestom.server.instance.Instance.getEntityTracker()" because "instance" is null
        at net.minestom.server.entity.Entity.refreshCoordinate(Entity.java:1384) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.Entity.refreshPosition(Entity.java:1298) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.Entity.refreshPosition(Entity.java:1338) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.Entity.refreshPosition(Entity.java:1343) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.pathfinding.followers.GroundNodeFollower.moveTowards(GroundNodeFollower.java:52) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.pathfinding.Navigator.tick(Navigator.java:180) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.EntityCreature.update(EntityCreature.java:52) ~[minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.entity.Entity.tick(Entity.java:617) ~[minestom-2025.07.11-1.21.7.jar:?]
        at dev.blu.engine.entity.living.creature.CustomCreature.tick(CustomCreature.java:86) [server-runtime.jar:?]
        at net.minestom.server.thread.TickThread.tick(TickThread.java:80) [minestom-2025.07.11-1.21.7.jar:?]
        at net.minestom.server.thread.TickThread.run(TickThread.java:55) [minestom-2025.07.11-1.21.7.jar:?]

### Steps to reproduce the bug

Not that sure, I think if i remove the entity, it sometimes do not check that in the tick method there

### Code sample

```java
[Paste your code here]
```

SorkoPiko:
can you please provide a code sample? this seems like an issue with how you're mutating the Entity.

SorkoPiko:
i haven't found the actual issue (why these methods are being called), but what i did find was somewhat hilarious:
[...]

PR #2985, opened 2025-11-26
refactor: remove assertions

SorkoPiko:
## Proposed changes

Completely remove all usages of the `assert` keyword. Due to it only being enabled when a certain JVM flag is present, it is not safe for production code, especially if the project implementing Minestom doesn't enable this flag. This causes issues like #2831 (not directly caused by the lack of assertions, but they should have prevented that exception from happening).

## Types of changes

## Checklist

merging your code._

## Further comments

I'm considering setting up Checkstyle to disallow `assert`. However, this is a fairly large change that I wouldn't want to implement without prior approval. Additionally, an existing Checkstyle configuration already exists, which implies Checkstyle has been removed in the past. However, I think bringing it back wouldn't be a bad idea, especially considering how useful it can be for the project at this scale. Let me know what you think.

### DIFF
--- a/src/main/java/net/minestom/server/registry/RegistryData.java
+++ b/src/main/java/net/minestom/server/registry/RegistryData.java
@@ -258,7 +258,6 @@ public final class RegistryData {
         private final Shape occlusionShape;
 
         private BlockEntry(String namespace, Properties main, Map<Object, Object> internCache, @Nullable BlockEntry parent, @Nullable Properties parentProperties) {
-            if (parent != null && parent.blockEntityType() != null) Check.fail("BlockEntry cannot have a parent block entity type");
             this.key = parent != null ? parent.key : Key.key(namespace);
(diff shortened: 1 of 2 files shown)


SorkoPiko:
well there you go, these assertions were hiding 450 failed tests. either these are redundant, or need to be handled properly. either way, something needs to be done about it.

mworzala:
Asserts are not intended for production code generally. They are intended as sanity checks to catch unexpected conditions in tests. In production, that code may result in undefined behavior or issues. People can enable assertions or write tests for their own logic if they want.

Our usage of assertions is not super consistent admittedly, and im sure there are some cases where a real check makes more sense (eg expected bad input). However blanket removing them is not the solution. If you want to reopen a pr changing all the junit test asserts to junit assertions im down, and if you have specific cases where you think a non-assert check is better lets discussion.

PR #2986, opened 2025-11-28
fix: precision loss in SoundEffectPacket coordinates

unsurprisable:
## Proposed changes

Currently, SoundEffectPacket casts coordinates to int before the fixed-point multiplication (* 8) takes place. This results in sounds snapping to block corners, breaking the behavior expected by the Adventure API and the Minecraft Protocol.

- Updated SoundEffectPacket record to store double coordinates instead of int.

- Updated serialization to multiply by 8.0 before casting to int ((int) (x * 8.0)).

- Updated deserialization to divide by 8.0 (read(INT) / 8.0) to ensure symmetry.

- Updated createSoundPacket helper to remove premature int casting.

## Types of changes

## Checklist

merging your code._

## Further comments

(I didn't add a test since the fix is pretty trivial but I'm open to doing so)

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataDef.java
+++ b/src/main/java/net/minestom/server/entity/MetadataDef.java
@@ -114,9 +114,8 @@ public sealed class MetadataDef {
 
     public static final class AreaEffectCloud extends MetadataDef {
         public static final Entry<Float> RADIUS = index(0, Metadata::Float, 0.5f);
-        public static final Entry<Integer> COLOR = index(1, Metadata::VarInt, 0);
-        public static final Entry<Boolean> IGNORE_RADIUS_AND_SINGLE_POINT = index(2, Metadata::Boolean, false);
-        public static final Entry<Particle> PARTICLE = index(3, Metadata::Particle, Particle.EFFECT);
+        public static final Entry<Boolean> WAITING = index(1, Metadata::Boolean, false);
+        public static final Entry<Particle> PARTICLE = index(2, Metadata::Particle, Particle.EFFECT);
     }
 
     public static final class FishingHook extends MetadataDef {
--- a/src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/other/AreaEffectCloudMeta.java
@@ -19,20 +19,12 @@ public class AreaEffectCloudMeta extends EntityMeta {
         metadata.set(MetadataDef.AreaEffectCloud.RADIUS, value);
     }
 
-    public int getColor() {
-        return metadata.get(MetadataDef.AreaEffectCloud.COLOR);
+    public boolean isWaiting() {
+        return metadata.get(MetadataDef.AreaEffectCloud.WAITING);
     }
 
-    public void setColor(int value) {
-        metadata.set(MetadataDef.AreaEffectCloud.COLOR, value);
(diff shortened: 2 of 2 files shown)


src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
```diff
@@ -12,9 +12,9 @@
 public record SoundEffectPacket(
         SoundEvent soundEvent,
         Source source,
-        int x,
-        int y,
-        int z,
+        double x,
```
kermandev: This could probably be a `Point` now. Also you deleted the signature for the original constructor.

Also you can add x,y,z int but mark them deprecated for removal to instead use `#origin#x()` or something.

kermandev (changes requested):

src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
unsurprisable: I re-added the int constructor as deprecated.

I also opted to just keep the fields as doubles since there's already a constructor that takes in a `Point`.

src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
```diff
@@ -36,16 +36,26 @@ public void write(NetworkBuffer buffer, SoundEffectPacket value) {
...
                     buffer.read(LONG));
         }
     };
 
+    /**
+     * @deprecated Sound effect position will be snapped to block corners. Instead, use
+     * {@link #SoundEffectPacket(SoundEvent, Source, Point, float, float, long)} or
+     * {@link #SoundEffectPacket(SoundEvent, Source, double, double, double, float, float, long)}
+     */
+    @Deprecated(forRemoval = true)
```
kermandev, line 48: This is not inherently wrong, you can play sounds if you want from the block corners if you truly wanted to.

src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
kermandev: I guess that is acceptable for now, but x/y/z are now doubles instead of ints which have no good way of remedying that without renaming the fields. Thats why I thought Point would be more acceptable and they are freely reusable.

src/main/java/net/minestom/server/network/packet/server/play/SoundEffectPacket.java
SorkoPiko: doesn't make much sense to keep it in, though. we don't need helper functions for everything, if they want to make it play from a corner they can just make a Point with int values.

kermandev: Yes that is true, It would be much more preferable if a Point was passed to ensure your intent is clear, thats kind of what I would expect the delegate to do.

SorkoPiko:
beat me to it lol i was about to just hand you the diff

unsurprisable:
lol, yeah I do see now that a `Point` definitely would be better since we're changing the signature anyway. This should be good now?

SorkoPiko:
i approve 👍🏻 (my input means nothing though, so have fun waiting for the next review)

kermandev (approved):

PR #2991, opened 2025-12-09
refactor: make BenchmarkManager#getUsedMemory static

SorkoPiko:
## Proposed changes

Moved it from `BenchmarkManager#getUsedMemory()` to `Stats#getUsedMemory()`. The method isn't a benchmark, and it doesn't access any fields from the manager, so there's no reason to keep it there.

## Types of changes

## Checklist

merging your code._


demo/src/main/java/net/minestom/demo/PlayerInit.java
```diff
@@ -428,8 +429,8 @@ public void init() {
             if (LAST_TICK.get() == null || MinecraftServer.getConnectionManager().getOnlinePlayerCount() == 0)
                 return;
 
-            long ramUsage = benchmarkManager.getUsedMemory();
-            ramUsage /= 1e6; // bytes to MB
+            long ramUsage = Stats.getUsedMemory();
+            ramUsage /= 1024 * 1024; // bytes to MB
```
kermandev: This is wrong thats a MiB

kermandev (changes requested):
I don't see the major reason to pull this out of the benchmark manager.

kermandev (changes requested):
I still really dont see the reason to add this when the benchmark manager already exists and is always loaded. We also don't distinct from Minestom's memory footprint (there is only one we could technically track, network buffers).

SorkoPiko:
just requires extra calls to get the benchmark manager, and overall it's best practice to use static methods for methods that don't use any data from the method, and in this case the method doesn't even belong in this class as it's not being actively measured by the class.

kermandev:
I think this code snippet is better off in usercode than inside Minestom, as users can have their own stats gathering method.

SorkoPiko:
> I think this code snippet is better off in usercode than inside Minestom, as users can have their own stats gathering method.

This snippet is already in Minestom though. In that case, should it just be removed?

kermandev:
> > I think this code snippet is better off in usercode than inside Minestom, as users can have their own stats gathering method.
> 
> This snippet is already in Minestom though. In that case, should it just be removed?

Ideally Benchmark manager should just be moved to demo in my opinion, Id make a discussion in the discord about it. Ive seen people ignore it and just implement bossbar MSPT tracking and different stats dashboards.

PR #2996, opened 2025-12-14
Rework horse variant component to match vanilla

mudkipdev:
- Renamed `HorseMeta.Color` to `HorseMeta.Variant`, matching vanilla's color-only variant enum
- `HORSE_VARIANT` now serializes identically to vanilla (string in NBT, var-int id on the network)
- Removed the marking + color `Variant` wrapper class and `getVariantID`/`getVariantFromID`
- `setVariant` only changes the color and preserves the marking, like vanilla
- Added `getMarking`, `setMarking`, and `setVariantAndMarking` to `HorseMeta`
- Fixed the component `get`/`set` overrides using the marking where the color was expected
- Targets `next` due to breaking API changes


src/main/java/net/minestom/server/component/DataComponents.java
```diff
@@ -124,7 +124,7 @@ public class DataComponents {
     public static final DataComponent<RegistryKey<CowVariant>> COW_VARIANT = register("cow/variant", CowVariant.NETWORK_TYPE, CowVariant.CODEC);
     public static final DataComponent<RegistryKey<ChickenVariant>> CHICKEN_VARIANT = register("chicken/variant", ChickenVariant.NETWORK_TYPE, ChickenVariant.CODEC);
     public static final DataComponent<RegistryKey<FrogVariant>> FROG_VARIANT = register("frog/variant", FrogVariant.NETWORK_TYPE, FrogVariant.CODEC);
-    public static final DataComponent<HorseMeta.Color> HORSE_VARIANT = register("horse/variant", HorseMeta.Color.NETWORK_TYPE, HorseMeta.Color.NBT_TYPE);
+    public static final DataComponent<HorseMeta.Variant> HORSE_VARIANT = register("horse/variant", HorseMeta.Variant.NETWORK_TYPE, HorseMeta.Variant.CODEC);
```
kermandev, line 141: Color is actually correct here. The markings is determined by the horse metadata (annoying right?)

src/main/java/net/minestom/server/entity/metadata/animal/HorseMeta.java
```diff
@@ -45,20 +47,31 @@ public static Variant getVariantFromID(int variantID) {
     @SuppressWarnings("unchecked")
     protected <T> @Nullable T get(DataComponent<T> component) {
         if (component == DataComponents.HORSE_VARIANT)
-            return (T) getVariant().getMarking();
+            return (T) getVariant();
```
kermandev, line 41: Hmm it seems the original was wrong should be getColor

src/main/java/net/minestom/server/entity/metadata/animal/HorseMeta.java
```diff
@@ -45,20 +47,31 @@ public static Variant getVariantFromID(int variantID) {
...
     }
 
     @Override
     protected <T> void set(DataComponent<T> component, T value) {
         if (component == DataComponents.HORSE_VARIANT) {
-            var variant = getVariant();
-            variant.setMarking((Marking) value);
-            setVariant(variant);
-        } else super.set(component, value);
+            setVariant((Variant) value);
```
kermandev, line 48: <T> should be Color, we probably should retain the marking though

src/main/java/net/minestom/server/entity/metadata/animal/HorseMeta.java
```diff
@@ -45,20 +47,31 @@ public static Variant getVariantFromID(int variantID) {
...
     }
 
     public static class Variant {
+        public static final NetworkBuffer.Type<Variant> NETWORK_TYPE = NetworkBufferTemplate.template(
+                Marking.NETWORK_TYPE, Variant::getMarking,
+                Color.NETWORK_TYPE, Variant::getColor,
+                Variant::new
+        );
+
+        public static final Codec<Variant> CODEC = StructCodec.struct(
```
kermandev: No codec needs to exist for this

src/main/java/net/minestom/server/entity/metadata/animal/HorseMeta.java
```diff
@@ -7,6 +7,8 @@
 import net.minestom.server.entity.MetadataDef;
 import net.minestom.server.entity.MetadataHolder;
 import net.minestom.server.network.NetworkBuffer;
+import net.minestom.server.network.NetworkBufferTemplate;
+import net.minestom.server.codec.StructCodec;
 import org.jetbrains.annotations.Nullable;
 
 public class HorseMeta extends AbstractHorseMeta {
```
kermandev, line 12: No clue why the original setVariant was deprecated (as marking needs to be set), to be fair there should be little reason to have a Variant builder class and Color should become Variant. Maybe a change you would want to target next instead. 

So HoreMeta would have

set(Variant, etc...)
setVariantAndMarking(Variant variant, Marking marking)
setMarking(Marking)

getMarking()

I agree its a little weird, but horse variant is probably a weird oversight from mojang that havent bothered to fix it yet.

demo/src/main/java/net/minestom/demo/commands/HorseCommand.java
```diff
@@ -62,7 +65,7 @@ private void onHorseCommand(CommandSender sender, CommandContext context) {
         var horse = new EntityCreature(EntityType.HORSE);
         var meta = (HorseMeta) horse.getEntityMeta();
         meta.setBaby(baby);
-        meta.setVariant(new HorseMeta.Variant(marking, color));
+        horse.set(DataComponents.HORSE_VARIANT, new HorseMeta.Variant(marking, color));
```
kermandev: Sadly this component could only ever change the color, not the marking.

kermandev (changes requested):
Correct me if im wrong, but..

demo/src/main/java/net/minestom/demo/commands/HorseCommand.java
mudkipdev: Not true, in my screenshot you can see it's setting the marking correctly. The PR is correct as Minecraft represents markings + color as a single entity component named 'variant'.

kermandev: Yes cause the data component is incorrectly implemented. If you were to use this data component generically it would be incorrect

kermandev:
@mudkipdev you missed ba3873e15c420b6df4454c142e39257dd61b9947 when swapping

kermandev (approved):

PR #3000, opened 2025-12-20
Fix off-by-one error in villager level encoding

### LINKED ISSUE #2998: [Bug]: Villager level is encoded as 0-4 instead of protocol-defined 1-4
### Affected Version

+ `2025.10.31-1.21.10`
+ latest commit on master: 4b78d50

### Describe the bug

Minestom encodes villager levels using the enum ordinal values 0 to 4 (`NOVICE` to `MASTER`) via `NetworkBuffer.Enum`[1]. However, according to the Minecraft Protocol specification[2], villager level values are defined as `1` to `5`.

Due to this off-by-one encoding, the villager level sent by the server doesn't match the protocol-defined value.
Clients still render level `0` as `NOVICE`, which likely masked this issue visually, but the encoded value itself is incorrect and differs from the official protocol [2].
As a result, a villager with the `MASTER ` level is rendered by the client as `EXPERT`, appearing one level lower than intended.

[1]: https://github.com/Minestom/Minestom/blob/d34c45fbb83c437bc367315b6dfa4195256f738a/src/main/java/net/minestom/server/entity/metadata/villager/VillagerMeta.java#L64C1-L72C6
[2]: https://minecraft.wiki/w/Java_Edition_protocol/Packets#Merchant_Offers

1. Spawn a villager with `VillagerMeta.Level.NOVICE`
3. Spawn a villager with `VillagerMeta.Level.APPRENTICE`
4. Look at both villagers in the client.
5. They are rendered with the same level appearance.

### Code sample

```java
Entity novice = new Entity(EntityType.VILLAGER);
novice.editEntityMeta(VillagerMeta.class, meta -> {
    meta.setVillagerData(new VillagerMeta.VillagerData(
            VillagerType.TAIGA,
            VillagerProfession.LIBRARIAN,
            VillagerMeta.Level.NOVICE
    ));
});
Entity apprentice = new Entity(EntityType.VILLAGER);
apprentice.editEntityMeta(VillagerMeta.class, meta -> {
    meta.setVillagerData(new VillagerMeta.VillagerData(
            VillagerType.TAIGA,
            VillagerProfession.LIBRARIAN,
            VillagerMeta.Level.APPRENTICE
    ));
});
novice.setInstance(instance);
apprentice.setInstance(instance);
```

PeyaPeyaPeyang:
## Proposed changes

Villager levels were previously encoded using enum ordinal values `0` to `4` via `NetworkBuffer.Enum`. However, the Minecraft protocol defines villager level values as `1` to `5`. This mismatch caused the encoded metadata to differ from the protocol specification.

This PR fixes an off by one error in villager level encoding by removing the use of enum ordinals and introducing an explicit mapping to protocol defined values. A small test is also added to ensure correct encoding behavior.

Resolves #2998

## Types of changes

## Checklist

### DIFF
--- a/src/main/java/net/minestom/server/entity/metadata/villager/VillagerMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/villager/VillagerMeta.java
@@ -68,15 +68,17 @@ public class VillagerMeta extends AbstractVillagerMeta {
         EXPERT,
         MASTER;
 
-        private int encode() {
+        private static final Level[] VALUES = values();
+
+        private int toProtocolId() {
             return this.ordinal() + 1;  // Villager levels are 1-indexed
         }
 
-        private static Level decode(int value) {
-            return Level.values()[value - 1];
+        private static Level fromProtocolId(int value) {
+            return VALUES[value - 1];
         }
 
-        public static final NetworkBuffer.Type<Level> NETWORK_TYPE = NetworkBuffer.VAR_INT.transform(Level::decode, Level::encode);
+        public static final NetworkBuffer.Type<Level> NETWORK_TYPE = NetworkBuffer.VAR_INT.transform(Level::fromProtocolId, Level::toProtocolId);
     }
 
 }


src/main/java/net/minestom/server/entity/metadata/villager/VillagerMeta.java
```diff
@@ -68,7 +68,15 @@ public enum Level {
         EXPERT,
         MASTER;
 
-        public static final NetworkBuffer.Type<Level> NETWORK_TYPE = NetworkBuffer.Enum(Level.class);
+        private int encode() {
+            return this.ordinal() + 1;  // Villager levels are 1-indexed
+        }
+
+        private static Level decode(int value) {
+            return Level.values()[value - 1];
```
mworzala: should make values a private constant no need to copy the array constantly. And use `toProtocolId`/`fromProtocolId` or something, encode is not obvious

PeyaPeyaPeyang: Thanks for your review.
I updated the implementation to avoid repeated `values()` allocation and renamed `encode`/`decode` to `toProtocolId`/`fromProtocolId` to clarify the intent.

PR #3001, opened 2025-12-22
feat: dynamic entity meta assignment

SorkoPiko:
## Proposed changes

Widens visibility of `EntityMeta.metadata.set()/get()` by adding helper methods (`T EntityMeta#get(MetadataDef.Entry<T>)` and `void set(MetadataDef.Entry<T>, T)`).

## Types of changes

## Checklist

merging your code._

## Further comments

There's no need for these to be private - for example, I want to dynamically set metadata definitions based on a registry, and there is no way to do that except for a massive switch statement that detects the type of metadata and maps it to the respective method. All that method does is set it directly, so it is fine to expose this for users to utilise. The same applies for the getter.

### DIFF
--- a/src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
@@ -209,7 +209,7 @@ public class EntityMeta {
      * @param <T>   The type of the metadata value.
      * @return The value associated with the specified metadata entry.
      */
-    @ApiStatus.Internal
+    @ApiStatus.Experimental
     public <T> T get(MetadataDef.Entry<T> entry) {
         return metadata.get(entry);
     }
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
```diff
@@ -202,4 +202,27 @@ protected <T> void set(DataComponent<T> component, T value) {
...
     }
 
+    /**
+     * Retrieves the value of the specified metadata entry.
+     *
+     * @param entry The metadata entry to retrieve the value from.
+     * @param <T>   The type of the metadata value.
+     * @return The value associated with the specified metadata entry.
+     */
+    @ApiStatus.Internal
```
kermandev: Experimental is probably the annotation you want here

SorkoPiko: it's not really experimental though, it does what you ask it to do. it is somewhat unsafe though which is why i marked it as internal.

kermandev: Experimental is probably the annotation you want on that though, as it can still change and has side effects but can still be used as API.

SorkoPiko: done @kermandev

PR #3002, opened 2025-12-22
feat: Instance Biome interfaces

SorkoPiko:
## Proposed changes

Allows easy access to `Biome.Getter/Setter` through `Instance`.

## Types of changes

## Checklist

merging your code._

## Further comments

This just removes the need to use `getChunkAt(x >> 4, y >> 4).get/setBiome(x, y, z, ...)`. You can now call it directly.

### DIFF
--- a/src/main/java/net/minestom/server/item/ItemStack.java
+++ b/src/main/java/net/minestom/server/item/ItemStack.java
@@ -130,6 +130,14 @@ public sealed interface ItemStack extends TagReadable, DataComponent.Holder, Hov
     @Contract(value = "_, -> new", pure = true)
     ItemStack with(Consumer<Builder> consumer);
 
+    /**
+     * Returns a new ItemStack with the given Material set.
+     *
+     * @param material The material to apply
+     * @return A new item stack with the new material
+     *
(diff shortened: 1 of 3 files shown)


DasBabyPixel:
Haven't checked, but this seems like the kind of change that isn't threadsafe. Can this be safely called during chunk tick by a chunk other than the modified one?

SorkoPiko:
should be fixed.

src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -741,6 +751,12 @@ public Collection<Entity> getNearbyEntities(Point point, double range) {
         return block;
     }
 
+    @Override
+    public RegistryKey<Biome> getBiome(int x, int y, int z) {
+        Chunk chunk = getChunk(CoordConversion.globalToChunk(x), CoordConversion.globalToChunk(z));
+        return Objects.requireNonNull(chunk).getBiome(x, y, z);
```
kermandev: You need to be synchronized on chunk to retrieve a biome

kermandev (changes requested):
Needs tests

PR #3003, opened 2025-12-22
feat: `BlockInstanceEvent`

SorkoPiko:
## Proposed changes

Adds a `BlockInstanceEvent` that takes a block, its position and its instance for easier locating of blocks without relying on `EntityInstanceEvent` or inheritors.

## Types of changes

## Checklist

merging your code._

## Further comments

Using a precise instance makes more sense over just using the executing player's instance. This allows more control, and allows custom block events to be fired that aren't directly related to a player but still have an instance attached.

### DIFF
--- a/src/main/java/net/minestom/server/event/player/PlayerBlockBreakEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerBlockBreakEvent.java
@@ -7,7 +7,7 @@ import net.minestom.server.instance.Instance;
 import net.minestom.server.instance.block.Block;
 import net.minestom.server.instance.block.BlockFace;
 
-public class PlayerBlockBreakEvent implements PlayerEvent, BlockEvent, CancellableEvent {
+public class PlayerBlockBreakEvent implements PlayerInstanceEvent, BlockEvent, CancellableEvent {
 
     private final Player player;
     private final Instance instance;
--- a/src/main/java/net/minestom/server/event/player/PlayerBlockInteractEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerBlockInteractEvent.java
@@ -13,7 +13,7 @@ import net.minestom.server.instance.block.BlockFace;
  * Called when a player interacts with a block (right-click).
  * This is also called when a block is placed.
  */
-public class PlayerBlockInteractEvent implements PlayerEvent, BlockEvent, CancellableEvent {
+public class PlayerBlockInteractEvent implements PlayerInstanceEvent, BlockEvent, CancellableEvent {
 
     private final Player player;
     private final PlayerHand hand;
--- a/src/main/java/net/minestom/server/event/player/PlayerBlockPlaceEvent.java
+++ b/src/main/java/net/minestom/server/event/player/PlayerBlockPlaceEvent.java
@@ -12,7 +12,7 @@ import net.minestom.server.instance.block.BlockFace;
 /**
  * Called when a player tries placing a block.
  */
-public class PlayerBlockPlaceEvent implements PlayerEvent, BlockEvent, CancellableEvent {
+public class PlayerBlockPlaceEvent implements PlayerInstanceEvent, BlockEvent, CancellableEvent {
 
     private final Player player;
     private final Instance instance;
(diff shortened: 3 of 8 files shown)


mudkipdev:
can you give an example of an event which is a BlockEvent but not BlockInstanceEvent?

SorkoPiko:
none, but i kept it in as legacy as per matt's request
image

DasBabyPixel:
From what I see every BlockEvent already has a position associated, I'd just go the step further and say it also has an instance.

SorkoPiko:
yeah that would probably make more sense. i've marked the base BlockEvent as internal to discourage use, but i definitely think that a position without an instance doesn't really make sense (kind of like bukkit's Location). @mworzala am i good to change BlockEvent to have an instance baked into it without an intermediary?

SorkoPiko:
should be good now

src/main/java/net/minestom/server/event/player/PlayerBlockBreakEvent.java
```diff
@@ -2,33 +2,38 @@
...
-import net.minestom.server.event.trait.BlockEvent;
-import net.minestom.server.event.trait.CancellableEvent;
-import net.minestom.server.event.trait.PlayerInstanceEvent;
+import net.minestom.server.event.trait.*;
+import net.minestom.server.instance.Instance;
 import net.minestom.server.instance.block.Block;
 import net.minestom.server.instance.block.BlockFace;
 
-public class PlayerBlockBreakEvent implements PlayerInstanceEvent, BlockEvent, CancellableEvent {
+public class PlayerBlockBreakEvent implements PlayerEvent, BlockInstanceEvent, CancellableEvent {
```
kermandev: Whats the reasoning behind dropping PlayerInstanceEvent?

They should be the same

SorkoPiko: because PlayerInstanceEvent has it's own `getInstance` that derives from the entity's instance, while the whole point here is to set it ourselves.

kermandev: You should still be able to have the trait though since its overridden, as this event can only be performed by a player in a instance.

SorkoPiko: `PlayerInstanceEvent` literally just derives `EntityInstanceEvent`, and all that provides is a default handle to get the entity's instance. there is no point in having this default, as it is never used, the assertion never runs and the event can theoretically be called with a player that isn't in an instance (as it's never used).

kermandev: Your already overriding the `getInstance` of  `EntityInstanceEvent` and `PlayerInstanceEvent` if implemented. The reasoning to add it is this even can only fire with a PlayerEvent if the player is in a instance, why would it not be a PlayerInstanceEvent.

SorkoPiko: yes but i am saying what is the point of implementing it if everything it provides is either provided elsewhere, or overridden? attached is the only benefit EntityInstanceEvent provides:
image
and even then we're overriding it. so i don't see the point, as all it adds is this one method that will never even be applied in any of the events that use `BlockInstanceEvent`. we're just double inheriting `InstanceEvent` for no real reason other than confusion, it would make sense if we didn't pass the `Instance` to the method, but the whole point of `EntityInstanceEvent` is to get an instance **from the entity**, which we're **not doing**.

DasBabyPixel: The point of EntityInstanceEvent is not getting the instance from the entity. It's just a shortcut. The point of EntityInstanceEvent is to allow for event filtering and having a single type represent the combination. `event instanceof EntityInstanceEvent`. Applying the same logic to PlayerInstanceEvent:
```java
PlayerInstanceEvent event = (PlayerInstanceEvent) originalEvent;
event.getInstance();
event.getPlayer();
```
is much nicer than
```java
InstanceEvent instanceEvent = (InstanceEvent) originalEvent;
PlayerEvent playerEvent = (PlayerEvent) originalEvent;
instanceEvent.getInstance();
playerEvent.getPlayer();
```

In the end I just agree with @kermandev. This is a PlayerInstanceEvent, not just a PlayerEvent.

SorkoPiko: sounds good, i'll get that changed 👍

SorkoPiko: should be fixed now @kermandev

PR #3007, opened 2025-12-30
refactor: Point methods, docs, etc...

kermandev:
### General Changes
- All Point sub types now declare their conversion utility as deprecated to warn users (BlockVec#asBlockVec, Pos#asPos, Vec#asVec)
	- So for example `Vec.ONE.asVec()` should now show a warning.
- The Epsilion was moved from Vec to Point, as it should be used for any doubles in coordinates.
- Move section sizes and region sizes to Point as its a common measurement.
- Operators exist for every type, they cannot be pure as the function is not guaranteed to be pure or return a new type.
- Point has a few more base methods, min, max, normalize, lerp, cross, abs, neg, dot, length, angle, similarPoint, etc...
	- Point#similarPoint was added to automatically check against epsilon to prevent a breaking change.
- ~~Easing now exists for every implementation to support lerp.~~ Consider sampling EaseFunction to lerp against a point instead.
- Add slight documentation. 
- Explained in Point that all types are value based and warned against equals usage.
- Terminally deprecated fromPoint methods, instanceOf or asX are better to use instead.
- Removed MathUtil usage

#### Pos
- Yaw no longer supports `-180` (now (-180, 180]) as its a bug and not allowed by the protocol regardless.
- Pitch is now clamped between [-90, 90]
- Epsilon for view was added in Pos for floats in the range of yaw and pitch.
- similarView was added with similar reasoning as similarPoint.
- negView, absView exist to mimic abs and neg for views.
- Operators now have a static method to wrap Vec operators for x/y/z.
- Pos(Point) & Pos(Point, float, float) for Pos as asPos should be used instead and withView if required. (Not terminally up for more debate)

#### BlockVec
- Many methods are duplicated to take input BlockVec's and integers.
- Vec is now typed as the return instead of Point for double required functions.
- withBlock? now exists to take unary ints.
- BlockVec(Point) is now terminally deprecated as asBlockVec is a better substitution all the way.
- Removed alot of the usages of BlockVec(point).

#### Vec
- Easing now replaces Interpolation and is available for every implementation. Usages are now terminally deprecated.
- Vec method types have directly been moved to be Point instead where applicable.
- There are some outstanding formulas in Vec that probably could be moved to Point, but I dont understand enough of the usecase for them to mind them being Vec.
- asPosition is terminally deprecated for removal.

## Types of changes

## Checklist

merging your code._

## Further comments
If required the yaw change could be taken out because it is a bug fix/optimization. Anyways this is definitely an ABI break from my understanding at least in BlockVec changing the return types.

### PosBenchmark
```
Benchmark              (inside)  (sampleSize)  Mode  Cnt  Score   Error  Units
PosBenchmark.fixPitch      true          1024  avgt   30  0.991 ± 0.010  ns/op
PosBenchmark.fixPitch      true          4096  avgt   30  0.989 ± 0.004  ns/op
PosBenchmark.fixPitch     false          1024  avgt   30  0.989 ± 0.004  ns/op
PosBenchmark.fixPitch     false          4096  avgt   30  0.990 ± 0.012  ns/op
PosBenchmark.fixYaw        true          1024  avgt   30  0.744 ± 0.062  ns/op
PosBenchmark.fixYaw        true          4096  avgt   30  0.725 ± 0.006  ns/op
PosBenchmark.fixYaw       false          1024  avgt   30  0.777 ± 0.098  ns/op
PosBenchmark.fixYaw       false          4096  avgt   30  0.745 ± 0.038  ns/op
```


DasBabyPixel:
How about adding a SectionVec to accurately represent a section? Pweaease pretty please? If we're already here

kermandev:
> How about adding a SectionVec to accurately represent a section? Pweaease pretty please? If we're already here

I dont see the major usecase here, since BlockVec.SECTION can already do the scaling you need.

DasBabyPixel:
The use-case is to be able to directly work in section coordinates, and being able to see that in code by the type. Misusing BlockVec is possible:
We could directly use BlockVec and just say it represents sections instead of blocks. But that is not very good design imo. because a BlockVec should represent blocks, not sections. Also blockX(), blockY(), blockZ() and sectionX(), sectionY(), ... are all completely incorrect in this context.
We could also use BlockVec and multiples of 16, but that is error-prone, because we allow sub-section positions.

If we apply this logic to even BlockVec, we could say: Vec can be used instead of BlockVec, lets get rid of BlockVec.

Concrete use-cases: Lighting calculations, Chunk generation calculations, calculations for bulk-editing of blocks.

In fact, current lighting code uses BlockVec instead of something like SectionVec, with the issue that we are representing sections, not blocks, thus making the code unreadable/unmaintainable. 

TL;DR
It's a nice utility. While *technically* BlockVec can be used for this, it's ugly and leads to either poor or error-prone code.

kermandev:
> The use-case is to be able to directly work in section coordinates, and being able to see that in code by the type. Misusing BlockVec is possible: We could directly use BlockVec and just say it represents sections instead of blocks. But that is not very good design imo. because a BlockVec should represent blocks, not sections. Also blockX(), blockY(), blockZ() and sectionX(), sectionY(), ... are all completely incorrect in this context. We could also use BlockVec and multiples of 16, but that is error-prone, because we allow sub-section positions.
> 
> If we apply this logic to even BlockVec, we could say: Vec can be used instead of BlockVec, lets get rid of BlockVec.
> 
> Concrete use-cases: Lighting calculations, Chunk generation calculations, calculations for bulk-editing of blocks.
> 
> In fact, current lighting code uses BlockVec instead of something like SectionVec, with the issue that we are representing sections, not blocks, thus making the code unreadable/unmaintainable.
> 
> TL;DR It's a nice utility. While _technically_ BlockVec can be used for this, it's ugly and leads to either poor or error-prone code.

I do agree its more explicit and less error prone, but there is just too many performance downsides adding an additional type currently, and for performance sensitive operations like lighting, generation and block editing would likely never be used. I think it could be more revisited when Valhalla is more mature.

The reason BlockVec exists is because its way faster to operate on integers instead of doubles, and the memory savings is a nice side effect. The reasoning you had Pos is actually more of a contender to be removed as its basically just a wrapper of a Vec with a view (similar to WorldPos). Though it probably wont be removed.

This is also likely not the PR to add this anyways. @mworzala if he has anything else to say.

DasBabyPixel:
Huh performance impact wat? How would adding a type to a sealed interface impact performance in any measurable way?
is there a benchmark I'm not aware of?

kermandev:
> Huh performance impact wat? How would adding a type to a sealed interface impact performance in any measurable way? is there a benchmark I'm not aware of?

Sealed provides no performance improvements at run time if i remember correctly its pure semantics. The performance impact would mostly be at megamorphic call sites, which I doubt there is many but we already can have the issue since we are using 3 implementations instead of two or one. It would defensibly require testing, but likely only matters for an array of Point[].

I think the bigger problem is that its not a free abstraction (yet) that value types could be (as they are meant to be scalar replaced), so they wouldnt see much use in performance sensitive code like you have described.

DasBabyPixel:
> Sealed provides no performance improvements at run time if i remember correctly its pure semantics. The performance impact would mostly be at megamorphic call sites, which I doubt there is many but we already can have the issue since we are using 3 implementations instead of two or one. It would defensibly require testing, but likely only matters for an array of Point[].
> 
> I think the bigger problem is that its not a free abstraction (yet) that value types could be (as they are meant to be scalar replaced), so they wouldnt see much use in performance sensitive code like you have described.

Performance sensitive code usually uses the concrete type directly. Then call sites don't matter, because all methods are effectively final. I don't think this impact is measurable, but only actual tests will tell. Those tests/considerations are not really part of a refactor, so maybe another PR in the future.

src/main/java/net/minestom/server/coordinate/BlockVec.java
```diff
@@ -5,282 +5,670 @@
...
+ * coordinates are floored to the nearest integer block position.
+ * <p>
+ * Instances are immutable. All operations return new instances
+ * (either {@link BlockVec} for integer results or {@link Vec} where doubles are used).
+ *
+ * @param blockX the block X coordinate
+ * @param blockY the block Y coordinate
+ * @param blockZ the block Z coordinate
  */
+
```
mworzala: weird newline

src/main/java/net/minestom/server/coordinate/Point.java
```diff
@@ -316,34 +565,244 @@ default boolean sameBlock(int blockX, int blockY, int blockZ) {
...
-        };
+        assert !(this instanceof BlockVec) : "Should be overridden";
+        return new BlockVec(blockX(), blockY(), blockZ());
+    }
+
+    /**
+     * Represents different easing functions to use with {@link #lerp(Point, double, Easing)}.
+     */
+    @FunctionalInterface
+    interface Easing {
```
mworzala: should use ease function, it can be changed to doubles but duplicating is dumb

src/main/java/net/minestom/server/coordinate/Pos.java
```diff
@@ -178,131 +348,277 @@ public Direction facing() {
...
+     */
+    @FunctionalInterface
+    public interface Operator {
+        /**
+         * Uses a {@link Vec.Operator} for the position components (x/y/z) for a {@link Operator}.
+         *
+         * @param operator the vector operator
+         * @return the position operator
+         */
+        static Operator operator(Vec.Operator operator) {
```
mworzala, line 602: can't we just have pos operator extend vec operator and default impl a delegation

kermandev: Its not quite that simple if we want to keep the `Pos.Operator` as a functional interface as there is no pitch/yaw to delegate. We also dont want to make it so `Vec apply(...)` is the one you implement for `Pos.Operator`.

src/main/java/net/minestom/server/coordinate/Point.java
kermandev: After revisiting this. I think we shouldn't have an easing dependency on this package, instead the same behavior can simply be done another way. If you need more precision you can make double versions or more complex ones using Operator.

A basic version lerp would just be
```java
var vec = new Vec(10);
var vec2 = new Vec(1);

var eased = vec.lerp(vec2, EaseFunction.OUT_QUINT.sample(0.5f);
```

mworzala: yeah that works

PR #3008, opened 2025-12-30
Fix wrong alpha color codec being used in environment attributes

thecodertommy:
## Proposed changes

Environment attributes used RGBA-formatted strings instead of ARGB, which caused clouds and other alpha color-based properties to be improperly sent to clients.
This splits the string codec into RGBA and ARGB variants and fixes their usages. I have confirmed that existing alpha color usages, specifically particles, are not broken.

## Types of changes

## Checklist

### DIFF
--- a/src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/EntityMeta.java
@@ -202,4 +202,27 @@ public class EntityMeta {
             metadata.set(MetadataDef.CUSTOM_NAME, (Component) value);
     }
 
+    /**
+     * Retrieves the value of the specified metadata entry.
+     *
+     * @param entry The metadata entry to retrieve the value from.
+     * @param <T>   The type of the metadata value.
+     * @return The value associated with the specified metadata entry.
+     */
+    @ApiStatus.Experimental
+    public <T> T get(MetadataDef.Entry<T> entry) {
+        return metadata.get(entry);
+    }
+
+    /**
+     * Sets the value of the specified metadata entry.
+     *
+     * @param entry The metadata entry to be updated.
+     * @param value The value to assign to the specified metadata entry.
+     * @param <T>   The type of the metadata value.
+     */
+    @ApiStatus.Experimental
+    public <T> void set(MetadataDef.Entry<T> entry, T value) {
+        metadata.set(entry, value);
+    }
 }
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -53,6 +53,7 @@ import net.minestom.server.utils.chunk.ChunkCache;
 import net.minestom.server.utils.chunk.ChunkSupplier;
 import net.minestom.server.utils.validate.Check;
 import net.minestom.server.world.DimensionType;
+import net.minestom.server.world.biome.Biome;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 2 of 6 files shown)


src/main/java/net/minestom/server/color/AlphaColor.java
```diff
@@ -20,11 +21,17 @@ public final class AlphaColor extends Color implements ARGBLike {
 
     public static final NetworkBuffer.Type<ARGBLike> NETWORK_TYPE = NetworkBuffer.INT.transform(
             AlphaColor::new, color -> fromARGBLike(color).asARGB());
+
     public static final Codec<ARGBLike> CODEC = Codec.INT.<ARGBLike>transform(AlphaColor::new, color -> fromARGBLike(color).asARGB())
             .orElse(Codec.FLOAT.list(4), floats -> new AlphaColor(floats.get(3), floats.get(0), floats.get(1), floats.get(2)));
-    public static final Codec<ARGBLike> STRING_CODEC = Codec.STRING.transform(
```
kermandev, line 25: Could you keep the original but mark it is depercated for removal with javadocs pointing to the new ones.

src/main/java/net/minestom/server/color/AlphaColor.java
```diff
@@ -87,14 +94,67 @@ public AlphaColor withAlpha(int alpha) {
...
+     * This string must be in the format {@code #AARRGGBB}.
+     * @param hexARGB the input value
+     * @return a color if possible, or null if any components are invalid
+     */
+    static @Nullable AlphaColor fromARGBHexString(@Pattern("#[0-9a-fA-F]{8}") final String hexARGB) {
+        if (hexARGB.length() != 9) return null;
+        if (!hexARGB.startsWith("#")) return null;
+
+        try {
+            final int a = Integer.parseInt(hexARGB.substring(1, 3), 16);
```
kermandev: Use HexFormat#fromHexDigits here

src/main/java/net/minestom/server/color/AlphaColor.java
```diff
@@ -87,14 +94,67 @@ public AlphaColor withAlpha(int alpha) {
...
+     * This string must be in the format {@code #RRGGBBAA}.
+     * @param hexRGBA the input value
+     * @return a color if possible, or null if any components are invalid
+     */
+    static @Nullable AlphaColor fromRGBAHexString(@Pattern("#[0-9a-fA-F]{8}") final String hexRGBA) {
+        if (hexRGBA.length() != 9) return null;
+        if (!hexRGBA.startsWith("#")) return null;
+
+        try {
+            final int r = Integer.parseInt(hexRGBA.substring(1, 3), 16);
```
kermandev: Use HexFormat#fromHexDigits here and swap the lower and upper or cyclic rotate it right

src/main/java/net/minestom/server/color/AlphaColor.java
```diff
@@ -20,12 +23,27 @@ public final class AlphaColor extends Color implements ARGBLike {
...
+     * This codec uses RGBA.
+     */
+    @Deprecated
     public static final Codec<ARGBLike> STRING_CODEC = Codec.STRING.transform(
             hex -> (ARGBLike) Objects.requireNonNull(ShadowColor.fromHexString(hex)),
             color -> ShadowColor.shadowColor(color).asHexString()).orElse(CODEC);
 
+    public static final Codec<ARGBLike> RGBA_STRING_CODEC = Codec.STRING.transform(
+            hex -> (ARGBLike) Objects.requireNonNull(fromRGBAHexString(hex)),
+            color -> String.format("#%08X", AlphaColor.fromARGBLike(color).asRGBA())).orElse(CODEC);
```
kermandev, line 41: You should probably add testing the Codec in tests to ensure it can read what is written, also this doesnt need to use String#format and probably could use a similar HexFormat, but its currently better than the original shadow color implementation it seems.

kermandev (approved):

PR #3013, opened 2026-01-03
Add information about how much a blockstate blocks light

DasBabyPixel:
## Proposed changes

## Types of changes

## Checklist

merging your code._

## Further comments

### DIFF
--- a/src/main/java/net/minestom/server/registry/RegistryData.java
+++ b/src/main/java/net/minestom/server/registry/RegistryData.java
@@ -253,7 +253,7 @@ public final class RegistryData {
         private final float jumpFactor;
         private final byte packedFlags;
         private final byte lightEmission;
-        private final int lightBlocked;
+        private final byte lightBlocked;
         private final @Nullable BlockEntityType blockEntityType;
         private final @Nullable Material material;
         private final @Nullable BlockSoundType blockSoundType;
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/registry/RegistryData.java
```diff
@@ -276,6 +277,7 @@ private BlockEntry(String namespace, Properties main, Map<Object, Object> intern
             var occludes = fromParent(parent, BlockEntry::occludes, main, "occludes", Properties::getBoolean, true);
             var requiresTool = fromParent(parent, BlockEntry::requiresTool, main, "requiresTool", Properties::getBoolean, true);
             this.lightEmission = fromParent(parent, BlockEntry::lightEmission, main, "lightEmission", Properties::getInt, 0).byteValue();
+            this.lightBlocked = fromParent(parent, BlockEntry::lightBlocked, main, "lightBlock", Properties::getInt, 0).intValue();
```
kermandev: isnt the default 15? Also could prob be a byte like lightEmission (or even better use the upper 4 bits but I hate how i made this properties thing, so dont worry!)

DasBabyPixel: If we are talking air then the default is 0. If we are talking a solid block then the default is 15. I just said 0, but overall shouldn't matter since it is set for every blockstate, so the default *should* never be selected

kermandev (approved):

PR #3015, opened 2026-01-04
replace chunk synchronized with RW-Lock

DasBabyPixel:
## Proposed changes

Replace the synchronized keyword for chunks with a ReentrantReadWriteLock.
This allows future features like Area-Locks, or performance optimizations with multiple readers for a chunk at the same time (e.g. parallelize lighting calculations across Y-axis).
Also allows future APIs like tryLock(), which is not possible with the synchronized keyword.

Also, while replacing everything with the locks, I fixed one or two race conditions.

## Types of changes

## Checklist

### DIFF
--- a/src/main/java/net/minestom/server/utils/chunk/ChunkCache.java
+++ b/src/main/java/net/minestom/server/utils/chunk/ChunkCache.java
@@ -22,7 +22,7 @@ public final class ChunkCache implements Block.Getter {
         this.defaultBlock = defaultBlock;
     }
 
-    public ChunkCache(Instance instance, Chunk chunk) {
+    public ChunkCache(Instance instance, @Nullable Chunk chunk) {
         this(instance, chunk, Block.AIR);
     }
 


src/main/java/net/minestom/server/instance/LightingChunk.java
```diff
@@ -213,33 +213,41 @@ public void onGenerate() {
         }
     }
 
-    // Lazy compute occlusion map
-    public int[] getOcclusionMap() {
-        if (this.occlusionMap != null) return this.occlusionMap;
-        var occlusionMap = new int[CHUNK_SIZE_X * CHUNK_SIZE_Z];
+    protected OcclusionData getOcclusionData() {
+        assertReadLock();
```
AidanMars2: This should be a write-lock

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -313,6 +325,7 @@ protected LightData createLightData(boolean requiredFullChunk) {
     }
 
     protected Map<Heightmap.Type, long[]> getHeightmaps() {
+        assertReadLock();
```
AidanMars2: This should be a write-lock

kermandev:
I havent looked over everything but any reason you went for a read write lock instead of just a ReentrantLock? Palettes are not thread safe and you could observe some weird side effects. Also you probably dont need all the specialized methods, a simple try finally block should be fine.

DasBabyPixel:
> I havent looked over everything but any reason you went for a read write lock instead of just a ReentrantLock?

RW-Lock to allow multiple readers simultaneously. Would be very beneficial for things like parallelizing lighting across the y axis. A single lock would mean that light sections can't access the palette in parallel. It's mostly for scalability.

> Palettes are not thread safe and you could observe some weird side effects.

This is true, my assumption here is that reading does not mutate state, so read-only operations should not affect the palette. Therefore they are safe for read-only access while owning the read-lock.

> Also you probably dont need all the specialized methods, a simple try finally block should be fine.

I kept the lock/unlock methods for feature completion, while the other methods are there to make things easier/safer to use. Kind of like the acquirable API where you can lock manually or pass a consumer. I'm down to remove the helper methods, though I don't see an issue keeping them.

kermandev:
> > I havent looked over everything but any reason you went for a read write lock instead of just a ReentrantLock?
> 
> RW-Lock to allow multiple readers simultaneously. Would be very beneficial for things like parallelizing lighting across the y axis. A single lock would mean that light sections can't access the palette in parallel. It's mostly for scalability.
> 
> > Palettes are not thread safe and you could observe some weird side effects.
> 
> This is true, my assumption here is that reading does not mutate state, so read-only operations should not affect the palette. Therefore they are safe for read-only access while owning the read-lock.

Hmm seems I assumed `ReadWriteLock` allowing reading whilst writing, oops! Though there is defiantly going to be a footprint increase and probably a performance decrease (where contention is low). Probably need a jmh and jcstress bench to verify though

> > Also you probably dont need all the specialized methods, a simple try finally block should be fine.
> 
> I kept the lock/unlock methods for feature completion, while the other methods are there to make things easier/safer to use. Kind of like the acquirable API where you can lock manually or pass a consumer. I'm down to remove the helper methods, though I don't see an issue keeping them.

I do agree they are probably nice to use. Its probably overkill to have the supplier and runnable versions though as synchronized usages probably shouldn't be replaced with them.

DasBabyPixel:
> Though there is defiantly going to be a footprint increase and probably a performance decrease (where contention is low). Probably need a jmh and jcstress bench to verify though

There may be some overhead, though I hardly believe the overhead is measurable compared to the actual work that will be done once the lock is acquired. Though I'm also happy to be proven wrong. If I were to create a benchmark, creating a meaningful benchmark is not easy/obvious... It always depends on the work being done inside the lock...

> Its probably overkill to have the supplier and runnable versions though as synchronized usages probably shouldn't be replaced with them.

I don't get this... Synchronized can always be replaced with a `#withWriteLock(Runnable)` or `#supplyWithWriteLock(Supplier)`. This is the go-to replacement. Then if contention is relevant and only reads are performed, then it can be replaced with a read lock.
So why do you say usages probably shouldn't be replaced with the methods?

kermandev:
> > Though there is defiantly going to be a footprint increase and probably a performance decrease (where contention is low). Probably need a jmh and jcstress bench to verify though
> 
> There may be some overhead, though I hardly believe the overhead is measurable compared to the actual work that will be done once the lock is acquired. Though I'm also happy to be proven wrong. If I were to create a benchmark, creating a meaningful benchmark is not easy/obvious... It always depends on the work being done inside the lock...
> 
> > Its probably overkill to have the supplier and runnable versions though as synchronized usages probably shouldn't be replaced with them.
> 
> I don't get this... Synchronized can always be replaced with a `#withWriteLock(Runnable)` or `#supplyWithWriteLock(Supplier)`. This is the go-to replacement. Then if contention is relevant and only reads are performed, then it can be replaced with a read lock. So why do you say usages probably shouldn't be replaced with the methods?

Ive come back to this, sorry about seeming so hostile about this change, rereading this. Seems like a good overall change to use RRW lock instead as goal to a Chunk interface. Though I would not keep the Consumer & Supplier versions as they are bloating the API a bit and are almost always capturing lambdas which is usually bad for things that are super hot like block and biome and reading.

I would also target next because people need to be aware to not read and write at the same time or it will block forever. Well and synchronized will no longer work.

DasBabyPixel:
> > > Though there is defiantly going to be a footprint increase and probably a performance decrease (where contention is low). Probably need a jmh and jcstress bench to verify though
> > 
> > 
> > There may be some overhead, though I hardly believe the overhead is measurable compared to the actual work that will be done once the lock is acquired. Though I'm also happy to be proven wrong. If I were to create a benchmark, creating a meaningful benchmark is not easy/obvious... It always depends on the work being done inside the lock...
> > > Its probably overkill to have the supplier and runnable versions though as synchronized usages probably shouldn't be replaced with them.
> > 
> > 
> > I don't get this... Synchronized can always be replaced with a `#withWriteLock(Runnable)` or `#supplyWithWriteLock(Supplier)`. This is the go-to replacement. Then if contention is relevant and only reads are performed, then it can be replaced with a read lock. So why do you say usages probably shouldn't be replaced with the methods?
> 
> Ive come back to this, sorry about seeming so hostile about this change, rereading this. Seems like a good overall change to use RRW lock instead as goal to a Chunk interface. Though I would not keep the Consumer & Supplier versions as they are bloating the API a bit and are almost always capturing lambdas which is usually bad for things that are super hot like block and biome and reading.
> 
> I would also target next because people need to be aware to not read and write at the same time or it will block forever. Well and synchronized will no longer work.

done

DasBabyPixel:
Can we get this merged? Kinda want/need this for #2654

src/main/java/net/minestom/server/instance/LightingChunk.java
kermandev: No longer used

src/main/java/net/minestom/server/utils/chunk/ChunkCache.java
kermandev: Missing nullable for chunk

kermandev (approved):
Seems fine from my review over it. Ill wait for @mworzala to look over it to ensure I didnt miss anything.

We can revisit reducing critical sections in the future cause there seems some work that can be done there.

PR #3025, opened 2026-01-28
feat: add player input event

ThatGravyBoat:
## Proposed changes

Adds a `PlayerInputEvent` that is fired when the player sends the `ClientInputPacket`, the event contains methods indicating if the player has a key held down, if they have just released a key, or if they have just pressed a key. Also deprecates the previous `PlayerStartSneakingEvent` and `PlayerStopSneakingEvent` as both were based off the player client input and can be replaced by using the new event and checking if the player has pressed or released the shift key.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/build-src/src/main/kotlin/minestom.java-library.gradle.kts
+++ b/build-src/src/main/kotlin/minestom.java-library.gradle.kts
@@ -64,6 +64,8 @@ tasks.withType<Javadoc> {
         }
         links("https://javadoc.io/doc/com.google.code.gson/gson/${libs.versions.gson.get()}/")
         links("https://javadoc.io/doc/org.jetbrains/annotations/${libs.versions.jetbrainsAnnotations.get()}/")
+
+        tags("apiNote:a:API Note:", "implSpec:a:Implementation Requirements:", "implNote:a:Implementation Note:")
     }
 }
 


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -2207,16 +2207,24 @@ public void clearItemUse() {
     }
 
     public void refreshInput(boolean forward, boolean backward, boolean left, boolean right, boolean jump, boolean shift, boolean sprint) {
+        var oldForward = this.inputs.forward();
+        var oldBackward = this.inputs.backward();
+        var oldLeft = this.inputs.left();
+        var oldRight = this.inputs.right();
+        var oldJump = this.inputs.jump();
+        var oldShift = this.inputs.shift();
+        var oldSprint = this.inputs.sprint();
```
```suggestion
        boolean oldForward = this.inputs.forward();
        boolean oldBackward = this.inputs.backward();
        boolean oldLeft = this.inputs.left();
        boolean oldRight = this.inputs.right();
        boolean oldJump = this.inputs.jump();
        boolean oldShift = this.inputs.shift();
        boolean oldSprint = this.inputs.sprint();
```

src/main/java/net/minestom/server/event/player/PlayerInputEvent.java
```diff
@@ -0,0 +1,129 @@
...
+        this.oldShift = oldShift;
+        this.oldSprint = oldSprint;
+    }
+
+    @Override
+    public Player getPlayer() {
+        return this.player;
+    }
+
+    // Movement keys
```
kermandev, line 42: the `Key` suffix from the methods can probably be dropped, as we are not aware what key that is bound to.

src/main/java/net/minestom/server/event/player/PlayerInputEvent.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public boolean isHoldingForwardKey() {
+        return this.player.inputs().forward();
```
kermandev, line 48: isHoldingX seems like it would be checking the previous to ensure that it was held last update too.

src/main/java/net/minestom/server/event/player/PlayerInputEvent.java
```diff
@@ -0,0 +1,129 @@
...
+import net.minestom.server.event.trait.PlayerInstanceEvent;
+
+/**
+ * Called when a player's input state changes.
+ * This is raw input data and does not take into account any game mechanics.
+ * <br>
+ * For example, this event may say a player has their jump key held down
+ * even if they are in a situation where they can not actually jump.
+ */
+public class PlayerInputEvent implements PlayerInstanceEvent {
```
kermandev:
```suggestion
public final class PlayerInputEvent implements PlayerInstanceEvent {
```
For how hot it is, a value class would probably be nice in the future

src/main/java/net/minestom/server/event/player/PlayerInputEvent.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public PlayerInputEvent(Player player, boolean oldForward, boolean oldBackward, boolean oldLeft, boolean oldRight, boolean oldJump, boolean oldShift, boolean oldSprint) {
+        this.player = player;
+        this.oldForward = oldForward;
+        this.oldBackward = oldBackward;
+        this.oldLeft = oldLeft;
+        this.oldRight = oldRight;
+        this.oldJump = oldJump;
+        this.oldShift = oldShift;
+        this.oldSprint = oldSprint;
```
kermandev, line 33:
```suggestion
        this.oldSprint = oldSprint;
        super();
```
https://openjdk.org/jeps/8350458

src/main/java/net/minestom/server/event/player/PlayerStartSneakingEvent.java
```diff
@@ -4,8 +4,9 @@
 import net.minestom.server.event.trait.PlayerInstanceEvent;
 
 /**
- * Called when a player starts sneaking.
+ * @deprecated Use {@link PlayerInputEvent} instead.
  */
+@Deprecated
```
kermandev:
```suggestion
@Deprecated(forRemoval = true)
```
These are now completely redundant

src/main/java/net/minestom/server/event/player/PlayerStopSneakingEvent.java
```diff
@@ -4,8 +4,9 @@
 import net.minestom.server.event.trait.PlayerInstanceEvent;
 
 /**
- * Called when a player stops sneaking.
+ * @deprecated Use {@link PlayerInputEvent} instead.
  */
+@Deprecated
```
kermandev:
```suggestion
@Deprecated(forRemoval = true)
```
These are now completely redundant

kermandev (commented):
A test using PlayerInputEvent would be nice. Also some method documentation in the event for what state that actually means.

src/main/java/net/minestom/server/event/player/PlayerInputEvent.java
ThatGravyBoat: The key suffix is there to ensure that a developer knows these values are not representative of the actual player state, ie. a player can have the sprint key held down but may not have enough food to be able to actually sprint

src/main/java/net/minestom/server/event/player/PlayerInputEvent.java
ThatGravyBoat: `isHoldingX` should mean the current state, this is similar behavior to most keyboard input system you may use like glfws `glfwGetKey` or windows `GetKeyState` or SDLs `SDL_GetKeyboardState` I could change the name to something like `isXKeyHeldDown` but in my opinion that would be no different from `isHoldingX`

kermandev: I don't know if its worth aliasing then, I think it shouldn't really be under a different name as it creates confusion between using `player.inputs()` and then event. Maybe @mworzala could chime in what he thinks.

mworzala: I agree with gravy its pretty standard with key states, but @ThatGravyBoat please add a short javadoc just to make sure nobody is confused.

mworzala: nvm the javadoc is there

PR #3033, opened 2026-02-15
Add module class for better integration for external modules

theminecoder:
## Proposed changes

Adds a tiny interface that allows minestom modules developed outside of the codebase to have a somewhat unified way to integrate with the server.

This is no way intended to create a lifecycle management system that handles the different stages of a usual plugin system, it only serves as a replacement for modules having their own static classes to init themselves.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/src/main/java/net/minestom/server/ServerProcessImpl.java
+++ b/src/main/java/net/minestom/server/ServerProcessImpl.java
@@ -467,11 +467,6 @@ final class ServerProcessImpl implements ServerProcess {
         return new SnapshotImpl.Server(MappedCollection.plainReferences(instanceRefs), entityRefs);
     }
 
-    @Override
-    public void install(MinestomModule module) {
-        module.install(this);
-    }
-
     private final class TickerImpl implements Ticker {
         @Override
         public void tick(long nanoTime) {


src/main/java/net/minestom/server/ServerProcess.java
```diff
@@ -129,6 +129,8 @@ public interface ServerProcess extends Registries, Snapshotable {
 
     boolean isAlive();
 
+    void install(MinestomModule module);
```
mudkipdev:
```suggestion
    default void install(MinestomModule module) {
        module.install(this);
    }
```

src/main/java/net/minestom/server/ServerProcessImpl.java
```diff
@@ -467,6 +467,11 @@ public ServerSnapshot updateSnapshot(SnapshotUpdater updater) {
         return new SnapshotImpl.Server(MappedCollection.plainReferences(instanceRefs), entityRefs);
     }
 
+    @Override
+    public void install(MinestomModule module) {
+        module.install(this);
+    }
+
```
mudkipdev:
```suggestion
```

DasBabyPixel:
Unifying a method name that is never called by minestom directly: isn't this just an unused method definition? What's the point in that? It isn't called by minestom directly, so the API user would still have to call it themselves. Then they could also just directly execute their code...

mworzala:
Generally +1 on above. What is the advantage over people conventionally declaring some static init method? which could take options additionally in case the library has some parameters.

PR #3044, opened 2026-02-26
feat: Entity physics getter and setter

hapily04:
## Proposed changes

Allow people to better control what calculations are running on their entities without having to extend Entity.

## Types of changes

## Checklist

merging your code._


src/main/java/net/minestom/server/entity/Entity.java
```diff
@@ -755,6 +755,19 @@ public UUID getUuid() {
         return uuid;
     }
 
+    public boolean hasPhysics() {
```
kermandev, line 763: Thanks, add docs please, and a test would be nice.

kermandev (approved):
Seems like a good change to not have to extend entity to disable, disabled should really be the default behavior imo.

PR #3046, opened 2026-02-27
feat: typed block properties

AidanMars2:
## Proposed changes
This PR adds typed properties for all blocks using codegen.
Example:
```java
Block piston = Block.PISTON;
Block extended = piston.withProperty(Property.PISTON_EXTENDED, true);
assertTrue(extended.getProperty(Property.PISTON_EXTENDED));
Block facingEast = block.withProperty(Facing.EAST); // utility function intended for use with hard coded property values
assertEquals(Facing.EAST, facingEast.getProperty(Property.FACING));
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist

### Enum merging
The generator merges all enums that are a subset of another enum into that enum. 

If multiple properties have the same values, they use a single enum and that enum uses a hard coded combined property name for the enum. If no combined name exists, the generator throws an easily debuggable exception.
This currently only applies to the property names: `north`, `east`, `south`, and `west`, using the combined name `side`.
I tried to autogenerate such combined names, but that ended up with weird results.

### Naming
The generator tries to find a common sequence of words in the IDs of the blocks associated with each property and concatenates that with the property name. For boolean and integer properties, the process ends here.
For enum properties, the generator can decide to split some blocks off to up to 2 enum/property pairs. This currently only applies to `WallSide`+`PaleMossCarpetSide` and `ButtonFace`+`LeverFace`+`GrindStoneFace`.

If any names collide after this, they are prefixed with the values the enum contains. This currently only applies to `TopBottomHalf`+`UpperLowerHalf`.


mudkipdev:
When I [did this](https://gist.github.com/mudkipdev/55d047bbc42a185e95578e01bacc67c5), I split up the enums and properties code generation into separate steps. An example of the generated JSON would be:

```json
{
    "enums": {
        "Axis": [
            "x",
            "y",
            "z"
        ]
    },
    "properties": {
        "horizontal_axis": {
            "name": "axis",
            "type": "enum",
            "enum": "Axis",
            "properties": ["x", "z"]
        },
        "axis": {
            "name": "axis",
            "type": "enum",
            "enum": "Axis"
        }
    }
}
```

This is because Minecraft can use the same enum type for multiple properties:

```java
enum Axis {
    X, Y, Z
}

public static final EnumProperty<Direction.Axis> HORIZONTAL_AXIS = EnumProperty.create("axis", Direction.Axis.class, Direction.Axis.X, Direction.Axis.Z);
public static final EnumProperty<Direction.Axis> AXIS = EnumProperty.create("axis", Direction.Axis.class);
// both properties are "axis", but first one only accepts X/Z
```

Also, it seems like you are generating the enum names from the property names, then adding stuff in the beginning so they don't collide. (example: Mojang has `DoubleBlockHalf` and `Half`, you have `UpperLowerHalf` and `TopBottomHalf` instead) Why not just derive it from the Mojang name?

```java
public static final EnumProperty<DoubleBlockHalf> DOUBLE_BLOCK_HALF = EnumProperty.create("half", DoubleBlockHalf.class);
public static final EnumProperty<Half> HALF = EnumProperty.create("half", Half.class);
```

It's personal preference but I would propose generating the enums as inner classes. It can continued to be used the same way if you wanted to with static imports.

AidanMars2:
> Why not just derive it from the Mojang name?

The generator is currently only based on block data (from `net.minestom.data`) so I don't have access to the mojang names.

The generator is also perfectly fine with using one enum for multiple properties. It just merges subsets into one property (e.g. `Properties.HORIZONTAL_AXIS` is merged into `Properties.AXIS`) instead of making multiple property fields, but this can be changed.

src/main/java/net/minestom/server/instance/block/property/PropertyEnum.java
```diff
@@ -0,0 +1,25 @@
...
+ * <p>
+ * Implementations are expected to be {@link Enum}s
+ */
+public interface PropertyEnum {
+    /**
+     * @return the property this enum is associated with, or null if it's associated with multiple properties.
+     */
+    @Contract(pure = true)
+    @UnknownNullability
+    String property();
```
kermandev:
```suggestion
    String key();
```

src/main/java/net/minestom/server/instance/block/property/PropertyEnum.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * @return the string value associated with this enum constant.
+     */
+    @Contract(pure = true)
+    String untypedValue();
```
kermandev:
```suggestion
    String value();
```

src/main/java/net/minestom/server/instance/block/property/PropertyEnum.java
```diff
@@ -0,0 +1,25 @@
...
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.UnknownNullability;
+
+/**
+ * A property enum, associated with one or more properties in {@link Property}.
+ * Only subsets of values may be valid for some blocks.
+ * <p>
+ * Implementations are expected to be {@link Enum}s
+ */
+public interface PropertyEnum {
```
kermandev: This should probably be in the enum package so it can be sealed, unless you plan on allowing the usecase of custom properties?

src/main/java/net/minestom/server/instance/block/property/Property.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Gets the typed value associated with the untyped string value.
+     *
+     * @param value the untyped value
+     * @return the typed value equivalent of the untyped value
+     * @throws IllegalArgumentException if there is no typed value associated with the untyped value for this property.
+     */
+    @Contract(pure = true)
+    T typedValueOf(String value);
```
kermandev: parse
```suggestion
    T parse(String value);
```
or `valueOf` is also applicable

src/main/java/net/minestom/server/instance/block/property/Property.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Gets the untyped string value associated with the typed value.
+     *
+     * @param value the typed value
+     * @return the untyped value equivalent of the typed value
+     */
+    @Contract(pure = true)
+    String untypedValueOf(T value);
```
kermandev:
```suggestion
    String valueOf(T value);
```

src/main/java/net/minestom/server/instance/block/property/BooleanProperty.java
```diff
@@ -0,0 +1,21 @@
+package net.minestom.server.instance.block.property;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public record BooleanProperty(String key) implements Property<Boolean> {
```
kermandev:
```suggestion
record BooleanProperty(String key) implements Property<Boolean> {
```

src/main/java/net/minestom/server/instance/block/property/EnumProperty.java
```diff
@@ -0,0 +1,26 @@
+package net.minestom.server.instance.block.property;
+
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.function.Function;
+
+@ApiStatus.Internal
+public record EnumProperty<T extends PropertyEnum>(
```
kermandev:
```suggestion
record EnumProperty<T extends PropertyEnum>(
```

src/main/java/net/minestom/server/instance/block/property/IntegerProperty.java
```diff
@@ -0,0 +1,21 @@
+package net.minestom.server.instance.block.property;
+
+import org.jetbrains.annotations.ApiStatus;
+
+@ApiStatus.Internal
+public record IntegerProperty(String key) implements Property<Integer> {
```
kermandev:
```suggestion
record IntegerProperty(String key) implements Property<Integer> {
```

src/main/java/net/minestom/server/instance/block/property/Property.java
```diff
@@ -0,0 +1,34 @@
+package net.minestom.server.instance.block.property;
+
+import org.jetbrains.annotations.Contract;
+
+/**
+ * A property associated with some type {@code T}.
+ * Some valid values for {@code T} may not be valid for certain blocks.
```
kermandev, line 9: Documentation on usage might be helpful here

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -51,6 +53,35 @@ public sealed interface Block extends StaticProtocolObject<Block>, TagReadable,
...
+
+    /**
+     * Creates a new block with the property associated with the {@code value}
+     *
+     * @param value the typed property value with exactly one associated property
+     * @return a new block with its property changed
+     * @throws IllegalArgumentException if the value is associated with multiple properties,
+     * the property is invalid, or the value is invalid
+     */
+    default <V extends PropertyEnum> Block withProperty(V value) {
```
kermandev:
```suggestion
    default Block withProperty(PropertyEnum value) {
```
The erased version is clearer

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -51,6 +53,35 @@ public sealed interface Block extends StaticProtocolObject<Block>, TagReadable,
...
 
+    /**
+     * Creates a new block with the property {@code property} set to {@code value}.
+     *
+     * @param property the property
+     * @param value    the typed property value
+     * @return a new block with its property changed
+     * @throws IllegalArgumentException if the property or value are invalid
+     */
+    default <V, P extends Property<V>> Block withProperty(P property, V value) {
```
kermandev:
```suggestion
    default <V, P extends Property<? super V>> Block withProperty(P property, V value) {
```
Probably going to be unused bound but should be possible

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -166,6 +197,22 @@ default boolean hasNbt() {
...
 
+    /**
+     * Returns a typed property value from {@link #properties()}.
+     *
+     * @param property the property
+     * @return the typed property value
+     * @throws IllegalArgumentException if the property or the value are invalid.
+     */
+    @Contract(pure = true)
+    default <V, P extends Property<V>> V getProperty(P property) {
```
kermandev:
```suggestion
    default <V, P extends Property<? extends V>> V getProperty(P property) {
```
Probably unused but should be valid.

code-generators/src/main/java/net/minestom/codegen/PropertyGenerator.java
```diff
@@ -0,0 +1,543 @@
...
+        fromPropStringSwitch.add("default -> null;\n");
+
+        final MethodSpec.Builder typedValueOf = MethodSpec.methodBuilder("typedValueOf")
+                .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                .addAnnotation(Nullable.class)
+                .addParameter(ParameterSpec.builder(String.class, "value").build())
+                .returns(enumCN).addCode(fromPropStringSwitch.unindent().add("};").build());
+        final MethodSpec.Builder untypedValue = MethodSpec.methodBuilder("untypedValue")
+                .addModifiers(Modifier.PUBLIC)
+                .addAnnotation(Override.class).addAnnotation(NotNull.class)
```
kermandev: NotNull annotations are redundant here, please remove them from autogeneration.

src/autogenerated/java/net/minestom/server/instance/block/property/enums/Facing.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.Override;
+import java.lang.String;
+import net.minestom.server.instance.block.property.PropertyEnum;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * AUTOGENERATED by PropertyGenerator
+ */
+public enum Facing implements PropertyEnum {
```
kermandev: Might be worth using BlockFace for this case or not since this is more specialized.

code-generators/src/main/java/net/minestom/codegen/PropertyGenerator.java
```diff
@@ -0,0 +1,543 @@
...
+            return values.prefix();
+        }
+
+        @Override
+        public NamedProperty prefixedValue() {
+            return new NamedProperty(values, property, true);
+        }
+    }
+
+    private record Identifier(String[] words) {
```
kermandev: Consider using a list here, should make it more clear

code-generators/src/main/java/net/minestom/codegen/PropertyGenerator.java
```diff
@@ -0,0 +1,543 @@
...
+    }
+
+    private static class PropertyValues implements NameAllocator.Prefixable<PropertyValues> {
+        final List<String> values;
+        final Map<String, Property> properties = new LinkedHashMap<>();
+        boolean prefixIdentifier = false;
+        @Nullable List<PropertyValues> split = null;
+
+        // -1: unknown, 0: boolean, 1: integer, 2: enum
+        private int type = -1;
```
kermandev: enum instead of magic values

src/main/java/net/minestom/server/instance/block/property/IntegerProperty.java
```diff
@@ -0,0 +1,21 @@
...
+        try {
+            return Integer.valueOf(value);
+        } catch (NumberFormatException _) {
+            throw new IllegalArgumentException(
+                    "'" + value + "' is not a valid integer value for property '" + key + "'");
+        }
+    }
+
+    @Override
+    public String untypedValueOf(Integer value) {
```
kermandev: Might be worth adding a range constraint for some properties.

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -51,6 +53,35 @@ public sealed interface Block extends StaticProtocolObject<Block>, TagReadable,
...
+     * Creates a new block with the property associated with the {@code value}
+     *
+     * @param value the typed property value with exactly one associated property
+     * @return a new block with its property changed
+     * @throws IllegalArgumentException if the value is associated with multiple properties,
+     * the property is invalid, or the value is invalid
+     */
+    default <V extends PropertyEnum> Block withProperty(V value) {
+        final String property = value.property();
+        if (property == null) {
```
kermandev: Returning null here might be more applicable

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -166,6 +197,22 @@ default boolean hasNbt() {
...
+     * Returns a typed property value from {@link #properties()}.
+     *
+     * @param property the property
+     * @return the typed property value
+     * @throws IllegalArgumentException if the property or the value are invalid.
+     */
+    @Contract(pure = true)
+    default <V, P extends Property<V>> V getProperty(P property) {
+        final String stringValue = getProperty(property.key());
+        if (stringValue == null) {
```
kermandev: Returning null might be more applicable here

src/main/java/net/minestom/server/instance/block/Block.java
AidanMars2: The untyped variant of `withProperty` throws an exception if an invalid property is passed to it. Returning null here would be inconsistent with that. The exception here is also easier to debug than returning null and having a `NullPointerException` somewhere else.

src/main/java/net/minestom/server/instance/block/property/PropertyEnum.java
AidanMars2: This would still require a `permits` clause for all property enums, which is cumbersome.

src/main/java/net/minestom/server/instance/block/property/IntegerProperty.java
AidanMars2: This is something I would add if the typed properties got a specific property lookup/modification implementation, but currently it just calls the untyped version internally so it's mostly worthless here.

src/autogenerated/java/net/minestom/server/instance/block/property/enums/Facing.java
AidanMars2: I like that you can just do `Block.PISTON.withProperty(Facing.EAST)`, but I agree that `Direction` has more integration with other parts of the code base. So I don't know about this one yet.

AidanMars2:
> For naming wise if the property cannot be shared, then the `\BlockName\Property` is fine. If the value is shared it should be `Block\Property` or if it becomes a inner class `\Property` is fine.

Can you clarify?

src/main/java/net/minestom/server/instance/block/property/PropertyEnum.java
kermandev: That is true, its probably fine to put the NonExtenable annotation and put it in the docs as to why they arent extendable (Autogenerated)

src/main/java/net/minestom/server/instance/block/Block.java
kermandev: It might be better to split the API more to not allow such key() with null values to represent only the values, instead they may need to be a union like how RedstoneWireSide should be keyed by horizontal directions and then the value has to be a RedstoneWireSide.

AidanMars2: Please elaborate.

kermandev: I was thinking of something like creating a PropertyEnum.Keyed and only implement that interface where its not null, and remove key from PropertyEnum. Then the API should only accepted Keyed properties `withProperty(PropertyEnum.Keyed)`, for this method, and then delegate into a method like `withProperty(String, PropertyEnum)`. Feel free to change the naming or the ordering from this.

Ideally we would also want to type the String from `withProperty(String, PropertyEnum)` to an actual property key instead of accepting anything.

AidanMars2: The current correct way to do that is `withProperty(Property<T>, T)`, which is already typed. `withProperty(PropertyEnum.Keyed)` is reasonable though.

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -51,6 +53,30 @@ public sealed interface Block extends StaticProtocolObject<Block>, TagReadable,
...
+
+    /**
+     * Creates a new block with the property associated with the {@code value}
+     *
+     * @param value the typed property value with exactly one associated property
+     * @return a new block with its property changed
+     * @throws IllegalArgumentException if the value is associated with multiple properties,
+     * the property is invalid, or the value is invalid
+     */
+    default Block withProperty(PropertyEnum.Keyed value) {
```
kermandev: After coming back to this, I had some ideas.

Is `PropertyEnum` useful anymore? You would think there would be a method that accepts it with `Block#withProperty(String, PropertyEnum)`?

But maybe this is a good idea to show that Keyed now is flawed since its equivalent to `Property`. where `value()` is just a shortcut to `valueOf(this)`. Maybe `PropertyEnum.Keyed` is wrong abstraction.

A better idea instead is `PropertyEnum` doesnt need to exist at all and instead typed through a key defined by `EnumProperty`, The derived properties enums would simply be enums.

Would change the API from `Block.Y.withProperty(Enum.A)` to `Block.Y.withProperty(Properties.A, Enum.A)`, So it aligns with all other enums, not just these auto generated ones.

I think keeping the same format instead of making a special case API would make this better and less breaking in the future.

kermandev (commented):
Dont worry about the naming, its okay for now for autogeneration.

PR #3048, opened 2026-02-28
Add logic to discard echoed player input

GreatWyrm:
Many thanks to https://github.com/Term4/minestom-echo-fix/ for documenting the bug and outlining the fix.

## Proposed changes

Adds some logic to the sendPacketToViewersAndSelf to strip out certain parts of player metadata if we're currently processing player input and sending the metadata update packet to self. Also adds a ServerFlag to enable this behavior, by default nothing should change in the server logic.

## Types of changes

## Checklist

merging your code._

## Further comments

The tests have to be manually run, since they require the server flag to be set and the Properties class is only read once during startup. I can do some cursed reflection if needed to manually set the flag if that is desired.

### DIFF
--- a/demo/src/main/java/net/minestom/demo/Main.java
+++ b/demo/src/main/java/net/minestom/demo/Main.java
@@ -28,7 +28,6 @@ import net.minestom.server.recipe.RecipeBookCategory;
 import net.minestom.server.registry.RegistryKey;
 import net.minestom.server.registry.RegistryTag;
 import net.minestom.server.registry.TagKey;
-import net.minestom.server.statistic.StatisticType;
 import net.minestom.server.utils.time.TimeUnit;
 
 import java.io.IOException;
@@ -162,8 +161,6 @@ public class Main {
                         .set(DataComponents.CUSTOM_NAME, Component.text("abc"))
                         .build()
         ));
-
-        System.out.println(StatisticType.SLEEP_IN_BED);
 
         new PlayerInit().init();
 


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -600,10 +602,54 @@ public void remove(boolean permanent) {
 
     @Override
     public void sendPacketToViewersAndSelf(SendablePacket packet) {
```
kermandev, line 604: I think the fix being packet based promotes another problem like MetadataHolder not being extensible enough to fix this issue.

Packet filtering is likely never the correct option ~~here~~ as its very fragile.

GreatWyrm: Fair enough. The issue I see with moving it to metadata holder would be we'd still need to check packets before sending, probably just to call a transformForClient method or something.

Would that be a more acceptable solution? That way, the only code in sendPacketToViewersAndSelf would be: check ServerFlag + client input flag, check packet type, call a transform method.

kermandev: I wouldnt mind if the ServerFlag backed hack was done inside the current MetadataHolder just for instanceof Player behind a server flag, it shouldn't be as complex as the current solution and shouldn't require extra fields to track filtering. Consider offloading the filtering to a separate method that can be easily removed in the future when its no longer required.

PR #3052, opened 2026-03-03
fix: socket is blocking

### LINKED ISSUE #3050: [Bug]: Client times out in this scenario
### Affected Version

latest

### Describe the bug

I was testing something in my project when I noticed that my client timed out (still able to rejoin after timing out). If I pretty much replicate this code on a paper server the client does not time out.

0. be on windows? (server and client are on windows)

1. Join the server
2. Type b and wait for your client to catch up
3. try to type anything but b and notice how it doesn't send and you eventually time out

### Code sample

```java
import net.minestom.server.coordinate.Point;
import net.minestom.server.coordinate.Vec;
import net.minestom.server.entity.GameMode;
import net.minestom.server.entity.Player;
import net.minestom.server.event.GlobalEventHandler;
import net.minestom.server.event.player.*;
import net.minestom.server.instance.InstanceContainer;
import net.minestom.server.instance.LightingChunk;
import net.minestom.server.instance.anvil.AnvilLoader;

import java.nio.file.Path;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.ThreadLocalRandom;

public class MinimalRepro {

    static void main() {
        MinecraftServer server = MinecraftServer.init(new Auth.Online());
        InstanceContainer instance = MinecraftServer.getInstanceManager().createInstanceContainer();
        instance.setGenerator(unit -> unit.modifier().fillHeight(0, 1, Block.STONE));
        GlobalEventHandler geh = MinecraftServer.getGlobalEventHandler();
        geh.addListener(AsyncPlayerConfigurationEvent.class, event -> {
            event.getPlayer().setRespawnPoint(new Pos(0, 2, 0));
        });
        geh.addListener(PlayerChatEvent.class, event -> {
            if (!event.getRawMessage().startsWith("b")) return;
            ThreadLocalRandom current = ThreadLocalRandom.current();
            for (int i = 0; i < 33552; i++) {
                int x = current.nextInt(-1000, 1000);
                int y = current.nextInt(-1000, 1000);
                int z = current.nextInt(-1000, 1000);
                event.getPlayer().sendMessage("block vector: x: " + x + " y: " + y + " z: " + z);
            }
        });
    }
}
```

hapily04:
## Proposed changes

Fixes #3050 
Ran this on my own project and I didn't notice any side effects

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/src/main/java/net/minestom/server/network/socket/Server.java
+++ b/src/main/java/net/minestom/server/network/socket/Server.java
@@ -98,6 +98,7 @@ public final class Server {
             socket.setReceiveBufferSize(ServerFlag.SOCKET_RECEIVE_BUFFER_SIZE);
             socket.setTcpNoDelay(ServerFlag.SOCKET_NO_DELAY);
             socket.setSoTimeout(ServerFlag.SOCKET_TIMEOUT);
+            channel.configureBlocking(false);
         }
     }
 


kermandev:
Is the issue linked actually fixed? This is already the case if you use virtual threads

hapily04:
> Is the issue linked actually fixed? This is already the case if you use virtual threads

Here's a video of running the issue's code without the channelBlocking(false) line:

https://github.com/user-attachments/assets/ccd7cb7f-2d78-4180-8253-5110a4a41b47

With the channelBlocking(false) line:

https://github.com/user-attachments/assets/b5a5b3c9-bb75-4843-9d79-49bfddc7f5f7

xZeGoe:
I believe that setting blocking to false makes it listen every possible time, which makes the cpu usage work at 100%

kermandev:
I cannot reproduce #3050

hapily04:
> I cannot reproduce #3050

what if you up the loop amount?

kermandev:
> > I cannot reproduce #3050
> 
> what if you up the loop amount?

Even with >1 million it does nothing. It may be a windows issue?

> I believe that setting blocking to false makes it listen every possible time, which makes the cpu usage work at 100%

Yes this is also true, maybe it might be an idea to not send 30k chat packets... Ive experienced client stalls with high message/s

hapily04:
> > > I cannot reproduce #3050
> > 
> > 
> > what if you up the loop amount?
> 
> Even with >1 million it does nothing. It may be a windows issue?
> 
> > I believe that setting blocking to false makes it listen every possible time, which makes the cpu usage work at 100%
> 
> Yes this is also true, maybe it might be an idea to not send 30k chat packets... Ive experienced client stalls with high message/s

That's strange maybe it is a windows issue. Ideally yes I would never really be sending this many packets, but my project is a sandbox and I can't control that too much. If a user accidentally does this or something similar I can't exactly have people timing out. I confirmed with spark on my project that it does have higher idle usage and mspt as a result. I have changes that park briefly using `LockSupport.parkNanos(100_000)` in `PlayerSocketConnection` and mspt looks identical to prior the channelBlocking(false) change upon testing, so it seems to mitigate the cpu usage issue.

kermandev:
Consider trying the new write lock to see if that helps, or disabling compression. My tests were also ran without these but you can try them

```java
System.setProperty("minestom.new-socket-write-lock", "true");
MinecraftServer.setCompressionThreshold(0);
// init here...
```

Though this seems more like a reading issue, but I cannot reproduce.

hapily04:
> Consider trying the new write lock to see if that helps, or disabling compression. My tests were also ran without these but you can try them
> 
> ```java
> System.setProperty("minestom.new-socket-write-lock", "true");
> MinecraftServer.setCompressionThreshold(0);
> // init here...
> ```
> 
> Though this seems more like a reading issue, but I cannot reproduce.

Still timing out tried using each on its own and together same result

PR #3062, opened 2026-03-13
Add null checks to prevent NPE when finding nearbyEntities in the EntityTracker

### LINKED ISSUE #2812: [Bug]: NPE on getNearbyEntities()
### Affected Version

1.21.6 branch but also happend on the master branch with 1.21.5

### Describe the bug

throws the following exception a lot of times since i use the getNearbyEntities() a lot

```
java.lang.NullPointerException: Cannot invoke "net.minestom.server.instance.EntityTrackerImpl$EntityTrackerEntry.getLastPosition()" because the return value of "space.vectrix.flare.fastutil.Int2ObjectSyncMap.get(int)" is null
	at net.minestom.server.instance.EntityTrackerImpl.lambda$nearbyEntities$5(EntityTrackerImpl.java:192) ~[minestom-snapshots-1_21_6-a40d7115d4.jar:?]
	at java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:891) ~[?:?]
	at net.minestom.server.instance.EntityTrackerImpl.lambda$nearbyEntities$6(EntityTrackerImpl.java:191) ~[minestom-snapshots-1_21_6-a40d7115d4.jar:?]
	at net.minestom.server.coordinate.ChunkRange.chunksInRange(ChunkRange.java:67) ~[minestom-snapshots-1_21_6-a40d7115d4.jar:?]
	at net.minestom.server.coordinate.ChunkRange.chunksInRange(ChunkRange.java:77) ~[minestom-snapshots-1_21_6-a40d7115d4.jar:?]
	at net.minestom.server.instance.EntityTrackerImpl.nearbyEntities(EntityTrackerImpl.java:188) ~[minestom-snapshots-1_21_6-a40d7115d4.jar:?]
	at net.minestom.server.instance.Instance.getNearbyEntities(Instance.java:699) ~[minestom-snapshots-1_21_6-a40d7115d4.jar:?]
```

I am. not that sure and I cant provide my whole code base where that occured. Only thing that i can tell is that i use the following code that throws the error but I think it also requires spawning a few mobs calling that:

```java
private CustomPlayer findTarget() {
    if (this.isDead || this.instance == null) {
      return null;
    }

    var nearestEntityComparator =
        Comparator.<Entity>comparingDouble(e -> e.getDistanceSquared(this));

    CustomPlayer best = null;
    for (Entity entity : this.instance.getNearbyEntities(getPosition(), TRACKING_DISTANCE)) {
      if (!(entity instanceof CustomPlayer player)) {
        continue;
      }

      if (!player.getGameMode().equals(GameMode.SURVIVAL)) {
        // Only target players in survival mode
        continue;
      }

      var outOfSpawnZone = this.outOfSpawnZone(player.getPosition());
      if (outOfSpawnZone) {
        continue;
      }

      if (best == null || nearestEntityComparator.compare(entity, best) < 0) {
        best = player;
      }
    }

    return best;
  }
```

CustomPlayer is just the Player and outOfSpawnZone checks the player position distance to the entity. 

### Code sample

```java
As mentioned above I cant make a simple small reproducable. Hopefully that bug report helps you and maybe the bug is found based on a spontaneous idea when looking at the exception
```

### LINKED ISSUE #2914: [Bug]: The null pointer exception caused by the tick method of ItemEntity when picking up items.
### Affected Version

2025.07.11-1.21.7

### Describe the bug

Image

Creating a SchedulerTask to generate an ItemStack for each tick, then let the player pick it up, will cause the following error; my dispatcher-threads is set to 3.

### Code sample

```java
        System.setProperty("minestom.dispatcher-threads", "3")
        val instance = MinecraftServer.getInstanceManager().createInstanceContainer()
        scheduler.scheduleTask({
            val entity = ItemEntity(ItemStack.of(Material.DIAMOND))
            entity.setInstance(instance, Pos(0.0, 0.0, 0.0))
        }, TaskSchedule.nextTick(), TaskSchedule.tick(1))
```

GodModed:
## Proposed changes

Fix NPE in the EntityTracker when calling nearbyEntities 
Fixes #2812, #2914 

## Types of changes

## Checklist

## Further comments

This needs to be examined further. I believe it has to due with a race condition within nearbyEntities. entriesByEntityId seems to be out of sync with `targetEntries[target.ordinal()].chunkEntities;`


mudkipdev:
@GodModed it also closes #2914, could you add this to PR description?

TFJ5183 (commented):
Maybe its personal preference but I think not nested ifs would be better

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
```diff
@@ -183,25 +183,22 @@ public <T extends Entity> void nearbyEntities(Point point, double range, Target<
...
-                chunkEntities.forEach(entity -> {
-                    if (entriesByEntityId.containsKey(entity.getEntityId())) {
-                        final Point position = entriesByEntityId.get(entity.getEntityId()).getLastPosition();
-                        if (point.distanceSquared(position) <= squaredRange) query.accept(entity);
-                    }
-                });
-            }
+            if (chunkEntities == null || chunkEntities.isEmpty()) return;
+            chunkEntities.forEach(entity -> {
+                if (!entriesByEntityId.containsKey(entity.getEntityId())) return;
```
TFJ5183: Is this return wanted? When one entry is not in the list the whole method returns. In that case continue should be used

mudkipdev:
It's annoying but github did not auto-detect it, you'll probably need to put the second issue on a separate line

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
GodModed: It is inside of a forEach runnable. It will not.

src/main/java/net/minestom/server/instance/EntityTrackerImpl.java
```diff
@@ -183,23 +183,24 @@ public <T extends Entity> void nearbyEntities(Point point, double range, Target<
...
         } else {
             // Multiple chunks
             final int chunkRange = (int) (range / Chunk.CHUNK_SECTION_SIZE) + 1;
             ChunkRange.chunksInRange(point, chunkRange, (chunkX, chunkZ) -> {
                 @SuppressWarnings("unchecked") final var chunkEntities = (List<T>) entities.get(CoordConversion.chunkIndex(chunkX, chunkZ));
                 if (chunkEntities == null || chunkEntities.isEmpty()) return;
                 chunkEntities.forEach(entity -> {
+                    EntityTrackerEntry trackerEntry = entriesByEntityId.get(entity.getEntityId());
+                    if (trackerEntry == null) return;
                     final Point position = entriesByEntityId.get(entity.getEntityId()).getLastPosition();
```
kermandev:
```suggestion
                    final Point position = trackerEntry.getLastPosition();
```

kermandev (approved):
Fine fix for now, likely need to be refactored to remove the cause of the issue in the future.

PR #3089, opened 2026-04-04
Metadata fix

OmeWillem:
## Proposed changes

This solves the so-called sprinting bug and possibly more bugs regarding metadata changes.

## Types of changes

## Checklist

## Further comments

There has been a bug in Minestom since 1.21.4 (if I recall correctly), this bug essentially makes you jitter or stop sprinting for a brief moment if you use your "Walk Forward" key twice to sprint.
This really bothered me, since I primarily use this to sprint, and that's not great for parkouring servers like Hollow Cube.

So, this became a four hour journey of trying to understand the Minestom codebase, and the cause is quite simple.
Essentially, when the player sends the Entity Action packet "START_SPRINTING", it also sends a different packet called Player Input, this contains the status of all the Movement keys at that time. 

The MetadataHolder sends a EntityMetadataPacket right after setting the specific entry (in most cases), which caused a sort-of race condition on the client due to the SNEAKING value that's being sent by the Player Input packet. Since the sneaking value is false in most cases, it literally sends an empty metadata packet, clearing the sprinting value that is sent before.
- Client sends Input packet
- Client sends Entity Action packet (start sprinting)
- Server sends EntityMetadataPacket with sneaking set to true (value of 8)
- Server sends EntityMetadataPacket with no metadata (value of 0) (same tick btw)
- Client sends Entity Action packet (stop sprinting)

I feel like the instant update of the MetadataHolder might be a bug of it's own, for now I've added a comparison between the current value and the set value and that fixes it. It could be fixed on a higher level (EntityMeta), but this solves any other potential issues.

TL;DR: Metadata got reset due to Player Input being sent in the same tick

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataHolder.java
+++ b/src/main/java/net/minestom/server/entity/MetadataHolder.java
@@ -115,7 +115,6 @@ public final class MetadataHolder {
                 }
             } else {
                 entity.sendPacketToViewersAndSelf(new EntityMetaDataPacket(entity.getEntityId(), Map.of(id, result)));
-                new Exception("metadata set stack trace, result=" + result).printStackTrace();
             }
         }
     }
--- a/src/main/java/net/minestom/server/entity/MetadataHolder.java
+++ b/src/main/java/net/minestom/server/entity/MetadataHolder.java
@@ -88,7 +88,7 @@ public final class MetadataHolder {
         final int id = entry.index();
 
         T current = get(entry);
-        if (current.equals(value)) return;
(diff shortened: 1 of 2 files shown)


mworzala (commented):
seems reasonable regardless of any other changes later.

it should also deal with null values more gracefully. If you set a null value to null again it will still pass that check.

Can you add a test for the specific case you found so that if other changes are made in the future it preserves (or at least warns of changes to) this behavior.

mworzala (approved):

PR #3094, opened 2026-04-09
Add missing registry getters to MinecraftServer

mudkipdev:
Also change `process()` to `serverProcess` in some methods for consistency


kermandev (approved):
Seems fine for now though process() will likely be reverted later.

PR #3095, opened 2026-04-10
fix: custom model data crash exploit

mudkipdev:
Sending a `ClientCreativeInventoryActionPacket` with a malicious custom model data can cause a huge 16+ GB allocation and out-of-memory crash due to the unbounded list. You did not need to be in creative mode to trigger this.


kermandev (approved):
Seems like a reasonable limit for creative

PR #3096, opened 2026-04-10
Fix GameSpy query handler to parse only received datagram bytes

### LINKED ISSUE #3103: [Bug]: GameSpy query handling can allow stale bytes to be incorrectly read
### Affected Version

2026.04.11-1.21.11

### Describe the bug

The query loop for GameSpy query handling reused a fixed-size byte array and wrapped the entire array for parsing, which could allow stale bytes from previous packets to be considered part of the current packet.

1. Enable query on the server.
2. Use one UDP client socket and send a normal handshake packet: FE FD 09 plus any 4-byte session ID.
3. Read the challenge token from the reply.
4. Send a full stat request from the same socket: FE FD 00 plus the same session ID, the challenge token, and 4 padding bytes.
5. Immediately send a basic stat request from the same socket: FE FD 00 plus the same session ID and the challenge token, but without the 4 padding bytes.

### Code sample

```java
[Paste your code here]
```

proplayer919:
Fix UDP query packet parsing so each packet is decoded strictly from the bytes actually received.

The query loop reused a fixed-size byte array and wrapped the entire array for parsing, which could allow stale bytes from previous packets to be considered part of the current packet.

Fixes #3103.

## Changes

- Parse query payloads with `DatagramPacket` offset/length.
- Ignore datagrams shorter than the minimum header size.
- Add defensive length checks before reading:
  - handshake session id (`int`)
  - stat session id + challenge token (`int` + `int`)

## Types of changes

## Checklist

merging your code._

## Further comments


kermandev:
We have had internal discussions and the Query API is going to be removed in the next/26.1 release. If you need to use it in your project consider extracting it, as it should be very portable.

PR #3115, opened 2026-04-20
feat: Make MetadataHolder not only for Entity class

MelonHell:
This change allows the notNotifiedChanges functionality to be used independently of Entity.


src/main/java/net/minestom/server/entity/MetadataHolder.java
```diff
@@ -145,6 +145,20 @@ private byte setMaskByte(byte data, byte byteMask, int offset, byte newValue) {
         return (byte) ((data & ~byteMask) | ((newValue << offset) & byteMask));
     }
 
+    public Map<Integer, Metadata.Entry<?>> setNotifyAboutChangesAndGet() {
```
kermandev, line 148: Should update the existing method instead to return the copy instead of adding this.

mworzala:
Returning the map from setNotifyAboutChanges seems fine, but why not go a step further and just take a `Consumer<the map>` in the constructor instead of the entity and remove the dependency on `Entity` completely?

As is, you would have to wrap every change with a notifyAboutChanges pair otherwise you miss the change completely with no entity.

kermandev (changes requested):
Please add a unit test for ensuring the consumer is called without an entity should be added, for notifyAboutChanges and without.

kermandev (approved):

PR #3122, opened 2026-04-29
feat: Add NetworkBufferTemplate ClassFile backend

kermandev:
## Proposed changes

Using class generation should make every callsite megamorphic but its likely already the case since templating has 20 implementations. Instead, this change allows the write/read implementation to become monomorphic after the megamorphic call. Improving performance in polluted call sites like almost all packet reading and writing.

Currently we will introduce this into master, by having it disabled by default behind `minestom.template-compiler`  server flag which will be disabled by default until the next `next` release, where it will be enabled by default, and the option may be removed.

#3142 could be merged before this to reduce the diff.

### Benchmarks
Before (next)
```
Benchmark                                               Mode  Cnt   Score   Error  Units
NetworkSerializerLargeTemplateBenchmark.readPacket      avgt   30  15.243 ± 0.324  ns/op
NetworkSerializerLargeTemplateBenchmark.writePacket     avgt   30  15.499 ± 0.125  ns/op
NetworkSerializerPollutedTemplateBenchmark.readPacket   avgt   30   7.884 ± 0.041  ns/op
NetworkSerializerPollutedTemplateBenchmark.writePacket  avgt   30   8.377 ± 0.079  ns/op
NetworkSerializerTemplateBenchmark.readPacket           avgt   30   1.952 ± 0.011  ns/op
NetworkSerializerTemplateBenchmark.writePacket          avgt   30   1.654 ± 0.010  ns/op
```
After
```
Benchmark                                               Mode  Cnt   Score   Error  Units
NetworkSerializerLargeTemplateBenchmark.readPacket      avgt   30  11.197 ± 0.102  ns/op
NetworkSerializerLargeTemplateBenchmark.writePacket     avgt   30  10.067 ± 0.031  ns/op
NetworkSerializerPollutedTemplateBenchmark.readPacket   avgt   30   1.437 ± 0.007  ns/op
NetworkSerializerPollutedTemplateBenchmark.writePacket  avgt   30   1.195 ± 0.016  ns/op
NetworkSerializerTemplateBenchmark.readPacket           avgt   30   1.580 ± 0.006  ns/op
NetworkSerializerTemplateBenchmark.writePacket          avgt   30   1.319 ± 0.008  ns/op
```
## Types of changes

## Checklist

merging your code._

## Further comments
Needs further testing in native-image environments. As always these templates should be created during build time.
Fixes small regression caused by #3043 accidentally changed F1 -> Function, so we revert that here.
I didn't bother with the supplier and constant for NetworkBufferTemplate, maybe an addition in the future.
Yes the same strategy can be applied to StructCodec, though that is quite more complex as it has branching.

https://gist.github.com/kermandev/efa5d93df936600b0cb671d00a165ffc is what it generates


kermandev:
~~Marked as draft as I work on a ir and compiler version to do more optimizations.~~ Decided to not pursue this route due to complexity

PR #3125, opened 2026-04-30
Add support for PROXY protocol

theminecoder:
## Proposed changes

Add support for servers running infrastructure like HAProxy that provides the original connection details through the use of PROXY protocol

## Types of changes

## Checklist

merging your code._

## Further comments

I'm aware this might not be the correct place to do this but given ive had no more then 2 hours to actually get this rolling for our situation this is where it is. Am accepting suggestions for better placement, but the core issue I was running into when looking for where it put it is that I need to have it be early enough in the network process that I can read the raw bytes before packet decoding gets in the way, but also be able to manipulate there buffer so mc doesn't know the extra data ever existed.


src/main/java/net/minestom/server/network/player/PlayerSocketConnection.java
```diff
@@ -110,6 +111,15 @@ public void read(PacketParser<ClientPacket> packetParser) throws IOException {
         NetworkBuffer readBuffer = this.readBuffer;
         final long writeIndex = readBuffer.writeIndex();
         final int length = readBuffer.readChannel(channel);
+
+        if (ServerFlag.PROXY_PROTOCOL && !attemptedProxyProtocolDetection) {
+            attemptedProxyProtocolDetection = true;
```
kermandev, line 118: This should be done in the HandshakeListener where the remote address is assigned then attemptedProxyProtocolDetection can be removed

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
```diff
@@ -0,0 +1,161 @@
...
+        final long current = buffer.readIndex();
+        byte[] sig = buffer.read(NetworkBuffer.FixedRawBytes(12));
+        buffer.readIndex(current);
+
+        return Arrays.equals(sig, V2_SIG);
+    }
+
+    private static ClientInfo parseV1(NetworkBuffer buffer, SocketAddress fallback) {
+        final long current = buffer.readIndex();
+        byte[] bytes = buffer.read(NetworkBuffer.RAW_BYTES);
```
kermandev: FixedRawBytes isnt used here

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private static boolean looksLikeV1(NetworkBuffer buffer) {
+        if (buffer.readableBytes() < 6) return false;
+
+        final long current = buffer.readIndex();
+        byte[] p = buffer.read(NetworkBuffer.FixedRawBytes(6));
+        buffer.readIndex(current);
+
+        return Arrays.equals(p, "PROXY ".getBytes());
```
kermandev: NetworkBuffer.FixedRawBytes and "PROXY " should be hoisted into private static final fields.

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private static boolean looksLikeV2(NetworkBuffer buffer) {
+        if (buffer.readableBytes() < 12) return false;
+
+        final long current = buffer.readIndex();
+        byte[] sig = buffer.read(NetworkBuffer.FixedRawBytes(12));
```
kermandev: buffer.readAt is likely better here

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
```diff
@@ -0,0 +1,161 @@
...
+            return new ClientInfo(fallback, false);
+        }
+
+        int family = (famProto >> 4) & 0x0F;
+
+        if (family == 0x1) { // AF_INET
+            if (len < 12 || buffer.readableBytes() < 12)
+                return new ClientInfo(fallback, false);
+
+            byte[] src = buffer.read(NetworkBuffer.FixedRawBytes(4));
```
kermandev: Consider storing IPV4 and IPV6 FixedRawBytes types in private static final fields

kermandev (changes requested):

src/main/java/net/minestom/server/network/player/PlayerSocketConnection.java
theminecoder: As noted in discord, handshake listener is too late as the decode needs to happen before packets are constructed due to the extra data in the buffer

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
theminecoder: This is intentional, the v1 header is just a string at the start of the pack that is terminated with \r\n so reading the entire packet to find the end of the string is the quickest way to decode it.

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private static boolean looksLikeV2(NetworkBuffer buffer) {
+        if (buffer.readableBytes() < 12) return false;
+        byte[] sig = buffer.readAt(buffer.readIndex(), V2_SIG_BYTES);
+
+        return Arrays.equals(sig, V2_SIG);
+    }
+
+    private static ClientInfo parseV1(NetworkBuffer buffer, SocketAddress fallback) {
+        byte[] bytes = buffer.readAt(buffer.readIndex(), NetworkBuffer.RAW_BYTES);
```
kermandev: Still would rather this value be capped to the `min(max v1 length, readableBytes);`

theminecoder: Found some documentation that gave me the max bytes in the protocol spec, so have capped it there, should be good to go now!

src/main/java/net/minestom/server/network/player/ProxyProtocolDecoder.java
```diff
@@ -21,7 +21,12 @@
 final class ProxyProtocolDecoder {
 
     private static final byte[] V1_SIG = "PROXY ".getBytes();
+    // Number pulled from https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
+    // "So a 108-byte buffer is always enough to store all the line and a trailing zero
+    // for string processing."
+    private static final int V1_HEADER_MAX_LENGTH = 108;
     private static final NetworkBuffer.Type<byte[]> V1_SIG_BYTES = NetworkBuffer.FixedRawBytes(V1_SIG.length);
+    private static final NetworkBuffer.Type<byte[]> V1_FULL_HEADER = NetworkBuffer.FixedRawBytes(V1_HEADER_MAX_LENGTH);
```
kermandev: FixedRawBytes is fixed as in it will throw if it cant find 108 bytes which is likely not what you meant.

theminecoder: Looking at the impl of the RawBytesType it uses the length provided or the length of readable bytes, whatever is smaller. Works perfectly in testing. If that is unintended then a new type likely needs to be created but that is outside the scope of this PR (and possibly a breaking change)
 https://github.com/Minestom/Minestom/blob/master/src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java#L334-L339

kermandev: Oh sorry, I changed this on my ffm branch and forgot it still had this behavior on master/next

theminecoder:
@kermandev is there a chance this can be pulled in soon if all the discussions are closed, we would like to not have to sit on a local jar for too long if we can avoid it

kermandev:
It will likely be merged eventually, its just a change that requires more attention and testing as we dont have full integration from connection to playable

kermandev (approved):

PR #3131, opened 2026-05-04
fix: Trident/Spear ItemAnimation

vebym:
## Proposed changes

Adds the TRIDENT enum value for the `consumable` data component, added in 1.21.11
see https://minecraft.wiki/w/Data_component_format#consumable

## Types of changes

## Checklist

merging your code._

## Further comments

Should the spear be added to UseItemListener? Seems to be the only one not in it, but it has a different useItemTime for each material

### DIFF
--- a/src/main/java/net/minestom/server/item/ItemAnimation.java
+++ b/src/main/java/net/minestom/server/item/ItemAnimation.java
@@ -9,13 +9,13 @@ public enum ItemAnimation {
     DRINK,
     BLOCK,
     BOW,
-    SPEAR,
+    TRIDENT,
     CROSSBOW,
     SPYGLASS,
     TOOT_HORN,
     BRUSH,
     BUNDLE,
-    TRIDENT;
+    SPEAR;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/item/ItemAnimation.java
kermandev:
```suggestion
    TRIDENT,
```

src/main/java/net/minestom/server/item/ItemAnimation.java
```diff
@@ -14,7 +14,8 @@ public enum ItemAnimation {
     SPYGLASS,
     TOOT_HORN,
     BRUSH,
-    BUNDLE;
+    BUNDLE,
+    TRIDENT;
```
kermandev:
```suggestion
    SPEAR;
```

kermandev (changes requested):
TRIDENT and SPEAR need to trade places to be accurate.

kermandev (approved):

PR #3133, opened 2026-05-08
fix: DisplayContext is missing ON_SHELF

Thojs:
## Proposed changes

While parsing item display entities I've encountered that the DisplayContext enum doesn't match the vanilla one. 
There was already an attempt to solve this with 3b12cb7, but these changes also weren't correct.
With Minecraft 1.21.9 shelves were introduced, and these are also not available in the enum.

## Types of changes

## Checklist

merging your code._

## Further comments

### DIFF
--- a/src/main/java/net/minestom/server/entity/metadata/display/ItemDisplayMeta.java
+++ b/src/main/java/net/minestom/server/entity/metadata/display/ItemDisplayMeta.java
@@ -28,14 +28,15 @@ public class ItemDisplayMeta extends AbstractDisplayMeta {
 
     public enum DisplayContext {
         NONE,
-        THIRDPERSON_LEFT_HAND,
-        THIRDPERSON_RIGHT_HAND,
-        FIRSTPERSON_LEFT_HAND,
-        FIRSTPERSON_RIGHT_HAND,
+        THIRDPERSON_LEFTHAND,
+        THIRDPERSON_RIGHTHAND,
+        FIRSTPERSON_LEFTHAND,
+        FIRSTPERSON_RIGHTHAND,
         HEAD,
         GUI,
         GROUND,
-        FIXED;
+        FIXED,
+        ON_SHELF;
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/entity/metadata/display/ItemDisplayMeta.java
```diff
@@ -28,14 +28,15 @@ public void setDisplayContext(DisplayContext value) {
...
     public enum DisplayContext {
         NONE,
-        THIRDPERSON_LEFT_HAND,
-        THIRDPERSON_RIGHT_HAND,
-        FIRSTPERSON_LEFT_HAND,
-        FIRSTPERSON_RIGHT_HAND,
+        THIRDPERSON_LEFTHAND,
+        THIRDPERSON_RIGHTHAND,
+        FIRSTPERSON_LEFTHAND,
+        FIRSTPERSON_RIGHTHAND,
```
kermandev: Likely not worth renaming these currently to the mojang name

Breaking for little gain

Thojs: Although I do understand your concern, having to replace `hand` with `_hand`, or creating a new enum with the correct values to allow for parsing entity nbt data is a bit of a hack.

It was worth a try. If you still think it's not the right decision, I'll revert these changes.

kermandev: Think its fine to add the fix of the missing value, and then to create a future pr or issue because this likely isnt the only case if you are using the enums name to parse this and its wrong.

kermandev: Looking at it more Mojang uses a seperate name between the enum and the key. Ive created an issue to audit more usages, and moving this PR as a fix to the missing ON_SHELF context.

PR #3134, opened 2026-05-09
Section invalidate

TheMode:
Avoid duplicate lookups, and get a centralized way to invalidate a section

### DIFF
--- a/src/main/java/net/minestom/server/instance/Instance.java
+++ b/src/main/java/net/minestom/server/instance/Instance.java
@@ -304,8 +304,7 @@ public abstract class Instance implements Block.Getter, Block.Setter, Biome.Gett
         final Chunk chunk = getChunk(sectionX, sectionZ);
         if (chunk != null) {
             Section section = chunk.getSection(sectionY);
-            section.skyLight().invalidate();
-            section.blockLight().invalidate();
+            section.invalidate();
             chunk.invalidate();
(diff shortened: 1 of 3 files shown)


kermandev (approved):

DasBabyPixel:
I do want to point out that lighting logic being in the Section class is in fact an anti pattern, since we differentiate between LightingChunk and DynamicChunk.
Because of that for the chunk system I decided to move all lighting logic out of any classes not directly related to lighting, so lighting could be offered as a standalone package with no dependencies of other classes on lighting. For this PR (since Section is final) I'd recommend an `invalidateSection(int sectonY)` to optionally be overridden on DynamicChunk, so custom chunk implementations can add extra invalidation logic.

kermandev:
I think this was answered in the discord but, LightingChunk is just a chunk that can keep recomputing its lighting on the fly, while DynamicChunk's dont have this behavior and require the loader to set their light, or the user to set it.

> I do want to point out that lighting logic being in the Section class is in fact an anti pattern, since we differentiate between LightingChunk and DynamicChunk. Because of that for the chunk system I decided to move all lighting logic out of any classes not directly related to lighting, so lighting could be offered as a standalone package with no dependencies of other classes on lighting. For this PR (since Section is final) I'd recommend an `invalidateSection(int sectonY)` to optionally be overridden on DynamicChunk, so custom chunk implementations can add extra invalidation logic.

PR #3135, opened 2026-05-09
fix: Stateless ClientHandshakePacket

TheMode:
Avoid `ServerProcess` dependency


kermandev (commented):
Seems like a good change, though now multiple packets can be read instead of throwing as soon as soon as possible.

kermandev (approved):
Seems okay for release, if it causes a massive regression we can roll it back

PR #3137, opened 2026-05-09
Properly read block entities in ChunkData

TheMode:
Had to comment out the test as array comparison is currently identity based (and from what I understand may be fixed later on globally)


src/main/java/net/minestom/server/network/packet/server/play/data/ChunkData.java
```diff
@@ -55,20 +55,22 @@ public void write(NetworkBuffer buffer, ChunkData value) {
...
-        final Map<Integer, Block> blockEntities = new HashMap<>();
         final int size = reader.read(VAR_INT);
+        final Map<Integer, Block> blockEntities = HashMap.newHashMap(size);
         for (int i = 0; i < size; i++) {
             final byte xz = reader.read(BYTE);
             final short y = reader.read(SHORT);
-            final BlockEntityType blockEntityType = reader.read(BlockEntityType.NETWORK_TYPE);
+            final BlockEntityType blockEntity = reader.read(BlockEntityType.NETWORK_TYPE);
+            final Block block = Block.fromKey(blockEntity.key());
+            if (block == null) continue;
```
kermandev, line 71: condition likely needs to be moved below the NBT_COMPOUND.

kermandev (commented):
We should just read BlockEntityType unsure if we should do it in this PR, or target this PR to next for that.

In ffm branch I did do something like

`record BlockEntity(int location, BlockEntityType type, CompoundBinaryTag nbt)` which can easily be templated

kermandev (approved):

PR #3142, opened 2026-05-11
fix!: Port network tests

kermandev:
Ports tests from higher branches like ffm and classfile generation into master.

This does require some breaking changes. 

`MessageSignature` is used more in chat packets.
`FixedRawBytes` now throws if the length isnt equivalent during read() like how write was working already (Read used to <= fixed, which is not correct).
`EnumSet` & `FixedBitSet` now properly read and write according to the spec, they used to not pad correctly to be the correct length in bytes.

Fixed some minor bugs and minor additional changes preventing tests from passing from FFM port.


src/main/java/net/minestom/server/adventure/AdventurePacketConvertor.java
```diff
@@ -64,7 +75,7 @@ public static int getBossBarOverlayValue(BossBar.Overlay overlay) {
     public static byte getBossBarFlagValue(Collection<BossBar.Flag> flags) {
         byte val = 0x0;
         for (BossBar.Flag flag : flags) {
-            val |= flag.ordinal();
+            val |= (byte) (1 << flag.ordinal());
```
mworzala, line 78: I havent looked that far, but was this just wrong previously? And what did it impact if so?

src/main/java/net/minestom/server/network/packet/server/play/TeamsPacket.java
```diff
@@ -83,7 +83,7 @@ public void write(NetworkBuffer buffer, CreateTeamAction value) {
             public CreateTeamAction read(NetworkBuffer buffer) {
                 return new CreateTeamAction(buffer.read(COMPONENT), buffer.read(BYTE),
                         buffer.read(NameTagVisibility.NETWORK_TYPE), buffer.read(CollisionRule.NETWORK_TYPE),
-                        NamedTextColor.namedColor(buffer.read(VAR_INT)), buffer.read(COMPONENT), buffer.read(COMPONENT),
+                        AdventurePacketConvertor.getNamedTextColor(buffer.read(VAR_INT)), buffer.read(COMPONENT), buffer.read(COMPONENT),
```
mworzala, line 86: as I learned today when looking at this, its actually any formatting code not just named text color https://mcsrc.dev/1/26.1.2/net/minecraft/ChatFormatting. While fixing, seems worth supporting those also (though becomes a bit more of a break)

src/main/java/net/minestom/server/network/packet/PacketRegistry.java
```diff
@@ -38,6 +42,10 @@ default PacketInfo<T> packetInfo(T packet) {
 
     ConnectionSide side();
 
+    @ApiStatus.Experimental
+    @Unmodifiable
+    List<PacketInfo<? extends T>> packets();
```
mworzala, line 47: kind of unopposed, but whats the use case?

src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
```diff
@@ -702,21 +703,25 @@ public float[] read(NetworkBuffer buffer) {
     // Combinators
 
     record EnumSetType<E extends Enum<E>>(Class<E> enumType,
-                                          E[] values) implements NetworkBufferTypeImpl<EnumSet<E>> {
+                                          E[] values, Type<BitSet> bitSetType) implements Type<EnumSet<E>> {
+        public EnumSetType {
```
mworzala, line 707: should make the bitset type in a 2nd constructor here, the caller doesnt need to know about that

src/main/java/net/minestom/server/network/NetworkBufferTypeImpl.java
```diff
@@ -727,21 +732,28 @@ public EnumSet<E> read(NetworkBuffer buffer) {
         }
     }
 
-    record FixedBitSetType(int length) implements NetworkBufferTypeImpl<BitSet> {
+    record FixedBitSetType(int length, Type<byte[]> arrayType) implements Type<BitSet> {
```
mworzala, line 739: same comment about array type

src/main/java/net/minestom/server/network/packet/server/play/TeamsPacket.java
kermandev: I noticed this too when working on this, but decided to just fix it and we can expand this with a custom enum later.

src/main/java/net/minestom/server/network/packet/PacketRegistry.java
kermandev: Used in the PacketWriteReadTest which requires all packets to be defined with tests

src/main/java/net/minestom/server/adventure/AdventurePacketConvertor.java
kermandev: Yes it was wrong before https://mcsrc.dev/1/26.1.2/net/minecraft/network/protocol/game/ClientboundBossEventPacket, just bossbar packets

PR #3157, opened 2026-05-12
Add RegistriesPacket

TheMode:
Remove ugly lists from `ConnectionManager` ensuring more potential code reuse and better responsibility separation.

Thought about having those in `Registries` directly, no strong opinion either way.


mworzala (commented):
I dont like RegistriesPacket in the same folder as packets with a name that looks like a packet (but its not), I think it would be reasonable to add that as static utils inside registries itself as noted

PR #3160, opened 2026-05-13
feat: Add ChunkData.Section

TheMode:
Avoid duplication for the ChunkData data array serialization.

Not sure I am a fan of this solution, might be better to directly replace the `ChunkData` byte array, or perhaps that `ChunkSectionData` is unnecessary and the serializer should just be defined elsewhere (`Palette`? `ChunkData`?). But the fact that it lacks light information make me think it shouldn't be in `Section` (or at least perhaps with a different name)

### DIFF
--- a/src/main/java/net/minestom/server/instance/DynamicChunk.java
+++ b/src/main/java/net/minestom/server/instance/DynamicChunk.java
@@ -12,7 +12,6 @@ import net.minestom.server.instance.block.BlockHandler;
 import net.minestom.server.instance.heightmap.Heightmap;
 import net.minestom.server.instance.heightmap.MotionBlockingHeightmap;
 import net.minestom.server.instance.heightmap.WorldSurfaceHeightmap;
-import net.minestom.server.instance.palette.Palette;
 import net.minestom.server.network.NetworkBuffer;
 import net.minestom.server.network.packet.server.CachedPacket;
 import net.minestom.server.network.packet.server.SendablePacket;
@@ -29,7 +28,6 @@ import net.minestom.server.utils.ArrayUtils;
 import net.minestom.server.utils.validate.Check;
 import net.minestom.server.world.DimensionType;
 import net.minestom.server.world.biome.Biome;
-import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,7 +35,6 @@ import org.slf4j.LoggerFactory;
 import java.util.*;
 
 import static net.minestom.server.coordinate.CoordConversion.globalToSectionRelative;
-import static net.minestom.server.network.NetworkBuffer.SHORT;
 
 /**
  * Represents a {@link Chunk} which store each individual block in memory.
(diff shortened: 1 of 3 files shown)


mworzala:
light data is effectively a different concept (also a different range since light extends a section above and below the blocks), so this type does to me feel like a "chunk data" section. perhaps as a ChunkData.Section inner record.

I would be very happy to get rid of the raw data array in ChunkData, but its just impossible to write a sane serializer for it because its unreadable without knowing the world height.

DasBabyPixel:
Brings me back to my point that section shouldn't contain light data.

I have an alternative solution for light data, which is much more extensible and can actually represent light correctly in my chunk system branch, if you wanna have a look at it and maybe share thoughts whether it's a good or bad idea

https://github.com/DasBabyPixel/Minestom/blob/71fa2cece0b124321c1bed52e662d870cbed9198/src/main/java/net/minestom/server/instance/Section.java
https://github.com/DasBabyPixel/Minestom/blob/71fa2cece0b124321c1bed52e662d870cbed9198/src/main/java/net/minestom/server/instance/light/LightingChunk.java#L246
https://github.com/DasBabyPixel/Minestom/blob/71fa2cece0b124321c1bed52e662d870cbed9198/src/main/java/net/minestom/server/instance/DynamicChunk.java#L297

src/main/java/net/minestom/server/network/packet/server/play/data/ChunkData.java
```diff
@@ -75,4 +77,32 @@ private static Map<Integer, Block> readBlockEntities(NetworkBuffer reader) {
...
     }
+
+    public record Section(short blockCount, short liquidCount, Palette blockStates, Palette biomes) {
+        private static final int[] LIQUID_STATE_IDS = Block.values().stream()
+                .filter(b -> b.registry().isLiquid())
+                .mapToInt(Block::stateId)
+                .toArray();
+
+
+        public static Section from(net.minestom.server.instance.Section section) {
```
kermandev: Shouldnt be here, calle should perform this

src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -269,14 +266,10 @@ private ChunkDataPacket createChunkPacket() {
...
-            NetworkBuffer.Type<Palette> biomeSerializer = Palette.biomeSerializer(MinecraftServer.getBiomeRegistry().size());
+            NetworkBuffer.Type<ChunkData.Section> sectionSerializer = ChunkData.Section.networkType(MinecraftServer.getBiomeRegistry().size());
             final byte[] data = NetworkBuffer.makeArray(networkBuffer -> {
                 for (Section section : sections) {
-                    final int fillCount = section.blockPalette().count();
-                    networkBuffer.write(SHORT, (short) fillCount);
-                    networkBuffer.write(SHORT, (short) (fillCount > 0 ? 1 : 0)); //TODO(26.1) proper fluid count
-                    networkBuffer.write(Palette.BLOCK_SERIALIZER, section.blockPalette());
-                    networkBuffer.write(biomeSerializer, section.biomePalette());
+                    networkBuffer.write(sectionSerializer, new ChunkData.Section((short) section.blockPalette().count(), (short) 0, section.blockPalette(), section.biomePalette()));
```
kermandev: 0 is incorrect for water worlds, it will cause the player to fall through the world, you likely still need to count these, maybe making a helper or calculating it manually.

kermandev (approved):

PR #3161, opened 2026-05-13
Add data component reset

TheMode:
* Add `DataComponentMap#reset`
* Add `ItemStack#components` which return the material prototype patched with the custom components
* Intern DataComponentMap into EMPTY
* ~Cache `ItemStack` hash and hashCode, both with a benign data race to avoid read overhead. Which unfortunately required moving away from record.~

~Item comparison could perhaps first compare the hashCodes as a fast exit, unsure if worth the trouble.~

### DIFF
--- a/src/main/java/net/minestom/server/component/DataComponentMap.java
+++ b/src/main/java/net/minestom/server/component/DataComponentMap.java
@@ -9,7 +9,6 @@ import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.Collection;
-import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.IntFunction;
 
@@ -145,12 +144,6 @@ public sealed interface DataComponentMap extends DataComponent.Holder permits Da
 
     default DataComponentMap set(DataComponent<Unit> component) {
         return set(component, Unit.INSTANCE);
-    }
-
-    default DataComponentMap with(Consumer<PatchBuilder> consumer) {
-        final PatchBuilder builder = toPatchBuilder();
-        consumer.accept(builder);
-        return builder.build();
     }
 
     /**
--- a/src/main/java/net/minestom/server/component/DataComponentMap.java
+++ b/src/main/java/net/minestom/server/component/DataComponentMap.java
@@ -88,7 +88,7 @@ public sealed interface DataComponentMap extends DataComponent.Holder permits Da
             }
         }
 
-        return diff.isEmpty() ? EMPTY : new DataComponentMapImpl(diff);
+        return DataComponentMapImpl.fromMap(diff);
     }
 
     static DataComponentMap applyPatch(DataComponentMap prototype, DataComponentMap patch) {
@@ -104,7 +104,7 @@ public sealed interface DataComponentMap extends DataComponent.Holder permits Da
                 result.put(entry.getIntKey(), entry.getValue());
             }
         }
-        return result.isEmpty() ? EMPTY : new DataComponentMapImpl(result);
+        return DataComponentMapImpl.fromMap(result);
     }
 
     boolean isEmpty();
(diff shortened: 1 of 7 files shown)


src/main/java/net/minestom/server/component/DataComponentMapImpl.java
```diff
@@ -117,7 +125,7 @@ public <T> Builder set(DataComponent<T> component, T value) {
 
         @Override
         public DataComponentMap build() {
-            return new DataComponentMapImpl(new Int2ObjectArrayMap<>(components));
+            return components.isEmpty() ? DataComponentMap.EMPTY : new DataComponentMapImpl(new Int2ObjectArrayMap<>(components));
```
mworzala: would prefer a fromMap to avoid duplicating this ternary 6 times

src/main/java/net/minestom/server/component/DataComponentMap.java
```diff
@@ -138,6 +154,8 @@ default DataComponentMap set(DataComponent<Unit> component) {
      */
     DataComponentMap remove(DataComponent<?> component);
 
+    DataComponentMap reset(DataComponent<?> component);
```
mworzala, line 184: needs a doc explaining the difference between reset and remove (and in particular that reset is only relevant for patches)

src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -196,6 +199,15 @@ default <T> ItemStack with(DataComponent<T> component, UnaryOperator<T> operator
     @Contract(value = "_, -> new", pure = true)
     ItemStack without(DataComponent<?> component);
 
+    /**
+     * Returns a new ItemStack with the given component reset to the material default.
+     *
+     * @param component The component to reset
+     * @return A new ItemStack without an explicit patch for the given component
```
mworzala: nitpick but override may be a more obvious term than patch here

src/main/java/net/minestom/server/item/ItemStackImpl.java
```diff
@@ -126,6 +150,11 @@ public ItemStack without(DataComponent<?> component) {
         return new ItemStackImpl(material, amount, components.remove(component));
     }
 
+    @Override
+    public ItemStack reset(DataComponent<?> component) {
+        return new ItemStackImpl(material, amount, components.reset(component));
```
mworzala: perhaps worth handling the case where the component map doesnt change and returning this

src/main/java/net/minestom/server/item/ItemStack.java
```diff
@@ -127,6 +127,9 @@ static ItemStack fromItemNBT(CompoundBinaryTag nbtCompound) {
     @Contract(pure = true)
     DataComponentMap componentPatch();
 
+    @Contract(pure = true)
+    DataComponentMap components();
```
mworzala, line 149: deserves a comment that its returning the resolved/absolute set of components not the patch (may be confusing that you set a component and then get components and its not just that one)

kermandev (commented):
`ItemStack#components` and `Itemstack#reset` seem nice to have, though I dont think its worth caching the hashcode and `ItemStack.Hash` as I've never seen be an issue on a profile report and prohibits value classes in its first iteration. It would make more sense to cache the hashCode of custom immutable `Int2ObjectArrayMap` implementation for `DataComponentMap` though then again it seems overkill

I dont see the usecase for `DataComponentMap#reset` as it seems to just be a remove. On the `PatchBuilder` it makes more sense, but I wonder if its an antipattern to have it in a builder.

mworzala (commented):
apparently I forgot to submit review.

I agree with what kerman said re caching, id prefer to wait until someone has an actual use case where its a problem then consider it.

@kermandev re reset its a different operation than remove for patch DataComponentMaps (which includes ItemStack components) since its applied on top of the material.

eg if I `#remove` ITEM_NAME on a stick item stack, the item will have no name. If I `#reset` it, the item will get back `Stick` as its name.

kermandev:
Hashcode for ItemStack doesnt seem like an issue currently, but ItemSack.Hash is registries dependent. I think we should avoid caching internally because different Registries can be used to create the Hash. 

TLDR Revisit caching ItemStack hashing in the future

PR #3168, opened 2026-05-14
fix: Potential security issues

TheMode:
* Crypto secure random
* Proper url param encoding
* Avoid potential zip bomb


kermandev (approved):
Seems fine to have

src/main/java/net/minestom/server/utils/mojang/MojangUtils.java
```diff
@@ -103,21 +103,26 @@ public static String getUsername(UUID playerUUID) throws IOException {
...
         } else {
-            url = String.format(BASE_AUTH_URL, username, serverId);
+            url = String.format(BASE_AUTH_URL, username, encodedServerId);
         }
 
         return retrieve(url);
     }
 
+    private static String encode(String value) {
+        return URLEncoder.encode(value, StandardCharsets.UTF_8);
```
kermandev, line 133: `URLEncoder#encode` is only good for query encoding, not path encoding. Those would have to be validated separately, and likely we dont need it for UUID's

kermandev (changes requested):

kermandev (approved):

PR #3169, opened 2026-05-14
Make auth code more testable and isolated

TheMode:
Bit sensitive. Goal is to avoid messy state in the socket connection and the packet listener.

Not particularly fond of encryption


TheMode:
Potentially waiting on #3208

PR #3170, opened 2026-05-14
OpenLAN update

TheMode:
* Fix warnings
* Avoid fields mess

The API probably need to be reworked, but this is already better


src/main/java/net/minestom/server/extras/lan/OpenToLAN.java
```diff
@@ -57,19 +61,18 @@ public static boolean open() {
...
         try {
             socket = new DatagramSocket(config.port);
         } catch (SocketException e) {
             LOGGER.warn("Could not bind to the port!", e);
             return false;
         }
-
-        eventCooldown = new Cooldown(config.delayBetweenEvent);
-        task = MinecraftServer.getSchedulerManager().buildTask(OpenToLAN::ping)
+        final Task task = MinecraftServer.getSchedulerManager().buildTask(OpenToLAN::ping)
```
kermandev: Unsure why this wouldnt just be simplified to not include this

kermandev:
I agree the current API isnt great, but I dont think it is worth committing work to fixing this if the api changes. Likely this class doesnt even need to depend on MinecraftServer at all and can be dependency free

PR #3172, opened 2026-05-15
Add more packet reading/writing utils

TheMode:
Isolate more packet handling code


kermandev:
This doesnt seem like a good idea to move the logic into `PacketWriting`/`PacketReading`, whilst yes it may make it reusable outside of `PlayerSocketConnection` it seems worse to delegate a method (your coupling for little gain). With `SendablePacket` being an optimization too, it seems even worse to make `PacketWriting` worry about it. This is also why I was moving towards deleting `PacketWriting#writeQueue` as its logic should not be that complex.

For `PacketReading` it also doesnt make sense to couple this simple logic by ignoring the result

PR #3175, opened 2026-05-16
fix: Particle's Blockstate codec

Sybsuper:
### The problem
When using biome ambient particles with blockstates in Minestom. Clients would disconnect with a network error. To trigger this unintended behaviour, cherry pick commit `0b6a7bdd` without he fix commit and run the command `/testbiomeambientparticle`.
Error log:
```
Description: Registry Loading
...
Details:
	Errors: 
		minecraft:worldgen/biome/minecraft:testbiome: Failed to parse value {attributes:{"minecraft:visual/ambient_particles":[{particle:{block_state:"minecraft:copper_bulb[lit=true,powered=false]",type:"minecraft:block_marker"},probability:5.0E-4f}]},downfall:0.4f,effects:{water_color:"#3F76E4"},has_precipitation:1b,temperature:0.8f} from server
```

### Why?
Because the [specs](https://minecraft.wiki/w/Particle_format#Configurations_of_particle_types) describe particle BlockData should be a map of "Name" and "Properties", not a string.

### The fix
Implement a `PARTICLE_BLOCK_STATE_CODEC` that encodes/decodes BlockData according to the particle spec.

### How to test?
Checkout the PR, run the example server and run the `/testbiomeambientparticle` command and observer the client not disconnecting with an error, but instead showing the particles as intended.

My result:

https://github.com/user-attachments/assets/e1035a79-5888-495f-baf2-64fc38e13133

## Types of changes

## Checklist

merging your code._

  - See commit `0b6a7bdd` which throws an error before the fix commit and does not anymore after the fix commit;
  - See `ParticleBlockDataCodecTest` which tests `encode(decode(x)) == x` equivalence and the handling of default valued properties.

## Further comments

Feel free to discard/revert the commit adding the test command (`0b6a7bdd`) if deemed too low quality or irrelevant.


Bloeckchengrafik (approved):
I'm unsure if that codec isn't already implemented elsewhere, if it is, it should be used from there. But other than that: works great!

Sybsuper:
> I'm unsure if that codec isn't already implemented elsewhere, if it is, it should be used from there. But other than that: works great!

Not that I could find, I did find however that the `AnvilLoader#blockStateNbtCompute` private method is very similar to the encoding method I wrote (but it's an NBT specific implementation). From that method it seems like it's "normal behavior" to skip default valued properties. So I implemented that as well.

Bloeckchengrafik (approved):
I think this is good then

src/main/java/net/minestom/server/particle/Particle.java
```diff
@@ -65,6 +66,53 @@ public <D> Result<D> encode(Transcoder<D> coder, @Nullable Particle value) {
         }
     };
 
+    Codec<net.minestom.server.instance.block.Block> PARTICLE_BLOCK_STATE_CODEC = new Codec<>() {
```
mworzala: move this to Block, this isn't a particle-specific format

mworzala (commented):
looks good overall

src/main/java/net/minestom/server/particle/Particle.java
Sybsuper: Done, I called it `BLOCK_STATE_MAP_CODEC` to not be confused with the already existing `STATE_CODEC` field (also added some small doc-string to explain the difference. Lemme know if you can think of a better name.

src/main/java/net/minestom/server/instance/block/Block.java
```diff
@@ -37,9 +39,64 @@ public sealed interface Block extends StaticProtocolObject<Block>, TagReadable,
...
+     * Format: <code>"minecraft:x[a=y,b=z]"</code>
+     */
     Codec<Block> STATE_CODEC = Codec.STRING.transform(state -> Objects.requireNonNull(
             Block.fromState(state), () -> "not a block state: " + state), Block::state);
 
+    /**
+     * Codec for block states as a map.
+     * Format: <code>{Name:"minecraft:x",Properties:{a:"y",b:"z"}}</code>
+     */
+    Codec<Block> BLOCK_STATE_MAP_CODEC = new Codec<>() {
```
kermandev: Should be a `StructCodec` because of the map usage. with the field renamed to `STATE_STRUCT_CODEC`

Then use `decodeFromMap` and `encodeToMap`

Note for @mworzala: Wonder if put/get should be public in StructCodec so you get better error handling here

Sybsuper: Done!

kermandev (approved):

PR #3178, opened 2026-05-17
feat: Add 26.1 support for AnvilLoader

CodeDoctorDE:
## Proposed changes

This PR updates the Anvil chunk loader to support the new Minecraft Java 26.1 world storage layout.

Minecraft 26.1 stores dimension data in namespaced folders under `dimensions/<namespace>/<dimension>/`, so the loader now resolves region files from the correct dimension path instead of always assuming the legacy root-level `region` folder.

The default constructor continues to load the Overworld, while additional constructors allow loading other vanilla or custom dimensions. The implementation also keeps backwards compatibility with older world layouts such as root `region`, `DIM-1`, and `DIM1`.

## Types of changes

## Checklist

merging your code._

## Further comments

The main goal of this change is to make the Anvil loader dimension-aware while preserving the existing behavior for older worlds.

Instead of hardcoding the region path to `<world>/region`, the loader now resolves the dimension path first. For Minecraft 26.1+ worlds this means loading from paths like:

```text
<world>/dimensions/minecraft/overworld/region

### DIFF
--- a/src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
+++ b/src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
@@ -5,6 +5,7 @@ import it.unimi.dsi.fastutil.ints.IntList;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import it.unimi.dsi.fastutil.longs.LongOpenHashSet;
 import it.unimi.dsi.fastutil.longs.LongSet;
+import net.kyori.adventure.key.Key;
 import net.kyori.adventure.nbt.*;
 import net.minestom.server.MinecraftServer;
 import net.minestom.server.adventure.MinestomAdventure;
@@ -27,10 +28,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.StandardCopyOption;
-import java.nio.file.StandardOpenOption;
+import java.nio.file.*;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/main/java/module-info.java [hunk 1]
```diff
@@ -19,6 +19,7 @@
     requires java.desktop;
     requires java.management;
     requires net.minestom.data;
+    requires net.kyori.adventure.key;
```
kermandev: extraneous

kermandev (changes requested):

src/main/java/module-info.java
CodeDoctorDE: important to specify the dimension in the loader :)

kermandev: ?

The requires statement is redundant because `net.kyori.adventure.api` includes it

CodeDoctorDE: Oh okay, it was added by intellij

CodeDoctorDE: fixed

kermandev:
After some internal discussion, we have decided to not add this to Minestom.

Why? Minestom never handled this case to begin with, as developers would pass a custom `Path`, this would be adding API moving control from the user to Minestom in which we try to avoid. Also, Minestom has never been fond about maintaining previous version formats, and this API would likely be removed in a future release.

Sorry about wasting your time on this PR.

CodeDoctorDE:
I strongly disagree here. Having 26.1 support for anvilloader is important for everyone using Minestom in the future.
If backwards compatibility is not desired, i can also remove this part of the code. Then the only change is the additional dimension key parameter. This code also doesnt allow custom paths so I dont see the argument here.

Please rethink this decision, having only support for older world formats before 26.1 on 26.1 protocol version doesnt really make sense

kermandev (approved):

src/main/java/net/minestom/server/instance/anvil/AnvilLoader.java
```diff
@@ -59,12 +58,33 @@ public class AnvilLoader implements ChunkLoader {
...
 
+    /**
+     * Creates a new AnvilLoader for the given world path and dimension.
+     * @param path The path to the world
+     * @param dimension The key for the dimension. Use {@link DimensionType} for getting vanilla keys for dimensions.
+     */
+    public AnvilLoader(Path path, Key dimension) {
+        this.path = path;
+        this.levelPath = path.resolve("level.dat");
+        this.regionPath = path.resolve("dimensions").resolve(dimension.namespace()).resolve(dimension.value());
```
kermandev: Are you missing  a `.resolve("region");` at the end of this?

CodeDoctorDE: oh yeah

CodeDoctorDE: done

kermandev (approved):

PR #3181, opened 2026-05-17
feat: Cleanup gitignore

TheMode:
Was a bit of a mess. Also included some LLM agents path.

Some path still seem redundant/legacy but just in case not removed.


kermandev (approved):
Seems like a fine change to change

kermandev:
probably should have been refactor: oops

PR #3184, opened 2026-05-18
Add applyRegistryDataPacket

TheMode:
Add a new way to update a `Registries` from received server packet, useful when making a client or proxy.


kermandev:
Instead of updating a Registries, why not have a custom Registries for each client?

PR #3190, opened 2026-05-23
add AGENTS.md + CLAUDE.md import

TheMode:
Very barebone AI coding assistant files. Might be a fairly contentious change. Been generated through claude `/init` command and then manually reviewed for obvious flaws.


kermandev:
Contentious is correct. If we ignore the maintenance burden of keeping it up to date. It is still likely not required as models are constantly changing and each model frames it differently. Though the biggest reason is we don’t want an influx of slop issues and PRs where authors are posing as a human. Though Minestom does not have an official policy on AI usage currently.

Maybe I change my view in the future but it’s likely these tools will become too expensive to use to justify adding this then.

PR #3195, opened 2026-05-27
feat: Builders for DimensionType, Biome, EnvironmentAttributeMap

thecodertommy:
## Proposed changes

When creating custom biomes or dimension types, it is often convenient to start with an existing one. This PR adds new `builder()` methods that take existing objects and create a Builder based on it.

## Types of changes

## Checklist


kermandev (changes requested):
Seems fine but a test could be added for the new API

Likely checking if passing a biome or map through actually provides the same result when built.

kermandev (approved):

PR #3196, opened 2026-05-28
fix: ChunkData, DeclareCommands & Ingredent packet parsing issue

TheMode:
Fix various packet parsing issues. Make me able to join a paper server using minestom api as proxy.


src/main/java/net/minestom/server/recipe/Ingredient.java
```diff
@@ -13,53 +13,56 @@
...
-    public static final NetworkBuffer.Type<Ingredient> NETWORK_TYPE = NetworkBufferTemplate.template(
-            // FIXME(1.21.2): This is really an ObjectSet, but currently ObjectSet does not properly support
-            //  non-dynamic registry types. We need to improve how the tag system works generally.
-            new NetworkBuffer.Type<>() {
-                @Override
-                public void write(NetworkBuffer buffer, List<Material> value) {
-                    // +1 because 0 indicates that an item tag name follows (in this case it does not).
-                    buffer.write(VAR_INT, value.size() + 1);
-                    for (Material material : value) {
+public sealed interface Ingredient {
```
kermandev: This change is more breaking and is duplicated code. This type seems to actually be a `RegistryTag` consider using that instead.

kermandev (approved):

PR #3206, opened 2026-05-31
refactor: Inline metatadata map suppliers

TheMode:
createMetaMap didn't seem necessary and involve unnecessary copies


src/main/java/net/minestom/server/entity/MetadataHolder.java
```diff
@@ -172,177 +174,169 @@ public Map<Integer, Metadata.Entry<?>> getEntries() {
         return Map.copyOf(this.entries);
     }
 
-    static final Map<String, BiFunction<@Nullable Entity, MetadataHolder, EntityMeta>> ENTITY_META_SUPPLIER = createMetaMap();
+    static final Map<String, BiFunction<Entity, MetadataHolder, ? extends EntityMeta>> ENTITY_META_SUPPLIER = Map.ofEntries(
```
kermandev: Regression on generic definition

`@Nullable` ~~and maybe dont need `? extends`~~

PR #3207, opened 2026-05-31
fix: detect eye level fluid in updatepose to enable swimming pose

vearendev:
## Proposed changes

`player.getPose()` never returns `EntityPose.SWIMMING` because
`setSwimming()` was never called anywhere. Fixed by checking sprint
state + eye level fluid in `updatePose()`

## Types of changes

## Checklist

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -830,8 +830,20 @@ public class Player extends LivingEntity implements CommandSender, HoverEventSou
         EntityPose oldPose = getPose();
         EntityPose newPose;
 
-        // Figure out their expected state
         var meta = getEntityMeta();
+
+        if (instance != null) {
+            Pos pos = getPosition();
+            int eyeBlockX = pos.blockX();
+            int eyeBlockY = (int) Math.floor(pos.y() + getEyeHeight());
+            int eyeBlockZ = pos.blockZ();
+            Block eyeBlock = instance.getBlock(eyeBlockX, eyeBlockY, eyeBlockZ, Block.Getter.Condition.TYPE);
+            boolean eyeInWater = eyeBlock.compare(Block.WATER)
+                    || "true".equals(instance.getBlock(eyeBlockX, eyeBlockY, eyeBlockZ).getProperty("waterlogged"));
+            meta.setSwimming(eyeInWater && isSprinting() && getVehicle() == null);
+        }
+
+        // Figure out their expected state
         if (meta.isFlyingWithElytra()) {
             newPose = EntityPose.FALL_FLYING;
         } else if (meta instanceof LivingEntityMeta livingMeta && livingMeta.getBedInWhichSleepingPosition() != null) {
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -836,13 +836,12 @@ public class Player extends LivingEntity implements CommandSender, HoverEventSou
             Pos pos = getPosition();
             boolean shouldSwim;
             if (meta.isSwimming()) {
-                Block feetBlock = getBlockSafe(pos.blockX(), pos.blockY(), pos.blockZ());
-                shouldSwim = isBlockInWater(feetBlock, 0.0) && isSprinting() && getVehicle() == null;
+                shouldSwim = isInWater() && isSprinting() && getVehicle() == null;
             } else {
                 double eyeY = pos.y() + getEyeHeight();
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -830,8 +830,20 @@ protected void updatePose() {
...
         var meta = getEntityMeta();
+
+        if (instance != null) {
+            Pos pos = getPosition();
+            int eyeBlockX = pos.blockX();
+            int eyeBlockY = (int) Math.floor(pos.y() + getEyeHeight());
+            int eyeBlockZ = pos.blockZ();
+            Block eyeBlock = instance.getBlock(eyeBlockX, eyeBlockY, eyeBlockZ, Block.Getter.Condition.TYPE);
+            boolean eyeInWater = eyeBlock.compare(Block.WATER)
+                    || "true".equals(instance.getBlock(eyeBlockX, eyeBlockY, eyeBlockZ).getProperty("waterlogged"));
```
kermandev: Likely better to not do the double `getBlock` call if you end up checking the properties regardless. There is likely better optimizations possible with an fast liquidCount in the future.

kermandev (commented):
Seems to work fine on general case (3x3 of water) the demo server and viewers can see the change. Used `/testinstabreak`, though it seems to break with the floating water as it tries to set the player to swim which causes a bobbing affect.

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -830,8 +830,23 @@ protected void updatePose() {
         EntityPose oldPose = getPose();
         EntityPose newPose;
 
-        // Figure out their expected state
         var meta = getEntityMeta();
+
+        if (instance != null) {
+            Pos pos = getPosition();
```
mworzala, line 836: you should never be swimming if you are flying or in spec (should be forced out of swimming state if you were before), should have a test also

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -861,6 +876,49 @@ protected void updatePose() {
...
+        if ("true".equals(block.getProperty("waterlogged"))) {
+            return localY <= 8.0 / 9.0;
+        }
+        return false;
+    }
+
+    private boolean isInWater() {
+        Pos pos = getPosition();
+        double absMinY = pos.y() + boundingBox.minY();
+        var iter = boundingBox.getBlocks(pos);
```
mworzala: you should be using getBoundingBox() so it accounts for the current pose

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -861,6 +876,49 @@ protected void updatePose() {
...
+        try {
+            return instance.getBlock(x, y, z);
+        } catch (NullPointerException ignored) {
+            return null;
+        }
+    }
+
+    private boolean isBlockInWater(@Nullable Block block, double localY) {
+        if (block == null) return false;
+        if (block.compare(Block.WATER)) {
```
mworzala: the fluid height of any water block is 1.0 if the block above it is also water/waterlogged, this doesnt account for that

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -861,6 +876,49 @@ protected void updatePose() {
...
+        var iter = boundingBox.getBlocks(pos);
+        while (iter.hasNext()) {
+            var blockPos = iter.next();
+            int by = blockPos.blockY();
+            Block block = getBlockSafe(blockPos.blockX(), by, blockPos.blockZ());
+            if (block == null) continue;
+            double fluidHeight;
+            if (block.compare(Block.WATER)) {
+                int level = Integer.parseInt(block.getProperty("level"));
+                fluidHeight = (9 - (level & 7)) / 9.0;
```
mworzala: I believe this and the other cases above should be 8- not 9-. 

I havent re-found the vanilla logic for this to confirm but please check (I did it as 8 on hc and pretty sure its right)
https://github.com/hollow-cube/mapmaker/blob/main/modules/common/src/main/java/net/hollowcube/common/extensions/ExtendedPlayer.java#L98

mworzala:
vanilla additionally requires that your feet be in water to start swimming which this does not cover
https://mcsrc.dev/1/26.1.2/net/minecraft/world/entity/Entity#L1569

In particular the last condition of checking the fluid state at the position (feet)

mworzala:
the & 7 is wrong I think. I believe all the >8s are full height for falling water and should be clamped to 8 rather than cycling down to lower levels again

PR #3209, opened 2026-05-31
Replace unsafe with FFM MemorySegment

TheMode:
Non-breaking unlike #2958

Additionally fixed a few nullability warnings.


src/main/java/net/minestom/server/network/NetworkBufferImpl.java
```diff
@@ -10,59 +10,51 @@
...
-    private BinaryTagReader nbtReader;
+    private @Nullable BinaryTagWriter nbtWriter;
+    private @Nullable BinaryTagReader nbtReader;
 
     final @Nullable AutoResize autoResize;
     @Nullable Registries registries;
 
-    ByteBuffer nioBuffer = null;
+    // Cached ByteBuffer view over the whole segment, invalidated when the segment changes.
+    private @Nullable ByteBuffer nioBuffer = null;
```
kermandev: Likely not required, same with `readOnly` and `capacity`

kermandev (approved):

kermandev (changes requested):
Need to store the arena per buffer too, also likely need to cycle the arena unless you want unreachable segments still allocated.

kermandev (approved):

PR #3225, opened 2026-06-12
Entity selector fix

mudkipdev:
- Filters now apply before `@p`/`@r`/`@n` selection, so `@p[gamemode=creative]` returns the nearest creative player instead of nothing
- `@p`/`@n` are now implemented as vanilla shorthands for `sort=nearest,limit=1` and `@r` for `sort=random,limit=1`, so explicit options like `@p[sort=furthest]` and `@r[limit=3]` override them instead of being ignored
- `@r` no longer throws when no players match; it returns an empty list
- `distance` is now a double range like vanilla (`distance=1.5..3.5` works), fixing int overflow on `distance=5..` and broken open lower bounds like `distance=..10`
- Negative, `NaN`, and `Infinity` distance values are rejected at parse time like vanilla
- Junk and dangling selector options (`@e[a=b,c]`, `@e[a=b,]`, `@e[garbage]`) are rejected instead of silently accepted
- `sort=nearest`/`sort=furthest` actually sort now (the old comparator violated the comparator contract)
- `sort=random` no longer throws `UnsupportedOperationException`, and shuffles before `limit` so `sort=random,limit=N` picks a random subset
- `limit=1` (the `@p`/`@r`/`@n` default) uses a single O(n) pass instead of copying and sorting all candidates

Breaking


mworzala (commented):
would like to have some more tests around
* nearest entity
* sort nearest and limit

kermandev (approved):

PR #3233, opened 2026-06-19
Perform next changes in NetworkBuffer

kermandev:
## Proposed changes

- `NetworkBuffer#wrap(byte[], ...)` deleted copy on wrap (Now mutates param1)
- `NetworkBuffer#copyTo(long, byte[], long, long)` deleted in favor of `NetworkBuffer#copyTo(long, byte[], int, int)`
- `NetworkBuffer#readOnly` now returns a read only view instead of setting a flag. They share the same memory region, so changes from the parent should show on the child when it comes to writing to the buffer. Dummy buffers that are read only will throw.
- `NetworkBuffer#writeChannel(SocketChannel)` now uses `WritableByteChannel`
- `NetworkBuffer#ensureWritable(long)` now requires lengths to be positive
- `NetworkBufferImpl` assertReadOnly may no longer throw UnsupportedOperation and can throw IllegalArgumentException instead for read only checks.
- `NetworkBufferImpl` dummy and read checks have been moved to be more hoisted.
- Some write and read operations that threw `UnsupportedOperationException` now throw `IllegalArgumentException` to align with MemorySegment read only exceptions.

## Types of changes

## Checklist

merging your code._

## Further comments
Currently targeting master, as it may be an idea to switch the head nearing the merge window (merge queue)

### DIFF
--- a/src/main/java/net/minestom/server/network/NetworkBufferImpl.java
+++ b/src/main/java/net/minestom/server/network/NetworkBufferImpl.java
@@ -166,9 +166,8 @@ final class NetworkBufferImpl implements NetworkBuffer {
     }
 
     @Override
-    public void readOnly() {
-        final MemorySegment segment = this.segment;
-        if (segment != null) this.segment = segment.asReadOnly();
+    public NetworkBuffer readOnly() {
+        return new NetworkBufferImpl(this.segment.asReadOnly(), this.readIndex, this.writeIndex, null, this.registries);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


src/main/java/net/minestom/server/network/NetworkBufferImpl.java
```diff
@@ -182,9 +170,10 @@ public long capacity() {
     }
 
     @Override
-    public void readOnly() {
+    public NetworkBuffer readOnly() {
         final MemorySegment segment = this.segment;
-        if (segment != null) this.segment = segment.asReadOnly();
+        if (segment == null) return new NetworkBufferImpl(null, this.readIndex, this.writeIndex, null, this.registries);
```
kermandev: Probably could assertDummy here if that would be better usage as readOnly is only really used for reading

kermandev: Gone ahead and did this

PR #3238, opened 2026-06-20
fix/entity-passengers-update

nnikitochka:
## Proposed changes

This PR fixes an issue where passengers riding an entity would not be displayed correctly to a newly added viewer. Previously, `updateNewViewer` relied entirely on `EntityView` to handle passenger visibility, but the `SetPassengersPacket` was never actually sent to a new viewer when it started observing an entity that already had passengers. As a result, players who came into view range of (or were otherwise added as a viewer to) an entity with existing passengers would not see those passengers correctly until some other update triggered a resend of the packet.

The fix explicitly sends a `SetPassengersPacket` to the new viewer at the end of `updateNewViewer` whenever the entity currently has passengers, ensuring the client receives the correct passenger state immediately upon becoming a viewer.

## Types of changes

## Checklist

merging your code._

## Further comments

The fix is minimal and targeted: instead of reworking how `EntityView` manages passenger visibility, it simply ensures the existing `SetPassengersPacket` is sent as part of the new-viewer initialization sequence, alongside the spawn packet, metadata, leash, and head rotation packets that are already sent there. This keeps the change small and low-risk while resolving the visual desync for passengers when a new viewer is added.


nnikitochka:
Я даун

PR #3239, opened 2026-06-20
Acquirable avoid varargs

TheMode:
Avoid potential varargs allocation for this hot path


kermandev (approved):
Not convinced that it would be an issue if its hot enough, though the fix is likely less code in general.

Parts of Check seem like a footgun for hot code it seems

Issue #3243, opened 2026-06-22
[Bug]: Player#sendNotification(Notification) not sending the correct icon


hapily04:
### Affected Version

latest master

### Describe the bug

The icon isn't sending properly for the notification

### Steps to reproduce the bug

1. join the game
2. shift
3. notice how the icon isn't what was specified

### Code sample

```java
package net.minestom.demo;

import net.kyori.adventure.text.Component;
import net.minestom.server.Auth;
import net.minestom.server.MinecraftServer;
import net.minestom.server.advancements.FrameType;
import net.minestom.server.advancements.Notification;
import net.minestom.server.coordinate.Pos;
import net.minestom.server.event.GlobalEventHandler;
import net.minestom.server.event.player.*;
import net.minestom.server.instance.InstanceContainer;
import net.minestom.server.instance.block.Block;
import net.minestom.server.item.ItemStack;
import net.minestom.server.item.Material;


public class MinimalRepro {

    static void main() {
        MinecraftServer server = MinecraftServer.init(new Auth.Online());

        InstanceContainer container = MinecraftServer.getInstanceManager().createInstanceContainer();
        container.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.STONE));
        
        GlobalEventHandler node = MinecraftServer.getGlobalEventHandler();
        node.addListener(AsyncPlayerConfigurationEvent.class, event -> {
           event.setSpawningInstance(container);
           event.getPlayer().setRespawnPoint(new Pos(0, 42, 0));
        });
        node.addListener(PlayerStartSneakingEvent.class, event -> {
            event.getPlayer().sendNotification(new Notification(Component.text("text"), FrameType.GOAL, ItemStack.of(Material.GRASS_BLOCK)));
        });
        server.start("0.0.0.0", 25566);
    }

}

```

Bloeckchengrafik:
To share my current triage findings:

The client interprets the DisplayInfo wrong: 
<img width="1111" height="88" alt="Image" src="https://github.com/user-attachments/assets/48a24bfe-55d3-47be-8733-5e39ee581572" />

ID 27 is grass block here, 1 is stone, so the id and count are swapped. What I don't yet know is whether that happens on more places or if this is just for this here

Bloeckchengrafik:
Found the fix, will PR

PR #3245, opened 2026-06-23
fix: use ItemStackTemplate for DisplayData icon in AdvancementsPacket

Bloeckchengrafik:
## Proposed changes

As described in #3243, the sent varint ordering was wrong for advancement packets. This fixes it by using ItemStackTemplate network type as vanilla is doing as well. (see https://mcsrc.dev/1/26.1.2/net/minecraft/advancements/DisplayInfo#L110 for confirmation)

closing  #3243

## Types of changes

## Checklist

merging your code._

- [ ] I have added tests that prove my fix is effective or that my feature works (how? this is client packet parsing logic for e2e which would be preferred here and I think we don't have such a setup right now)

### DIFF
--- a/src/main/java/net/minestom/server/network/packet/server/play/AdvancementsPacket.java
+++ b/src/main/java/net/minestom/server/network/packet/server/play/AdvancementsPacket.java
@@ -4,6 +4,7 @@ import net.kyori.adventure.text.Component;
 import net.minestom.server.advancements.FrameType;
 import net.minestom.server.adventure.ComponentHolder;
 import net.minestom.server.item.ItemStack;
+import net.minestom.server.item.ItemStackTemplate;
 import net.minestom.server.network.NetworkBuffer;
 import net.minestom.server.network.NetworkBufferTemplate;
 import net.minestom.server.network.packet.server.ServerPacket;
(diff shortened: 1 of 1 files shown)


src/main/java/net/minestom/server/network/packet/server/play/AdvancementsPacket.java
```diff
@@ -134,7 +135,7 @@ public record DisplayData(Component title, Component description,
             public void write(NetworkBuffer buffer, DisplayData value) {
                 buffer.write(NetworkBuffer.COMPONENT, value.title);
                 buffer.write(NetworkBuffer.COMPONENT, value.description);
-                buffer.write(ItemStack.NETWORK_TYPE, value.icon);
+                buffer.write(ItemStackTemplate.NETWORK_TYPE, value.icon);
```
kermandev, line 138: Needs to be updated in read too (read should mostly match write)

Bloeckchengrafik: True, thanks. Done

kermandev (approved):

PR #3247, opened 2026-06-24
Misc MetadataHolder improvements

TheMode:
* Update `#entries` to an array, using its index as a key. Avoid the various `Map` fields, but there is a risk of oversized array, with most metadata definitions holding between 20-40 potential entries. Improving performance at the cost of memory (metadata retrieval can take about 40% of entity collision resolution)
* Change `ENTITY_META_SUPPLIER` to store `EntityType` as key instead of namespace. It caused `minecraft:potion` to be removed, unsure if this was forgotten from a previous version.
* The supplier change required updating most specialized entity meta classes to take a nullable entity (as specified in super constructor)

### DIFF
--- a/src/main/java/net/minestom/server/entity/MetadataHolder.java
+++ b/src/main/java/net/minestom/server/entity/MetadataHolder.java
@@ -1,7 +1,6 @@
 package net.minestom.server.entity;
 
-import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
-import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
+import it.unimi.dsi.fastutil.ints.Int2ObjectArrayMap;
 import net.minestom.server.entity.metadata.EntityMeta;
 import net.minestom.server.entity.metadata.ambient.BatMeta;
 import net.minestom.server.entity.metadata.animal.*;
@@ -40,6 +39,7 @@ import org.jetbrains.annotations.UnknownNullability;
 
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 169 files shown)


src/main/java/net/minestom/server/entity/MetadataHolder.java
```diff
@@ -60,11 +60,11 @@ public final class MetadataHolder {
     }
 
     private final Consumer<Map<Integer, Metadata.Entry<?>>> changesListener;
-    private final Int2ObjectMap<Metadata.Entry<?>> entries = new Int2ObjectOpenHashMap<>();
+    private Metadata.@Nullable Entry<?>[] entries = new Metadata.Entry<?>[0]; // indexed by metadata id, resized on demand
```
kermandev, line 62: Wouldnt a Int2ObjectArrayMap be applicable too? I think depending on the way you set your entries your performance characteristics would be widely different with this design.


(Though its likely read is also pretty hot too)

src/main/java/net/minestom/server/entity/MetadataHolder.java
```diff
@@ -161,183 +161,199 @@ public void setNotifyAboutChanges(boolean notifyAboutChanges) {
...
+    private void putEntry(int id, Metadata.Entry<?> entry) {
+        var entries = this.entries;
+        if (id >= entries.length) this.entries = entries = Arrays.copyOf(entries, id + 1);
+        entries[id] = entry;
+    }
+
     public Map<Integer, Metadata.Entry<?>> getEntries() {
-        return Map.copyOf(this.entries);
+        final var entries = this.entries;
+        Map<Integer, Metadata.Entry<?>> map = HashMap.newHashMap(entries.length);
```
kermandev: Wouldnt a Int2ObjectArrayMap be better here?

src/main/java/net/minestom/server/entity/MetadataHolder.java
TheMode: An array map require iterating over the entire array for every read/write, which is likely not ideal for +20/40 entries

kermandev (approved):
Seems fine now, we can likely do some future improvement if we know the upperbound of the metadata def (Or we could just double each time)

PR #3249, opened 2026-06-25
skip empty equipment slots (like vanilla)

Term4:
## Proposed changes

`EquipmentHandler#getEquipmentsPacket()` built the packet from every
`EquipmentSlot.values()` entry, including empty ones, so a naked entity still
produced an 8-slot all-AIR packet on each pairing. Vanilla
(`ServerEntity#sendPairing`) only sends non-empty slots and omits the packet
entirely when there are none. This aligns Minestom with that behavior and avoids
sending empty-slot data on every entity spawn.

## Types of changes

## Checklist

merging your code._

## Further comments

`getEquipmentsPacket()` previously never returned null; it is now `@Nullable`.
This is source/binary compatible, but external callers that send the result
directly should null-check it (the in-tree callers have been updated).

### DIFF
--- a/src/main/java/net/minestom/server/entity/LivingEntity.java
+++ b/src/main/java/net/minestom/server/entity/LivingEntity.java
@@ -543,7 +543,8 @@ public class LivingEntity extends Entity implements EquipmentHandler {
     @Override
     public void updateNewViewer(Player player) {
         super.updateNewViewer(player);
-        player.sendPacket(this.getEquipmentsPacket());
+        final var equipmentsPacket = this.getEquipmentsPacket();
+        if (equipmentsPacket != null) player.sendPacket(equipmentsPacket);
 
         if (shouldSendAttributes())
             player.sendPacket(this.getPropertiesPacket());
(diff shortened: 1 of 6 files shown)


src/main/java/net/minestom/server/network/player/PlayerSocketConnection.java
```diff
@@ -380,7 +380,9 @@ private boolean writePacketSync(NetworkBuffer buffer, SendablePacket packet, boo
                     }
                 }
                 case LazyPacket lazyPacket -> {
-                    PacketWriting.writeFramedPacket(buffer, state, lazyPacket.packet(), compressionThreshold);
+                    final ServerPacket serverPacket = lazyPacket.packet();
+                    if (serverPacket != null)
+                        PacketWriting.writeFramedPacket(buffer, state, serverPacket, compressionThreshold);
```
kermandev: Blocked on #3257, note this solution likely isn't valid as it violates the supplier

kermandev (approved):

kermandev:
Breaking API change due to the getEquipmentPacket's contract change

PR #3251, opened 2026-06-25
 respect instance view distance when client view distance changes

Term4:
## Proposed changes

`Player#refreshSettings` loads/unloads chunks on a client view distance change
using the **raw** client view distance, while spawn (`spawnPlayer`) and movement
(`sendChunkUpdates`) use the "effective" range `min(viewDistance, instanceViewDistance) + 1`.

Two problems with this:

- **It ignores the instance's view distance cap.** A client requesting a view
  distance larger than `instance.viewDistance()` is sent chunks beyond the cap.
  The movement/unload path respects the cap, so those chunks are never unloaded
  and they stay loaded (server side and client side) for the rest of the session.
- **It double-sends the outer ring** `min(prev, instanceVD) + 1`, which the
  spawn already sent.

Fix: use the same effective bounds in `refreshSettings` as spawn/movement, so a
view-distance change stays within the instance cap and NEVER re sends chunks the
client already has.

## Types of changes

## Checklist

## Further comments


kermandev (changes requested):
Needs test accompanying the where the client view distance exceeds the instance cap to keep the big fixed

kermandev (approved):

PR #3253, opened 2026-06-27
Fix/chunk section fluid count

Term4:
## Proposed changes

The section fluid count in the chunk packet was a placeholder (blockCount > 0 ? 1 : 0). The client decrements it as fluid blocks are removed and ignores the section's fluid once it reaches 0, so any block updates make players fall through water. This computes the actual count.

## Types of changes

## Checklist

merging your code._

## Further comments


src/main/java/net/minestom/server/instance/DynamicChunk.java
```diff
@@ -284,6 +286,22 @@ private ChunkDataPacket createChunkPacket() {
...
+        final int[] count = {0};
+        blockPalette.getAllPresent((x, y, z, stateId) -> {
+            if (isFluid(stateId)) count[0]++;
+        });
+        return (short) count[0];
+    }
+
+    private static boolean isFluid(int blockStateId) {
+        final Block block = Block.fromStateId(blockStateId);
+        return block != null && (block.isLiquid() || "true".equals(block.getProperty("waterlogged")));
```
kermandev: Doesnt cover all cases (kelp). Likely needs a custom field in the [MinestomDataGenerator](https://github.com/Minestom/MinestomDataGenerator) `fluid` like which is equal to `!blockState.getFluidState().isEmpty()`

Then 
```suggestion
        return block != null && block.isFluid();
```
Which then the method can be deleted by inlining.

kermandev (changes requested):

kermandev (approved):

PR #3269, opened 2026-07-07
Eliminate biome registry static usage

mudkipdev:


mudkipdev:
@kermandev it seems that the abi checker doesn't exclude internal subclasses

kermandev:
> @kermandev it seems that the abi checker doesn't exclude internal subclasses

Ignore it for now, though it might be a better question why the impl is public

kermandev (approved):

PR #3270, opened 2026-07-07
Fix ABI checker on nested internal classes

mudkipdev:


kermandev:
I was doing an update for ABI checker, this will be included in the update. Thanks for letting me know though

PR #3288, opened 2026-07-12
Match vanilla/Paper registry tag ordering

WouterGritter:

This is a bugfix/correctness PR.

We're working on a feature that skips the reconfiguring screen when switching between backend servers through Velocity. This is only possible when the registries match exactly between backend servers, otherwise very weird behavior will occur (biomes mismatch, overworld looks like end, all blocks are different, etc. etc.). See https://github.com/GemstoneGG/Velocity-CTD/pull/1004.

Paper/vanilla uses a stable registry tag order; it sorts them alphabetically. This means that skipping the reconfig screen between Paper backends (that are on the same (Minecraft) version) is possible, but unfortunately, because Minestom sorts the registry tags differently, it currently isn't possible to skip the reconfig screen when transfering from a Minestom backend to a Paper backend or vice versa.

This PR aligns Minestom's tag ordering with that of Paper, making it possible to skip the reconfig screen through a PR like the referenced PR (vanilla Velocity also has a open, terminally-unmerged PR open for this reconfig-skipping behavior, this PR would of course make Minestom compatible with that one as well).

Happy to add more tests covering this if needed.

### DIFF
--- a/src/main/java/net/minestom/server/instance/Section.java
+++ b/src/main/java/net/minestom/server/instance/Section.java
@@ -1,20 +1,44 @@
 package net.minestom.server.instance;
 
+import net.minestom.server.MinecraftServer;
 import net.minestom.server.instance.light.Light;
 import net.minestom.server.instance.palette.Palette;
+import net.minestom.server.registry.DynamicRegistry;
+import net.minestom.server.world.biome.Biome;
 
 public record Section(Palette blockPalette, Palette biomePalette, Light skyLight, Light blockLight) {
+    // Biome ids are no longer guaranteed to place plains at 0 (biomes are now sent in vanilla
+    // alphabetical order, see Biome#createDefaultRegistry), so the default "empty" biome is resolved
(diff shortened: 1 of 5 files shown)


R00tB33rMan (approved):

kermandev:
This is unlikely to be merged as is and will probably be superseded by `registries-v3`, where vanilla registries should use natural key ordering.

The ordering goal is fine, but this implementation is not. Minestom's current section/palette model treats numeric ID `0` as the empty/default value and relies on that corresponding to plains. This PR works around that by globally caching the plains ID in `Section`, which ignores the `Registries` attached to an `Instance` and initializes sections with the wrong biome ID when those registries differ. In which we are avoiding adding new usages of MinecraftServer and global caches.

Fixing this properly requires making section and palette defaults registry aware and changing assumptions in the existing API, so it should be handled as part of the larger, breaking registry rework.

`registries-v3` is planned for 26.3, or earlier if a breaking release window is created.

PR #3296, opened 2026-07-18
fix: Replace replaceable blocks in place when clicked

Term4:
## Proposed changes

Clicking a replaceable block (grass, snow layers, water) places the new block into its cell in vanilla, and modern clients predict that locally. The listener only honored self-replaceable placement rules for the clicked block and bumped everything else to the neighbor face, desyncing the prediction: the block visibly jumps one cell toward the player, or fails when that neighbor is occupied. The neighbor position already gets the `isReplaceable` check; the clicked position now gets the same one.

## Types of changes

## Checklist

## Further comments

Easy to reproduce: place a block against tall grass and watch it land one cell toward you after the client predicted it in place. The test covers the in-place replacement and that non-replaceable blocks still place against the face.

### DIFF
--- a/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
+++ b/src/main/java/net/minestom/server/listener/BlockPlacementListener.java
@@ -104,7 +104,10 @@ public class BlockPlacementListener {
         //todo it feels like it should be possible to have better replacement rules than this, feels pretty scuffed.
         Point placementPosition = blockPosition;
         var interactedPlacementRule = BLOCK_MANAGER.getBlockPlacementRule(interactedBlock);
-        if (!interactedBlock.isAir() && (interactedPlacementRule == null || !interactedPlacementRule.isSelfReplaceable(
+        // A replaceable clicked block (grass, snow layers, water) is replaced in place, like vanilla and the
+        // client's own prediction; the same check the neighbor position already gets below.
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/listener/BlockPlacementListener.java
```diff
@@ -104,7 +104,10 @@ public static void listener(ClientPlayerBlockPlacementPacket packet, Player play
         //todo it feels like it should be possible to have better replacement rules than this, feels pretty scuffed.
         Point placementPosition = blockPosition;
         var interactedPlacementRule = BLOCK_MANAGER.getBlockPlacementRule(interactedBlock);
-        if (!interactedBlock.isAir() && (interactedPlacementRule == null || !interactedPlacementRule.isSelfReplaceable(
+        // A replaceable clicked block (grass, snow layers, water) is replaced in place, like vanilla and the
+        // client's own prediction; the same check the neighbor position already gets below.
+        if (!interactedBlock.isAir() && !interactedBlock.registry().isReplaceable()
```
kermandev, line 109: Likely needs to exempt you holding the same item where its block that you are replacing is equivalent, so like placing grass while the item used is grass shouldnt replace but place it against.

kermandev, line 109: Seems to be a bug in the registry data with snow, where its replaceable is true where level > 1 should be false, just documenting that here for now

PR #3298, opened 2026-07-18
feat: Extract block placement and digging pipelines over a world interface

Term4:
## Proposed changes

Extracts the block placement and digging pipelines from `BlockPlacementListener` and `PlayerActionListener` into a public `BlockActions`, parameterized by a `World` interface (block access, the placement write, the break, chunk/read-only checks, the placement entity scan). Defaults resolve through the player's instance; the stock listeners delegate with the instance-backed world, behavior unchanged.

Virtual worlds and block overlays can then run the same flow (target resolution, events, acknowledgements, client resyncs) against their own storage. Today that requires forking both listeners and re-syncing the copies on every version bump.

## Types of changes

## Checklist

## Further comments

The listener bodies moved without behavior changes; existing placement integration tests pass as-is. The new test runs both pipelines against a map-backed overlay world and verifies the instance is never written. There was no digging test coverage before, so those tests are new as well.


kermandev:
I understand this would be useful for scratch, but id rather not do this change just yet especially since for scratch you don't want to depend on any of the control flow that we do, Id prefer if you still copied sections you wanted. Its likely to provide a much better experience

PR #3302, opened 2026-07-20
Add hashed seed setter

mudkipdev:


src/main/java/net/minestom/server/instance/Instance.java
```diff
@@ -488,6 +495,36 @@ public String getDimensionName() {
...
+    /// colors near a biome border depend on this value. Vanilla sends [#obfuscateSeed(long)] of the
+    /// world seed.
+    public void setHashedSeed(long hashedSeed) {
+        this.hashedSeed = hashedSeed;
+    }
+
+    /// Hashes a world seed the way vanilla does before sending it to a client.
+    ///
+    /// @see #setHashedSeed(long)
+    public static long obfuscateSeed(long seed) {
```
kermandev: Doesnt seem necessary to include as we dont track the vanilla seed to begin with

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -1735,13 +1737,26 @@ public boolean setGameMode(GameMode gameMode) {
      * Mostly unsafe since it requires sending chunks after.
      *
      * @param dimensionType the new player dimension
+     * @deprecated use {@link #sendDimension(RegistryKey, String, long)}
      */
+    @Deprecated(forRemoval = true)
     protected void sendDimension(RegistryKey<DimensionType> dimensionType, String dimensionName) {
```
kermandev: pointless to deprecate as its a breaking change as its no longer called

kermandev (approved):

PR #3308, opened 2026-07-25
Fix chunks disappearing when switching instances

### LINKED ISSUE #3307: [Bug]: Chunks disappear when switching instances
### Affected Version

2026.07.22-26.2

### Describe the bug

When a player switches instances, if the dimension types are the same and the spawn position is near the player's current position in their previous instance, something seems to go wrong with world rendering. The chunks are there, but they're invisible. Particles (e.g. ambient particles from torches), beacon beams, and entities still render, and the player still collides with the blocks and can see their selection boxes.

In this video, I'm switching instances 3 seconds after spawning in. Then, as soon as I cross a chunk boundary, the world becomes invisible.

https://github.com/user-attachments/assets/eed27ef1-7e14-4749-8c5a-85f9af112279

I'm connecting from Minecraft 26.2 on Fedora Linux with no mods using an Nvidia GPU. Happens with both OpenGL and Vulkan and when using all 3 chunk builder settings (threaded/semi-blocking/blocking). I don't see any warnings in the client logs.

Workarounds include:
- Making sure the player's spawn position in the new instance is far away from their current position in the old instance
- Teleporting the player far away and then back once they're spawned in the new instance
- Giving each instance a different dimension type

I don't think it's related to lighting, since the sample reproduces the bug with and without setting the instances' chunk supplier to LightingChunk.

1. Copy the code below and run the server
2. Log in and wait 3 seconds (you will be sent to a different instance)
3. Move around - make sure you cross a chunk border (F3+G to show chunk borders)
4. The ground beneath you should disappear

### Code sample

```java
import net.minestom.server.event.player.AsyncPlayerConfigurationEvent;
import net.minestom.server.instance.Instance;
import net.minestom.server.timer.TaskSchedule;

import java.time.Duration;

public class MinestomBug {

    private static Instance instance1, instance2;

    static void main() {
        MinecraftServer minecraftServer = MinecraftServer.init(new Auth.Online());

        instance1 = MinecraftServer.getInstanceManager().createInstanceContainer();
        instance2 = MinecraftServer.getInstanceManager().createInstanceContainer();

        instance1.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.GRASS_BLOCK));
        instance2.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.STONE));

        MinecraftServer.getGlobalEventHandler().addListener(AsyncPlayerConfigurationEvent.class, (event) -> {
            event.setSpawningInstance(instance1);
            event.getPlayer().setRespawnPoint(new Pos(0, 42, 0));

            MinecraftServer.getSchedulerManager().scheduleTask(() -> {
                var _ = event.getPlayer().setInstance(instance2);
                return TaskSchedule.stop();
            }, TaskSchedule.duration(Duration.ofSeconds(3)));
        });

        minecraftServer.start("0.0.0.0", 25565);
    }
}
```

FluxCapacitor2:
## Proposed changes

Fixes #3307

Starting in 26.2, Minestom's fast instance switching (changing instances without a loading screen) started to make chunks invisible if the old and new player positions were close to each other. This seems to be due to a Minecraft client bug ([MC-310041](https://bugs.mojang.com/browse/MC/issues/MC-310041)) where processing a chunk unload packet and a chunk data packet in the same frame caused the chunk to become invisible until it was unloaded and reloaded. This client issue is fixed in 26.3 snapshot 5, but 26.3 won't be released for a while, so I think it's worth implementing a workaround in the meantime.

This PR fixes the issue by skipping sending the `ChunkUnloadPacket` for chunks that will be within the client's render distance from their new spawn position.

This does mean that some of the old chunks are visible for a split second after joining the new instance. If we wanted to avoid that, we'd probably have to wait a tick in between sending all the `ChunkUnloadPacket`s and sending the new chunk data. However, I think the bug is related to frame rate, so if the client ever took more than one tick to render a frame, the chunks still may become invisible.

I intend for this PR to be reverted when 26.3 is released.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/src/main/java/net/minestom/server/entity/Player.java
+++ b/src/main/java/net/minestom/server/entity/Player.java
@@ -708,8 +708,25 @@ public class Player extends LivingEntity implements CommandSender, HoverEventSou
                              boolean firstSpawn, boolean dimensionChange, boolean updateChunks) {
         if (!firstSpawn && !dimensionChange) {
             // Player instance changed, clear current viewable collections
-            if (updateChunks)
-                ChunkRange.chunksInRange(spawnPosition, this.effectiveViewDistance(), chunkRemover);
+            if (updateChunks) {
+                final int viewDist = this.effectiveViewDistance();
+                ChunkRange.chunksInRange(
+                        (int) this.chunksLoadedByClient.x(), (int) this.chunksLoadedByClient.z(), viewDist,
+                        (x, z) -> {
+                            boolean inNewView = Math.abs(x - spawnPosition.chunkX()) <= viewDist
+                                    && Math.abs(z - spawnPosition.chunkZ()) <= viewDist;
+                            if (!inNewView) {
+                                // Only send UnloadChunkPacket for chunks no longer in the new view.
+                                // This alleviates a 26.2 client bug where, if it processes an UnloadChunkPacket
+                                // and a ChunkDataPacket for the same chunk in the same frame, the chunk disappears.
+                                // https://bugs.mojang.com/browse/MC/issues/MC-310041
+                                // TODO(26.3): Revert this change; the client bug is fixed in 26.3-snapshot5
+                                sendPacket(new UnloadChunkPacket(x, z));
+                            }
+                            EventDispatcher.call(new PlayerChunkUnloadEvent(this, x, z));
+                        }
+                );
+            }
         }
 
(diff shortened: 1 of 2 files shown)


kermandev:
Forgot to click submit on my review so I will just do the changes instead.

kermandev (approved):

PR #3318, opened 2026-08-01
Fix tab complete suggestions

athinaslime:
## Proposed changes

Tab completion stops working for any command argument that actually validates what you type, not just plain text. So if you're typing something that has to match a real value, pressing Tab shows nothing while you're still typing it, even though the suggestion appears once you finish typing a valid value.

Turns out the parser figures out suggestions by trying to parse whatever you have typed so far. Since you're mid-typing, that attempt almost always fails, which is expected. The problem is when it fails, the code falls back to an older copy from before the suggestion was ever attached, so it just gets lost.

AI WARNING: I did use Claude to find the fix, but due to Minestom not having an AI policy yet i still decided to make a pull request for this.

## Types of changes

## Checklist

merging your code._

## Further comments

### DIFF
--- a/src/main/java/net/minestom/server/command/CommandParserImpl.java
+++ b/src/main/java/net/minestom/server/command/CommandParserImpl.java
@@ -249,7 +249,13 @@ final class CommandParserImpl implements CommandParser {
                     // correct thing would be to say that the command is unknown.
                     if (!(childResult.chain.size() == 2 && childResult.argumentResult instanceof ArgumentResult.IncompatibleType<?>)) {
                         // If the last successful result is null, throw an exception instead of having unintended behaviour
-                        error = Objects.requireNonNull(childResult.chain().lastSuccessfulResult());
+                        NodeResult lastSuccess = Objects.requireNonNull(childResult.chain().lastSuccessfulResult());
+                        final SuggestionCallback deepestSuggestion = childResult.chain().suggestionCallback;
+
+                        if (deepestSuggestion != null) {
+                            lastSuccess.chain().suggestionCallback = deepestSuggestion;
+                        }
+                        error = lastSuccess;
                     }
                 }
                 reader.cursor(start);
--- a/src/main/java/net/minestom/server/command/CommandParserImpl.java
+++ b/src/main/java/net/minestom/server/command/CommandParserImpl.java
@@ -253,7 +253,15 @@ final class CommandParserImpl implements CommandParser {
                         final SuggestionCallback deepestSuggestion = childResult.chain().suggestionCallback;
 
                         if (deepestSuggestion != null) {
-                            lastSuccess.chain().suggestionCallback = deepestSuggestion;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/command/CommandParserImpl.java
```diff
@@ -249,7 +249,13 @@ private static NodeResult parseNode(CommandSender sender, Node node, Chain chain
                     // correct thing would be to say that the command is unknown.
                     if (!(childResult.chain.size() == 2 && childResult.argumentResult instanceof ArgumentResult.IncompatibleType<?>)) {
                         // If the last successful result is null, throw an exception instead of having unintended behaviour
-                        error = Objects.requireNonNull(childResult.chain().lastSuccessfulResult());
+                        NodeResult lastSuccess = Objects.requireNonNull(childResult.chain().lastSuccessfulResult());
+                        final SuggestionCallback deepestSuggestion = childResult.chain().suggestionCallback;
+
+                        if (deepestSuggestion != null) {
+                            lastSuccess.chain().suggestionCallback = deepestSuggestion;
```
kermandev: Mutating the chain is wrong, you need to fork the chain, then modify the `suggestionCallback` for that, and set lastSuccess as a new `NodeResult` using that chain.

Like

```suggestion
                            final Chain errorChain = lastSuccess.chain().fork();
                            errorChain.suggestionCallback = deepestSuggestion;
                            
                            lastSuccess = new NodeResult(
                                            lastSuccess.node(),
                                            errorChain,
                                            lastSuccess.argumentResult(),
                                            lastSuccess.callback(),
                            );
```

src/test/java/net/minestom/server/command/CommandSuggestionIntegrationTest.java
```diff
@@ -121,4 +121,25 @@ public void suggestionWithTwoLiterals(Env env) {
...
+        var command = new Command("foo");
+        command.addSyntax((_, _) -> {}, intArg);
+        env.process().command().register(command);
+
+        var listener = connection.trackIncoming(TabCompletePacket.class);
+        player.addPacketToQueue(new ClientTabCompletePacket(1, "foo "));
+        player.interpretPacketQueue();
+
+        listener.assertSingle(tabCompletePacket -> assertEquals(List.of(new TabCompletePacket.Match("42", null)), tabCompletePacket.matches()));
+    }
```
kermandev, line 144: Please add another test using two commands with syntax like

`/foo <integer>` (with suggestion 42)
`/foo <word> <integer>` (with no suggestions)

You should see the issue with the current code as you modify the chain, as you will see 42 suggested where it shouldnt be like `foo text nope` <--- 42 recommended for nope, where it shouldnt be

kermandev (changes requested):

athinaslime:
i applied your fix and wrote a new test, every test passes and suggestions still work.

kermandev (approved):

PR #3333, opened 2026-08-13
Allow advancement tabs to be unregistered

Fisher2911:
## Proposed changes

This allows advancement roots to be unregistered. Currently, the advancement tabs are there forever once they are created. This allows old advancements to be cleaned up and not stuck in memory.

I also changed the annotation on the `Viewable#getViewers` method from `@Unmodifiable` to `@UnmodifiableView` to make it more accurate. The current `AdvancementTab#getViewers` just directly returns the backing viewers set, so I also changed it to return an unmodifiable view.

## Types of changes

## Checklist

merging your code._

### DIFF
--- a/src/main/java/net/minestom/server/Viewable.java
+++ b/src/main/java/net/minestom/server/Viewable.java
@@ -7,7 +7,6 @@ import net.minestom.server.network.packet.server.SendablePacket;
 import net.minestom.server.network.packet.server.ServerPacket;
 import net.minestom.server.utils.PacketSendingUtils;
 import org.jetbrains.annotations.Unmodifiable;
-import org.jetbrains.annotations.UnmodifiableView;
 
 import java.util.Collection;
 import java.util.List;
@@ -35,25 +34,11 @@ public interface Viewable {
     boolean removeViewer(Player player);
 
     /**
-     * Removes all viewers
-     *
-     * @return all players that were removed
-     */
-    @Unmodifiable
-    default Set<? extends Player> clearViewers() {
-        final Set<? extends Player> copy = Set.copyOf(getViewers());
-        for (final Player player : copy) {
-            this.removeViewer(player);
-        }
-        return copy;
-    }
-
-    /**
      * Gets all the viewers of this viewable element.
      *
      * @return A Set containing all the element's viewers
      */
-    @UnmodifiableView
+    @Unmodifiable
     Set<? extends Player> getViewers();
 
     /**
--- a/src/main/java/net/minestom/server/Viewable.java
+++ b/src/main/java/net/minestom/server/Viewable.java
@@ -7,6 +7,7 @@ import net.minestom.server.network.packet.server.SendablePacket;
 import net.minestom.server.network.packet.server.ServerPacket;
 import net.minestom.server.utils.PacketSendingUtils;
 import org.jetbrains.annotations.Unmodifiable;
+import org.jetbrains.annotations.UnmodifiableView;
 
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/net/minestom/server/Viewable.java
```diff
@@ -33,12 +34,26 @@ public interface Viewable {
...
+        return copy;
+    }
+
     /**
      * Gets all the viewers of this viewable element.
      *
      * @return A Set containing all the element's viewers
      */
-    @Unmodifiable
+    @UnmodifiableView
```
kermandev: Rather not change this contract, UnmodifableView would be an stronger implementation detail, and provides little benefit for the library to require.

src/main/java/net/minestom/server/Viewable.java
```diff
@@ -33,12 +34,26 @@ public interface Viewable {
      */
     boolean removeViewer(Player player);
 
+    /**
+     * Removes all viewers
+     *
+     * @return all players that were removed
+     */
+    @Unmodifiable
+    default Set<? extends Player> clearViewers() {
```
kermandev: This would not be a valid implementation of this method. Though I dont think this is a necessary method to add in the first place as some implementations this would be a no op.

Also, It wasnt a valid implementation because it assumes all `getViewers` can actually be removed by `removeViewer` returning true, where its not required to return true at all.

src/main/java/net/minestom/server/advancements/AdvancementManager.java
```diff
@@ -53,4 +53,23 @@ public Collection<AdvancementTab> getTabs() {
         return advancementTabMap.values();
     }
 
+    /**
+     * See {@link Map#remove(Object)}
```
kermandev: Seems weird to call out Map#remove in the javadocs, doesnt seem useful here. Rather would prefer an actual description

src/main/java/net/minestom/server/advancements/AdvancementManager.java
kermandev, line 13: Needs javadocs here

src/main/java/net/minestom/server/advancements/AdvancementManager.java
```diff
@@ -53,4 +53,23 @@ public Collection<AdvancementTab> getTabs() {
...
+     *
+     * @param rootIdentifier key whose mapping is to be removed from the map
+     * @param removeViewers whether the current viewers of the advancement tab should be ignored. This
+     *                      is only used if the return value of this method is not null.
+     * @return the previous advancement tab associated with {@code rootIdentifier}, or {@code null}
+     * if there was .no tab associated with {@code rootIdentifier}
+     */
+    @Nullable
+    public AdvancementTab removeTab(String rootIdentifier, boolean removeViewers) {
```
kermandev: Is there a usecase where you would actually want the behavior of removeViewers to be false? That just seems like a good way to desync what advancements are registered on the client. Just seems to me like the boolean can be removed and made always true.

After that it seems clearViewers can be more optimal by inlining or creating a package private method on AdvancementTab to unregister that calls sendPacketToViewers instead of removerViewer on each one.

src/main/java/net/minestom/server/advancements/AdvancementTab.java
```diff
@@ -139,8 +141,20 @@ public synchronized boolean removeViewer(Player player) {
...
 
     @Override
+    @Unmodifiable
     public Set<? extends Player> getViewers() {
-        return viewers;
+        return Collections.unmodifiableSet(viewers);
+    }
+
+    protected void removeAllViewers() {
+        viewers.removeIf(player -> {
```
kermandev, line 150: Probably could be using the entire viewer list for the packet but its likely fine to use the old behavior.

kermandev (approved):

PR #3335, opened 2026-08-15
Item pick up fix

mudkipdev:


src/main/java/net/minestom/server/entity/LivingEntity.java
```diff
@@ -210,12 +210,16 @@ public void update(long time) {
         }
 
         // Items picking
-        if (canPickupItem() && itemPickupCooldown.isReady(time)) {
+        if (this instanceof Player player && player.getGameMode() == GameMode.SPECTATOR) {
```
kermandev: Player should likely override `canPickupItem` for this

src/main/java/net/minestom/server/entity/Player.java
```diff
@@ -447,7 +447,7 @@ public void update(long time) {
         super.update(time); // Super update (item pickup/fire management)
 
         // Experience orb pickup
-        if (experiencePickupCooldown.isReady(time)) {
+        if (gameMode != GameMode.SPECTATOR && experiencePickupCooldown.isReady(time)) {
```
```suggestion
        if (gameMode != GameMode.SPECTATOR && !isDead() && experiencePickupCooldown.isReady(time)) {
```

kermandev (approved):

PR #3336, opened 2026-08-15
Server address test fix

mudkipdev:
find free port by requesting a random one then closing the socket (small risk of a race)


kermandev (approved):
Better than it was before

PR #3342, opened 2026-08-17
Scratch status fix

mudkipdev:
this is stupid but I don't have a better idea

### DIFF
--- a/src/main/java/net/minestom/server/ping/Status.java
+++ b/src/main/java/net/minestom/server/ping/Status.java
@@ -193,7 +193,7 @@ public record Status(
         private Builder() {
             this.description = DEFAULT_DESCRIPTION;
             this.versionInfo = VersionInfo.DEFAULT;
-            this.playerInfo = PlayerInfo.onlineCount();
+            this.playerInfo = MinecraftServer.process() == null ? null : PlayerInfo.onlineCount();
         }
 
         private Builder(Status status) {


src/main/java/net/minestom/server/ping/Status.java
```diff
@@ -193,7 +193,7 @@ public static final class Builder {
         private Builder() {
             this.description = DEFAULT_DESCRIPTION;
             this.versionInfo = VersionInfo.DEFAULT;
-            this.playerInfo = PlayerInfo.onlineCount();
+            this.playerInfo = MinecraftServer.process() == null ? null : PlayerInfo.onlineCount();
```
Tofaa2: 😱

kermandev:
Likely just do `MinecraftServer.init` for now, other things require it anyways for scratch usage. Working on removing the init dependency but eventually `MinecraftServer.process` may throw if not initialized,  so this API may need to be reworked.

mudkipdev:
> eventually `MinecraftServer.process` may throw if not initialized

Why?

> Likely just do `MinecraftServer.init` for now

this defeats the point

kermandev:
> > eventually `MinecraftServer.process` may throw if not initialized
> 
> Why?

This is kind of a requirement for an immutable server process, which after init would be frozen into a holder, so calling `MinecraftServer.process()` is frozen, I guess it could support null in this holder but it seems like a better user expierence if we throw on null as its likely a bug forgetting to call `MinecraftServer.init()`.
> 
> > Likely just do `MinecraftServer.init` for now
> 
> this defeats the point

I agree, but its likely an issue of the API to depend on MinecraftServer as it would make it a framework component not a library one.

PR #3348, opened 2026-08-23
Fix Palette#compare between single-value and indirect palettes

### LINKED ISSUE #3334: Palette.compare returns false for a single value palette against an equal indirect palette
On `2026.08.07-26.2`, `Palette.compare` misjudges a single value palette against an indirect palette with identical content.

```java
Palette indirect = Palette.biomes(1);
for (int x = 0; x < 4; x++)
    for (int y = 0; y < 4; y++)
        for (int z = 0; z < 4; z++)
            indirect.set(x, y, z, 1);
Palette single = Palette.biomes();
single.fill(1);
single.compare(indirect); // false, content is identical per getAll
```

Reading both palettes back through `getAll` shows the value 1 at all 64 positions, and the javadoc says compare works independently of storage format. The zero bit branch appears to substitute `count` where `singleValue` belongs, so a uniformly filled palette compares its entry count against the other side's values.

Two different single value palettes (`fill(Block.STONE.stateId())` vs `fill(Block.DIRT.stateId())`) do return false correctly on this version.

Found while asserting that `optimize(Optimization.SIZE)` is content transparent across a collapse from one bit indirect to single value form.

la8garlic:
## Proposed changes

Fixes #3334.

`Palette#compare` rejected a single-value palette compared against an indirect palette with identical content. The `count` field in `PaletteImpl` is dual-purpose: it stores the palette value itself when `bitsPerEntry == 0`, and the non-air entry count otherwise. The early `count` equality check therefore compared a stored value (e.g. `1`) against an entry count (e.g. `64`) and returned false before the per-entry loop could run.

The fix only compares `count` between palettes using the same storage format, and falls through to the per-entry loop for mixed formats, where using `count` as the single value is already correct.

## Types of changes

## Checklist

## Further comments

Added `PaletteCompareTest` covering the reproducer from the issue (both directions), differing values, partially filled palettes, single value `0` vs an empty indirect palette, two single-value palettes, and single value vs direct storage. All existing palette tests and the root test suite pass.

### DIFF
--- a/src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
+++ b/src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
@@ -533,9 +533,13 @@ final class PaletteImpl implements Palette {
         final PaletteImpl palette = (PaletteImpl) p;
         final int dimension = this.dimension();
         if (palette.dimension() != dimension) return false;
-        if (palette.count != this.count) return false;
-        if (palette.count == 0) return true;
-        if (palette.bitsPerEntry == 0 && this.bitsPerEntry == 0) return true;
+        // `count` stores the palette value when bitsPerEntry == 0, and the non-air entry count
+        // otherwise, so it is only comparable between palettes using the same storage format
+        if (bitsPerEntry == 0 && palette.bitsPerEntry == 0) return count == palette.count;
+        if (bitsPerEntry > 0 && palette.bitsPerEntry > 0) {
+            if (palette.count != count) return false;
+            if (palette.count == 0) return true;
+        }
         final long[] thisValues = this.values;
         final long[] thatValues = palette.values;
         final int thisBpe = this.bitsPerEntry;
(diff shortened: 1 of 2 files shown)


src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
```diff
@@ -533,9 +533,13 @@ public boolean compare(Palette p) {
...
-        if (palette.count != this.count) return false;
-        if (palette.count == 0) return true;
-        if (palette.bitsPerEntry == 0 && this.bitsPerEntry == 0) return true;
+        // `count` stores the palette value when bitsPerEntry == 0, and the non-air entry count
+        // otherwise, so it is only comparable between palettes using the same storage format
+        if (bitsPerEntry == 0 && palette.bitsPerEntry == 0) return count == palette.count;
+        if (bitsPerEntry > 0 && palette.bitsPerEntry > 0) {
+            if (palette.count != count) return false;
+            if (palette.count == 0) return true;
+        }
```
kermandev, line 542: Missing the obvious fast path when mixed bpe

```suggestion
        final int count = count();
        if (count != palette.count()) return false;
        if (count == 0) return true;
```

PR #3349, opened 2026-08-23
Fix MojangUtils rejecting legacy 1-2 character usernames

### LINKED ISSUE #3231: [Bug]: PlayerSkin#fromUsername(String) broken
### Affected Version

latest master

### Describe the bug

Returns null

1. Join the server
2. Sneak
3. View the console and see nul

### Code sample

```java
package net.minestom.demo;

import net.minestom.server.entity.PlayerSkin;
import net.minestom.server.event.GlobalEventHandler;
import net.minestom.server.event.player.AsyncPlayerConfigurationEvent;
import net.minestom.server.event.player.PlayerStartSneakingEvent;
import net.minestom.server.instance.InstanceContainer;

public class MinimalRepro {

    static void main() {
        MinecraftServer server = MinecraftServer.init(new Auth.Online());

        InstanceContainer container = MinecraftServer.getInstanceManager().createInstanceContainer();
        container.setGenerator(unit -> unit.modifier().fillHeight(0, 40, Block.STONE));

        GlobalEventHandler geh = MinecraftServer.getGlobalEventHandler();
        geh.addListener(AsyncPlayerConfigurationEvent.class, event -> {
           event.setSpawningInstance(container);
           event.getPlayer().setRespawnPoint(new Pos(0, 42, 0));
        });
        geh.addListener(PlayerStartSneakingEvent.class, event -> {
            System.out.println(PlayerSkin.fromUsername("Notch"));
        });

        server.start("0.0.0.0", 25566);
    }

}

```

la8garlic:
## Proposed changes

Fixes #3231 (the username-validation part).

The `USERNAME_PATTERN` introduced in #3168 requires 3-16 characters, but as discussed in the issue, accounts created before Mojang introduced the 3-character minimum may legitimately have 1-2 character names, so `PlayerSkin#fromUsername` and `MojangUtils#getUUID` returned `null` for them. Verified against the live API: the 2-character account `ab` resolves fine once the pattern is relaxed.

Changes:
- Relax the pattern to `[a-zA-Z0-9_]{1,16}`
- URL-encode the username when building the lookup URL in `fromUsername`, so the request is safe independent of the character check (matching what #3168 already does for `authenticateSession`)

Note: the 17-19 character legacy exceptions mentioned in the issue discussion are intentionally not addressed here since there was no consensus; happy to extend the limit if maintainers prefer.

## Types of changes

## Checklist

## Further comments

Added `MojangUtilsTest` covering valid names (1-16 chars) and invalid ones (empty, >16 chars, and injection attempts such as `Notch?unsigned=true`), with no network access required. One additional observation from debugging: transient `null`s from these methods can also come from Mojang rate limiting, since any empty or error response is currently reported as `null` — a separate issue if maintainers want to improve the error reporting.

### DIFF
--- a/src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
+++ b/src/main/java/net/minestom/server/instance/palette/PaletteImpl.java
@@ -533,9 +533,13 @@ final class PaletteImpl implements Palette {
         final PaletteImpl palette = (PaletteImpl) p;
         final int dimension = this.dimension();
         if (palette.dimension() != dimension) return false;
-        if (palette.count != this.count) return false;
-        if (palette.count == 0) return true;
-        if (palette.bitsPerEntry == 0 && this.bitsPerEntry == 0) return true;
+        // `count` stores the palette value when bitsPerEntry == 0, and the non-air entry count
+        // otherwise, so it is only comparable between palettes using the same storage format
+        if (bitsPerEntry == 0 && palette.bitsPerEntry == 0) return count == palette.count;
+        if (bitsPerEntry > 0 && palette.bitsPerEntry > 0) {
+            if (palette.count != count) return false;
+            if (palette.count == 0) return true;
+        }
         final long[] thisValues = this.values;
         final long[] thatValues = palette.values;
         final int thisBpe = this.bitsPerEntry;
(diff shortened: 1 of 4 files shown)


kermandev:
Superseeded by #3350

PR #3354, opened 2026-08-29
feat: server property API

kermandev:
## Proposed changes

Add ServerProperty API as a togglable mutable replacement for ServerFlag. Mostly for testing purposes, which should let us have more diagnostic flags similar to openjdk/jdk. Though a followup PR will likely add testing support, as it likely needs an audit for usages that are dependent on behavior.

ServerFlag exposes each setting as a static final constant, so a setting can only be changed through a system property read before the class initializes, and never after.

ServerProperty replaces those constants with an object read through get(). A property still resolves its value from its system property when ServerProperties initializes. What changes is that setting `<name>.mutable`, or `minestom.properties.mutable` for all of them, makes that one property accept `set()`, which is very useful in testing environments.

Note, not all usages can properly use this and may not actually respond to behavior changes, but that can be worked on. The most notable one is the temporal tick unit.

## Proposed changes

## Types of changes

## Checklist

merging your code._

## Further comments

The `minestom.inside-test` will likely be removed in a future release, once more flags are added to control the behaviors. I know some libraries use this property to setup the testing suite, so it needs to be in a breaking changelist.

It is possible for `ServerProperties` to be implemented by `ServerProperty`, but I have refrained from doing that as its likely better to see the written implementation itself than a hidden interface like Blocks.


kermandev:
Not API breaking but behavior breaking with INSIDE_TEST no longer mutable. (no affect)

kermandev:
Future non breaking pr will add support for testing framework likely through annotations to only set a flag during a test.

PR #3356, opened 2026-08-29
fix: match vanilla's neighbor update order

Term4:
## Proposed changes

Match vanilla block update order: notifies neighbors west, east, down, up, north, south (NeighborUpdater.UPDATE_ORDER, unchanged since 1.8). The block update faces ran north/south before the vertical pair, which alters the "winner" whenever two neighbors react to the same change.

## Types of changes

## Checklist


kermandev:
So there is two update orders, one for blocks and one for redstone. Minestom does already do the correct update order for blocks. So I think changing this would be incorrect, maybe adding the other order is necessary instead?

Term4:
> So there is two update orders, one for blocks and one for redstone. Minestom does already do the correct update order for blocks. So I think changing this would be incorrect, maybe adding the other order is necessary instead?

ah yes you're right. Up to y'all if you think holding the other order in stom for implementations of redstone is needed.

PR #3366, opened 2026-09-14
fix: reject non-living entity types in EntityCreature

### LINKED ISSUE #2352: EntityCreature causes a Network Protocol Error when supplying a non-living EntityType.
EntityCreature extends LivingEntity, but does not have any checking to see if the supplied entity is a living entity type. If you supply a non-living entity type (say a Minecart), and then spawn the entity in an instance, it will instantly disconnect clients with a Network Protocol Error.

Command to replicate this:

```java
     public AITestCommand() {
        super("ai-test");

        ArgumentEntityType entityType = ArgumentType.EntityType("entity-type");

            if (sender instanceof Player player) {
                EntityType type = context.get(entityType);
                EntityCreature creature = new EntityCreature(type);
                creature.setInstance(player.getInstance(), player.getPosition());
            }
        }, entityType);
    }
```

HP-network:
Fixes #2352

EntityCreature extends LivingEntity but previously accepted any EntityType. Constructing it with a non-living type created incompatible metadata and could disconnect clients with a protocol error.

Validate the entity type before the superclass initializes and add regression coverage for rejection and normal living-entity construction.


src/main/java/net/minestom/server/entity/EntityCreature.java
```diff
@@ -36,10 +37,16 @@ public class EntityCreature extends LivingEntity implements NavigableEntity, Ent
      */
     @SuppressWarnings("this-escape") // deliberate self registration during construction
     public EntityCreature(EntityType entityType, UUID uuid) {
-        super(entityType, uuid);
+        super(validateEntityType(entityType), uuid);
         heal();
     }
 
+    private static EntityType validateEntityType(EntityType entityType) {
+        Check.argCondition(!entityType.shouldSendAttributes(),
```
thecodertommy, line 45: Could be `entityType.registry().isLiving()`

thecodertommy (approved):
Overall I think the class should be removed, as some of the code could be moved to an interface or implemented in a better way.

